Issue #14, opened 2022-08-05
chore: add perf changes from esbuild browser plugin


brandonroberts:
Adds checks to see if babel transform is needed

https://github.com/angular/angular-cli/pull/23682

pjlamb12:
I'm unsure where the change that starts [on line 371](https://github.com/angular/angular-cli/pull/23682/files#diff-309812edd0ff24c480525d73b2e4b7a0d82b91d2a689720f2e77395f7958bea7R371) of that PR should be added in Analog. My guess is in the object whose name is `name: '@analogjs/vite-plugin-angular-optimizer'` in the `angular-vite-plugin.ts` file, inside the transform method?

pjlamb12:
[These are the changes I've begun making](https://github.com/pjlamb12/analog/commit/80ca785f56164e1dfd15ec99ebbb53e808e342f4). I haven't worked with esbuild (or anything similar) before so I know this may need to change.

brandonroberts:
@pjlamb12 Thanks! Looking on track so far. You don't need to add the `advancedOptimizations` flag. Just use the `isProd` boolean that already exists inside the plugin. These changes would be added to both the `@analogjs/vite-plugin-angular-optimizer` and `@analogjs/vite-plugin-angular` transform functions.

Would be cleaner to separate the plugins into different files, but that can come as a separate effort.

PR #27, opened 2022-08-12
perf(vite-plugin-angular): add perf changes from esbuild browser builder

pjlamb12:
Relates to #14 

Added the performance changes to the `@analogjs/vite-plugin-angular-optimizer` and `@analogjs/vite-plugin-angular` `transform` methods.

### DIFF
--- a/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
+++ b/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
@@ -17,6 +17,7 @@ import {
 interface PluginOptions {
   tsconfig: string;
   workspaceRoot: string;
+  advancedOptimizations?: boolean;
 }
 
 interface EmitFileResult {
@@ -186,10 +187,28 @@ export function angular(
 
           // return fileEmitter
           const data = typescriptResult?.content ?? '';
+          const forceAsyncTransformation =
+            /for\s+await\s*\(|async\s+function\s*\*/.test(data);
+          const useInputSourcemap = (!isProd ? undefined : false) as undefined;
+
+          if (
+            !forceAsyncTransformation &&
+            !pluginOptions.advancedOptimizations
+          ) {
+            return {
+              // Strip sourcemaps if they should not be used
+              contents: useInputSourcemap
+                ? data
+                : data.replace(/^\/\/# sourceMappingURL=[^\r\n]*/gm, ''),
+              loader: 'js',
+            };
+          }
 
           const babelResult = await transformAsync(data, {
             filename: id,
-            inputSourceMap: (!isProd ? undefined : false) as undefined,
+            inputSourceMap: (useInputSourcemap
+              ? undefined
+              : false) as undefined,
             sourceMaps: !isProd ? 'inline' : false,
             compact: false,
             configFile: false,
@@ -200,7 +219,7 @@ export function angular(
               [
                 angularApplicationPreset,
                 {
-                  forceAsyncTransformation: data.includes('async'),
+                  forceAsyncTransformation,
                   optimize: isProd && {},
                 },
               ],
--- a/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
+++ b/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
@@ -184,17 +184,24 @@ export function angular(
 
           const typescriptResult = await fileEmitter!(id);
 
+          const linkerPluginCreator = (
+            await loadEsmModule<
+              typeof import('@angular/compiler-cli/linker/babel')
+            >('@angular/compiler-cli/linker/babel')
+          ).createEs2015LinkerPlugin;
+
           // return fileEmitter
           const data = typescriptResult?.content ?? '';
           const forceAsyncTransformation =
             /for\s+await\s*\(|async\s+function\s*\*/.test(data);
           const useInputSourcemap = (!isProd ? undefined : false) as undefined;
+          const shouldLink = await requiresLinking(id, code);
 
           if (!forceAsyncTransformation && !isProd) {
             return {
-              code: !isProd
-                ? data
-                : data.replace(/^\/\/# sourceMappingURL=[^\r\n]*/gm, ''),
+              code: isProd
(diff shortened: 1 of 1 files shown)


packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
```diff
@@ -186,10 +186,25 @@ export function angular(
 
           // return fileEmitter
           const data = typescriptResult?.content ?? '';
+          const forceAsyncTransformation =
+            /for\s+await\s*\(|async\s+function\s*\*/.test(data);
+          const useInputSourcemap = (!isProd ? undefined : false) as undefined;
+
+          if (!forceAsyncTransformation && !isProd) {
+            return {
```
brandonroberts, line 194: Instead of `contents` and `loader`, this should just return an object of `code` and `map`. Contents and loader are specific to esbuild transforms, and we are reusing some of that logic in Rollup transforms

```ts
{
  code: string, // typescript result content minus sourcemap string
  map: string // sourcemap from typescript if enabled
}
```

packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
```diff
@@ -244,8 +259,21 @@ export function angular(
...
 
+          const forceAsyncTransformation =
+            !/[\\/][_f]?esm2015[\\/]/.test(id) &&
+            /for\s+await\s*\(|async\s+function\s*\*/.test(code);
           const useInputSourcemap = !isProd;
 
+          if (!forceAsyncTransformation && !isProd) {
+            return {
+              // Strip sourcemaps if they should not be used
+              contents: useInputSourcemap
```
brandonroberts, line 270: Same as above

brandonroberts (changes requested):

packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
pjlamb12: Okay, so something like this?

```
return {
  code: data.replace(/^\/\/# sourceMappingURL=[^\r\n]*/gm, ''),
  map: useInputSourcemap ? data : null,
};
```

brandonroberts: Something like this should be enough

```ts
              return {
                code: !isProd
                ? data
                : data.replace(/^\/\/# sourceMappingURL=[^\r\n]*/gm, ''),
            };
```

packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
```diff
@@ -244,8 +257,19 @@ export function angular(
             >('@angular/compiler-cli/linker/babel')
           ).createEs2015LinkerPlugin;
 
+          const forceAsyncTransformation =
```
brandonroberts, line 260: The check for the optimizer should be closer to these lines, along with the returned object of `code`

https://github.com/angular/angular-cli/pull/23682/files#diff-309812edd0ff24c480525d73b2e4b7a0d82b91d2a689720f2e77395f7958bea7R371

brandonroberts, line 260: This should be added to the `esbuild` object above

https://github.com/angular/angular-cli/pull/23682/files#diff-a99adf3526ebc926333013ad739eafc4edfc01af13f25822552c49295dea6defR310

brandonroberts (changes requested):
Looking good, needs a couple of tweaks

packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
pjlamb12: Do you mean closer as in it should look more like the code referenced here?

brandonroberts: The code I referenced should be copied into the transform for the optimizer plugin section. It's not the same as the code in the vite-plugin transform section

pjlamb12: Okay. Sorry for all the questions 🙈

I think this is more what you're looking for:

```
const forceAsyncTransformation =
  !/[\\/][_f]?esm2015[\\/]/.test(id) &&
  /for\s+await\s*\(|async\s+function\s*\*/.test(code);
const shouldLink = await requiresLinking(id, code);
const useInputSourcemap =
  pluginOptions.sourcemap &&
  (!!pluginOptions.thirdPartySourcemaps || !/[\\/]node_modules[\\/]/.test(id));

// If no additional transformations are needed, return the TypeScript output directly
if (!forceAsyncTransformation && !isProd && !shouldLink) {
  return {
    // Strip sourcemaps if they should not be used
    code: useInputSourcemap
      ? code
      : code.replace(/^\/\/# sourceMappingURL=[^\r\n]*/gm, ""),
  };
}

if (!forceAsyncTransformation && !isProd) {
  return {
    code: !isProd
      ? code
      : code.replace(/^\/\/# sourceMappingURL=[^\r\n]*/gm, ""),
  };
}
```

The linked code you posted has reference to `pluginOptions.sourcemap` and `pluginOptions.thirdPartySourcemaps`. I'm not sure what the equivalent would be here. In addition, the two returns in the if statements here are the same, so I'm not sure which I should use, or if both should be present but one changed at all.

brandonroberts: No problem! The shouldLink is present in the second transform but not in the first. I think that's the biggest difference. The isProd should take the place of any changes for advanced optimizations and sourcemaps flags.

eneajaho:
Hi, angular team added some other perf changes in its own builder regarding styles. It may interest you since you're working on this. 

image

brandonroberts:
Thanks Enea, we don't use the sass functionality from the compiler plugin, as that's handled by Vite. We mainly use the plugin with Vite for optimizing the Angular/third-party dependencies.

brandonroberts (approved):

brandonroberts:
Thanks! @allcontributors add @pjlamb12 for code

PR #28, opened 2022-08-13
feat(astro-integration-angular): move ng_astro to analog monorepo

### LINKED ISSUE #24: [LIB]: Migrate Angular Integration for Astro to this repo
I've developed a plugin for Astro that let's use Angular components inside of an Astro project. The repository currently lives here https://github.com/brandonroberts/ng-astro in the `integration-angular` folder.

This effort would:

- Move that integration to this repo
- Rename the package to @analogjs/astro-integration-angular
- Update the package to support the new `createApplication` API upcoming in Angular 14.2 (TBD)

mainawycliffe:
Work in Progress 

Closes #24


brandonroberts:
You don't need to bring over the Astro Blog example app. Will probably create something else from scratch that looks a bit better

brandonroberts:
And would prefer to convert the source files to TypeScript and compile them in the build step

mainawycliffe:
Will remove them, for now I need something to test the integration with, will clean up afterwards, probably shouldn't have merge that in

mainawycliffe:
In terms of converting to TS, I will do that, hoping to wrap this over the weekend.

brandonroberts:
👍 As you were 😂

mainawycliffe:
I think I have everything setup correctly, just need to figure out testing now

brandonroberts:
Is this ready for review?

packages/astro-integration-angular/src/index.ts
```diff
@@ -0,0 +1,32 @@
...
+  };
+}
+
+function getViteConfiguration() {
+  return {
+    optimizeDeps: {
+      include: ['@analogjs/astro-integration-angular/client.js'],
+      exclude: ['@analogjs/astro-integration-angular/server.js'],
+    },
+    plugins: [(viteAngular as any).default()],
```
brandonroberts, line 18: This should just be `angular()` without any casting

packages/astro-integration-angular/package.json
```diff
@@ -0,0 +1,40 @@
...
+  "bugs": {
+    "url": "https://github.com/analogjs/analog/issues"
+  },
+  "homepage": "https://github.com/analogjs/analog#readme",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/analogjs/analog.git"
+  },
+  "dependencies": {
+    "@analogjs/vite-plugin-angular": "^0.2.0-alpha.0"
```
brandonroberts, line 35: Change to `latest` for now instead of the alpha version

brandonroberts (changes requested):

packages/astro-integration-angular/package.json
mainawycliffe: Yeah, this makes sense

packages/astro-integration-angular/src/index.ts
mainawycliffe: `angular()` works, but the plugin doesn't work with a project, as I am getting the following error `angular is not a function` when testing, I am guessing this has to do with the output of the vite plugin, which produces an export in the form of `exports.default` which seems incompatible with Astro integration which is ESM based. any ideas?

mainawycliffe: I am guessing this is the reason you had the workaround to begin with

brandonroberts: Yep, that was the reason. The vite plugin is currently shipped as commonjs. I think when I tried shipping ESM, it broke when used with Vite. Something more to investigate there, but I think we're good for now here.

mainawycliffe: Let me add a comment to not this right now and then create an issue around this

brandonroberts (approved):
LGTM

Issue #64, opened 2022-08-30
[FEAT]: Add support for prod mode for multiple components in Astro


brandonroberts:
When rendering Angular components, you want to `enableProdMode()` when building for production. Using enableProdMode multiple times throws an error.

Should add a custom/internal function to call enableProdMode. Need to only call it one time, set a flag, and skip it upon subsequent calls.

PR #66, opened 2022-08-30
feat(astro-angular): enabled angular prod build

simitch1:
I think that to enable angular prod mode like requested in this issue #64 , we could simply use the astro:build:start or astro:build:setup hook to enable the prod compilation a single time. 
I tried the build locally and seems to work, but I don't really know how to test further.

Before:
image

After:
image

Tests are failing right know but build is good. If the solution is ok I can work on it.
Let me know what do you think.


brandonroberts:
The test failure is probably because it needs the jest-preset-angular package for testing. We can switch to running the tests with Vitest and the Angular plugin

brandonroberts:
Based on the Astro docs, this seems like the appropriate place to use this.

https://docs.astro.build/en/reference/integrations-reference/#astrobuildsetup

simitch1:
I switched from jest to vitest. Modified the tests and used the astro:build:setup hook instead of astro:build:start.

brandonroberts:
Sorry about the miscommunication. I meant to post the build start link instead of the build setup one. I think either is suitable though so it can stay as is.

brandonroberts (approved):
LGTM

brandonroberts:
Thanks a bunch! @allcontributors add @simitch1 for code

Issue #60, opened 2022-08-29
[FEAT]: Add support for passing static props to an Angular component in Astro


brandonroberts:
### Which scope/s are relevant/related to the feature request?

astro-angular

### Information

You can use a self-contained Angular component in Astro.

```html
<ProductsListComponent />
```

You should be able to pass down static props to the component and consume them in the Angular component.

In Astro template
```html
---
const products = await getProducts();
---

<ProductsListComponent products={products} />
```

In Angular component
```ts
import { CommonModule } from '@angular/common';
import { Component } from '@angular/core';
import { STATIC_PROPS, injectStaticProps } from '@analogjs/astro-angular';

@Component({
  selector: 'app-hello',
  standalone: true,
  imports: [CommonModule],
  template: `
    <p>Hello from Angular!!</p>

    <button (click)="toggle()">Toggle</button>
    
    <p *ngIf="show">Toggled</p>
  `,
})
export class ProductsListComponent {
  products = injectStaticProps<Product[]>(); // <-- using inject function

  constructor(@Inject(STATIC_PROPS) private ngProps: Product[]) {} // <-- using constructor injection
}
```

The token would be added to the `providers` of the component:

Through server rendering: https://github.com/analogjs/analog/blob/main/packages/astro-angular/src/server.ts#L21

Through client hydration: https://github.com/analogjs/analog/blob/main/packages/astro-angular/src/client.ts#L12

### I would be willing to submit a PR to fix this issue

- [x] Yes
- [ ] No

PR #67, opened 2022-08-30
feat(astro-angular): add props (inputs) processing

nartc:
- client: using `ComponentRef#setInput`
- server: using `BEFORE_APP_INITIALIZED` hook to run `setInput` 
- add `astro-app` to test Astro integration

Still need to add some sort of tests for this feature. Any suggestions on how we'd want to test this?

ISSUE: #60


brandonroberts:
Some e2e tests for the astro-app works for now. I'm not sure what the testing story is for integrations. I didn't find any in the source code for official Astro one yet.

package.json
```diff
@@ -8,7 +8,7 @@
     "postinstall": "node ./decorate-angular-cli.js && ngcc --properties es2020 browser module main",
     "prepare": "git config core.hookspath .githooks",
     "start": "nx serve",
-    "build": "nx run-many --target build --all",
+    "build": "nx run-many --target build --all --exclude=astro-app",
```
brandonroberts, line 11: Do we need to exclude this?

packages/astro-angular/src/client.ts
```diff
@@ -1,12 +1,12 @@
...
 import { ApplicationRef, NgZone, createComponent } from '@angular/core';
-import type { ɵComponentType } from '@angular/core';
+import { createApplication } from '@angular/platform-browser';
 
 export default (element: HTMLElement) => {
   return (
-    Component: ɵComponentType<unknown>,
-    _props?: unknown,
+    Component: ComponentType<unknown>,
+    _props?: Record<string, unknown>,
```
brandonroberts, line 9: I used the underscore for unused arguments.

```suggestion
    props?: Record<string, unknown>,
```

packages/astro-angular/src/client.ts
```diff
@@ -16,6 +16,13 @@ export default (element: HTMLElement) => {
           environmentInjector: appRef.injector,
           hostElement: element,
         });
+
+        if (_props) {
```
brandonroberts, line 20:
```suggestion
        if (props) {
```

packages/astro-angular/src/server.ts
```diff
@@ -1,35 +1,98 @@
...
+}>('Static props token w/ mirror', {
+  factory() {
+    return { props: {}, mirror: {} as ComponentMirror<unknown> };
+  },
+});
 
 function check(
-  Component: ɵComponentType<unknown>,
+  Component: ComponentType<unknown>,
   _props: Record<string, unknown>,
```
brandonroberts, line 28:
```suggestion
   props: Record<string, unknown>,
```

packages/astro-angular/src/server.ts
```diff
@@ -1,35 +1,98 @@
...
   _children: unknown
 ) {
-  return !!Component['ɵcmp'];
+  return !!reflectComponentType(Component);
 }
 
+// Run beforeAppInitialized hook to set Input on the ComponentRef
+// before the platform renders to string
+const STATIC_PROPS_HOOK_PROVIDER: Provider = {
```
brandonroberts, line 36: nit/question: Why HOOK_PROVIDER instead of just PROVIDER?

packages/astro-angular/src/server.ts
```diff
@@ -1,35 +1,98 @@
...
+};
+
 async function renderToStaticMarkup(
-  Component: ɵComponentType<unknown>,
+  Component: ComponentType<unknown>,
   _props: Record<string, unknown>,
   _children: unknown
 ) {
-  const appId = getSelector(Component);
+  const mirror = reflectComponentType(Component);
```
brandonroberts, line 77: 🔥

packages/astro-angular/src/server.ts
```diff
@@ -1,35 +1,98 @@
...
+} from '@angular/core';
+import {
+  BEFORE_APP_SERIALIZED,
+  renderApplication,
+} from '@angular/platform-server';
+
+const ANALOG_ASTRO_STATIC_PROPS = new InjectionToken<{
+  props: Record<string, unknown>;
+  mirror: ComponentMirror<unknown>;
+}>('Static props token w/ mirror', {
```
brandonroberts, line 20:
```suggestion
}>('@analogjs/astro-angular: Static Props w/ Mirror Provider', {
```

brandonroberts (changes requested):
🪄

packages/astro-angular/src/server.ts
nartc: It is mainly for the clear intent of the provider. 
- It is a `PROVIDER`
- It is to provide a `BEFORE_APP_INITIALIZED` hook -> `HOOK`
- The hook is to setup static props for the component -> `STATIC_PROPS`

package.json
nartc: We do not need to exclude it but it also doesn't bring any value to the CI (assuming e2e for it is in place)

brandonroberts: 👍

packages/astro-angular/src/server.ts
brandonroberts: Got it. Thanks

packages/astro-angular/src/server.ts
nartc: We're not using `_props` in `check()` yet so keeping `_props` as is

brandonroberts (approved):
LGTM 🚀

Issue #214, opened 2023-01-16
Feature: Add ability to provide meta tags via `RouteMeta`


markostanimirovic:
### Which scope/s are relevant/related to the feature request?

router

### Information

Add the ability to provide meta tags for a specific page via `RouteMeta` in the following way:

```ts
// routes/index.ts

export const routeMeta: RouteMeta = {
  title: 'Home',
  meta: [
    { name: 'description', content: 'Home Page Description' },
    { name: 'author', content: 'Marko' },
    // ...
  ],
};

@Component({ /* ... */ })
export default class HomeComponent {}
```

Meta tags can be also set via resolver:

```ts
export const routeMeta: RouteMeta = {
  title: 'Home',
  meta: () => inject(MetaService).getMetaTags(),
};
```

### I would be willing to submit a PR to fix this issue

- [X] Yes
- [ ] No

PR #230, opened 2023-01-23
feat(router): add ability to provide meta tags via RouteMeta

markostanimirovic:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] vite-angular-plugin
- [ ] astro-angular
- [ ] create-analog
- [x] router
- [ ] platform
- [ ] content

Issue Number: #214

- [ ] Yes
- [x] No


apps/blog-app/src/app/routes/blog.ts
```diff
@@ -2,6 +2,19 @@ import { Component } from '@angular/core';
 import { RouterLink, RouterOutlet } from '@angular/router';
 import { injectContentFiles } from '@analogjs/content';
 import { NgFor } from '@angular/common';
+import { RouteMeta } from '@analogjs/router';
+
+export const routeMeta: RouteMeta = {
+  title: 'Blog Title',
+  metaTags: [
```
markostanimirovic, line 9: @brandonroberts

Thoughts on naming this property `metaTags`, `metatags`, or `meta`? 👀

brandonroberts: I think `meta` is good. If they ever introduce this property upstream in the router, we may have to adjust things 😅

packages/router/src/lib/meta-tags.ts
```diff
@@ -0,0 +1,99 @@
...
+type MetaTagMap = Record<MetaTagSelector, MetaTag>;
+
+export function updateMetaTagsOnNavigationEnd(): void {
+  const router = inject(Router);
+  const metaService = inject(Meta);
+
+  router.events
+    .pipe(filter((event) => event instanceof NavigationEnd))
+    .subscribe(() => {
+      const metaTagMap = getMetaTagMap(router.routerState.snapshot.root);
```
brandonroberts, line 53: If they visit a route with meta tags then one without, are the previous tags left over?

markostanimirovic: Yes!

I'd keep it as it is to avoid additional DOM manipulations. I think the most important is to have valid meta tags when a specific page is initially loaded by the browser or any other app.

What do you think? 👀 

Btw, there is the same behavior with Angular `title` - if we don't provide a new title for a route, the previous title will remain when that route is activated.

brandonroberts: I agree with that behavior along with the router title also.

brandonroberts (approved):

PR #231, opened 2023-01-23
fix(vite-plugin-angular): use babel to make transform results compati…

### LINKED ISSUE #202: Analog not working in Webkit based browsers

Analog Project's do not work in webkit based browsers. 
I believe it has something to do with the Vite Angular plugin. 

_Unfortunately, you cannot reproduce this in a StackBlitz example so I cloned the analog repo and added my tests to the blog-app & analog-app applications, which both show the same behavior_

vite-plugin-angular

### Description

**Steps to reproduce:**
1. Add some basic functionality to the AppComponent, e.g.
```
@Component({
  selector: 'analogjs-root',
  standalone: true,
  imports: [TopBarComponent, RouterOutlet],
  template: `
    <app-top-bar></app-top-bar>

    Counter: {{count}}
    <button (click)="increment()">Increment</button>

    <div class="container">
      <router-outlet></router-outlet>
    </div>
  `,
})
export class AppComponent {
  public count = 0;
  public increment(): void {
    this.count++;
  }
}
```
2. Open a **webkit** based browser such as **Safari** (on MacOS/iOS), but also Chrome on iOS
3. Click on the Increment button

**Expected Result**
The counter increases on every click.

**Actual Result**
The counter does not increase.
The console shows the following error:
Screenshot 2023-01-10 at 10 52 35 AM

If you drill into the error the following is shown:
image

```true
[Error] SyntaxError: Unexpected token '{'
	(anonymous function) (app.component.ts:12)
```

### Other information

I would love to submit a PR to fix this issue but I currently am lacking the knowledge where to even start investigating this issue.

- [X] Yes
- [ ] No

goetzrobin:
…ble with Webkit

Angular's compilation results use static blocks by default. These static blocks are not supported by Safari or other webkit based browsers. https://caniuse.com/mdn-javascript_classes_static_initialization_blocks We need to use babel transformations to make the output compatible with iOS. This is for serving content in development as well as building for production.

Closes #202

## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [x] vite-angular-plugin
- [ ] astro-angular
- [ ] create-analog
- [ ] router
- [ ] platform
- [ ] content

Analog output is broken in Webkit based (Safari & all iOS) browsers.

Issue Number: #202

## What is the new behavior?

Analog works as expected in Webkit based browsers

- [ ] Yes
- [x] No

## Other information

This is just a PR to share my findings and the patch that got it to work.
I hope someone with more insight can explain why it works and if there is a better way to achieve the same results.

### DIFF
--- a/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
+++ b/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
@@ -25,6 +25,7 @@ export interface PluginOptions {
      */
     tsTransformers?: ts.CustomTransformers;
   };
+  supportedBrowsers?: string[];
 }
 
 interface EmitFileResult {
@@ -63,6 +64,7 @@ export function angular(options?: PluginOptions): Plugin[] {
           options?.advanced?.tsTransformers?.afterDeclarations ?? [],
       },
     },
+    supportedBrowsers: options?.supportedBrowsers ?? ['iOS <=15'],
   };
 
   // The file emitter created during `onStart` that will be used during the build in `onLoad` callbacks for TS files
@@ -273,6 +275,7 @@ export function angular(options?: PluginOptions): Plugin[] {
               [
                 angularApplicationPreset,
                 {
+                  supportedBrowsers: pluginOptions.supportedBrowsers,
                   forceAsyncTransformation,
                   optimize: isProd && {},
                 },


brandonroberts:
The second transform is only applied at build time to Angular libraries that need it. What's the error with Prism?

goetzrobin:
I'll need to find some time to look at it again, but it was something along the lines of:
Cannot find module ... index when index.js is expected in an mjs file.

I think applying the transformation only on build would be fine too. I think devs can live with ng serve only working in non WebKit browsers as long as the assets built work in Safari and every iOS browser

goetzrobin:
@brandonroberts the error was:
`"default" is not exported by "node_modules/front-matter/index.js", imported by "node_modules/@analogjs/content/fesm2015/analogjs-content.mjs".`

I updated the PR to what I think is the least intrusive way of getting 
your production analog app working in safari by default, 
while still allowing you to customize the behavior if needed.

This keeps the development experience as is but makes the built assets work on iOS and Safari.

Let me know what you think 👍

brandonroberts:
@goetzrobin I like this option and I think that's a reasonable change

brandonroberts (approved):

Issue #293, opened 2023-03-17
Feature: Add support for generating a new application in an Nx workspace


brandonroberts:
### Which scope/s are relevant/related to the feature request?

create-analog

### Information

This effort would add a generator to support creating a new Analog app within an Nx workspace

```sh
nx g @analogjs/platform:app
```

This would do the same as `create-analog` but would customize the app name, and enable integration with Nx console.

### Describe any alternatives/workarounds you're currently using

Generate a new Angular app and add Analog support manually.

https://github.com/brandonroberts/analog-nx-migration

### I would be willing to submit a PR to fix this issue

- [ ] Yes
- [ ] No

goetzrobin:
@brandonroberts I was playing around with a generator like this and have a POC working! 

I might need some help with things setting up. I am thinking of the publishing of the package and some input on a couple requirements, but I can work on this!

I assume it should take similar inputs to the create-analog cli?

The Nx workspace will already have an Angular version in its package.json. What validations should happen? What versions can be supported?

Should there be an option for an e2e app to be created or is that something we want to add later?

PR #308, opened 2023-03-27
feat(nx-plugin): add initial setup of nx-plugin. still a work in prog…

goetzrobin:
…ress

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] vite-angular-plugin
- [ ] astro-angular
- [ ] create-analog
- [ ] router
- [ ] platform
- [ ] content
- [x] nx-plugin

No Nx generator exists

Issue Number: #293 

## What is the new behavior?

Adds Nx generator to the project

- [ ] Yes
- [x] No


goetzrobin:
@brandonroberts moved the plugin into the `platform` package! 

We are able to get an application up and running with `nx g @analogjs/platform:app hello-world` and `nx run hello-world:serve`. We can also successfully build it with `nx run hello-world:build`. 

Documentation on this is still missing for now. I wanted to ask you for your input on that. Where does it make the most sense to put it? As part of the Getting Started page or somewhere else? 
Looking forward to reading your thoughts! 🚀 👍

packages/platform/generators.json
```diff
@@ -0,0 +1,12 @@
...
+  "$schema": "http://json-schema.org/schema",
+  "name": "nx-plugin",
+  "version": "0.0.1",
+  "generators": {
+    "app": {
+      "factory": "./src/lib/nx-plugin/generators/app/generator",
+      "schema": "./src/lib/nx-plugin/generators/app/schema.json",
+      "description": "Generates an AnalogJs application"
+    }
+  }
```
nartc, line 11: suggestion: let's also add a `schematics` property for AngularCLI users as well

packages/platform/package.json
```diff
@@ -26,5 +26,7 @@
   "dependencies": {
     "nitropack": "^1.0.0",
     "@analogjs/vite-plugin-angular": "latest"
-  }
+  },
+  "generators": "./generators.json",
```
nartc, line 30: suggestion: let's also add `schematics` property here pointing to the same `generators.json` with the above change

packages/platform/src/lib/nx-plugin/executors/build/executor.ts
```diff
@@ -0,0 +1,8 @@
+import { BuildExecutorSchema } from './schema';
+
+export default async function runExecutor(options: BuildExecutorSchema) {
```
nartc, line 3: suggestion: let's just remove executors if we're not having any executors

packages/platform/src/lib/nx-plugin/generators/app/generator.ts
```diff
@@ -0,0 +1,109 @@
...
+  projectRoot: string;
+  projectDirectory: string;
+  parsedTags: string[];
+}
+
+function normalizeOptions(
+  tree: Tree,
+  options: AnalogNxApplicationGeneratorOptions
+): NormalizedOptions {
+  const name = names(options.name).fileName;
```
nartc, line 29: suggestion: we already called `names()` here, let's return the value we need from `names()` as part of `NormalizeOptions` as well so we don't have to rerun `names()` in `addFiles()`

packages/platform/src/lib/nx-plugin/generators/app/generator.ts
```diff
@@ -0,0 +1,109 @@
...
+    parsedTags,
+  };
+}
+
+async function addDependencies(tree: Tree, majorAngularVersion: number) {
+  const {
+    dependencies,
+    devDependencies,
+    // eslint-disable-next-line @typescript-eslint/no-var-requires
+  } = require(`./files/v${majorAngularVersion}-package.json`);
```
nartc, line 53: suggestion: it is easier to maintain if you have a `versions.ts` file containing constants for versions. Then you can pass the versions in the templates instead of requiring from template file like this

packages/platform/src/lib/nx-plugin/generators/app/generator.ts
```diff
@@ -0,0 +1,109 @@
...
+) {
+  const templateOptions = {
+    ...options,
+    ...names(options.name),
+    offsetFromRoot: offsetFromRoot(options.projectRoot),
+    template: '',
+  };
+  generateFiles(
+    tree,
+    path.join(__dirname, 'files', 'template-angular-v' + majorAngularVersion),
```
nartc, line 75: nit (non-blocking): I notice a lot of template files can be reused for both angular versions but this is non-blocking. We can have follow up PR to consolidate the templates

nartc (changes requested):

packages/platform/src/lib/nx-plugin/executors/build/executor.ts
goetzrobin: Agreed! 

My initial idea was to have analog executors for serve, build, serve-production-files that would take care of most of the paths configuration internally, and also allow us to support e2e tests based on a node server using the build output.

The way I understand [this issue](https://github.com/nrwl/nx/issues/10814) is that the executor would need to return the host and port to make the cypress executor work with it. But this could be a future PR, too. 

I would remove the executors for now!

brandonroberts:
> > moved the plugin into the `platform` package!
> 
> Is there a reason why the `plugin` needs to be part of `platform` package?

Had a discussion about this also. My thoughts:

The platform package is the "glue" for cohesive support. If it's in the platform package, we can use one package for generating + upgrading Angular CLI and Nx workspaces. 

nx migrate @analogjs/platform
ng update @analogjs/platform

Using `@analogjs/nx` makes it seem like it doesn't work outside of Nx, but it does and should. It can still be listed as an Nx plugin, as it has all the necessary configuration supported.

nartc:
It (the nx-plugin lib) doesn't necessarily have to stay in `platform` to be released together with `platform`. 

Check https://github.com/angular-threejs/angular-three where I have `plugin` and `angular-three` as two separate Nx projects. Running build in `angular-three` also builds `plugin` and the outputPath controls where `plugin` gets built to.

brandonroberts:
Ahh I see. I see things under `packages` as things that get shipped to `npm`. What other advantages are there of it being a separate package/lib?

Thoughts @goetzrobin on keeping it as separate "package" but shipping it inside of `@analogjs/platform`?

nartc:
@brandonroberts it is mainly for ergonomics as well as prevent potential circular dependency in the future where

```mermaid
graph TD;
    plugin-->package-A;
    package-A-->package-B;
    package-B--contains-->plugin;
```

I personally don't feel too strongly about it.

goetzrobin:
@brandonroberts I am open to either solution. I think the chance of the circular dependency is fairly low because the plugin has such a specific use case? Really comes down to what you prefer, you guys have much more experience publishing libraries. (I have 0 😄 )

brandonroberts:
@goetzrobin @nartc I'd say keep it as a separate package like you had before, and just distribute it as part of the `platform` package. It could simplify the builds of the two things, and let us keep the standard structure of an Nx plugin. 

You should be able to revert the last commit or hard reset the branch back to the previous commit, whichever is easier for you.

packages/platform/generators.json
goetzrobin: I will make the changes but will leave a comment for you to double check this once I push them! Haven't worked with the Angular CLI and schematics besides using them 🙈

MDyrcz5: @goetzrobin you can use a `convertNxGenerator` from `@nrwl/devkit'` I think 🤔

goetzrobin:
@brandonroberts @nartc I reset the branch to before I moved it and implemented the suggestions. I'll need some help figuring out how to package them together and what needs to happen for it to work (merge package.json or anything else I am missing.) 

Feel free to add changes/documentation as you see fit!

brandonroberts:
Nice work! We'll have to see how this works on a fresh Nx workspace. My thought is that it works here because we have all the Vite dependencies installed but a fresh workspace won't. 

We may need to run the `@nrwl/vite:init` generator before generating a new app as it takes care of installing all the Vite deps.

That can be done in a follow-up though. I'll take another look and we can get this merged in to move forward. 🎉

goetzrobin:
@brandonroberts it should install the necessary dependencies if they don't already exist when running the generator. The e2e test generates a whole new workspace in the tmp folder. 

The biggest question for me right now is how we package them together but I am sure @nartc already has an answer for that 🤓

brandonroberts:
@goetzrobin you're right. I missed it scrolling through the changes. I can take a look at the build part. Essentially, we'll have a `build-plugin` target and a build `target`. 

When we run the build, it will build the platform package, then run the build for the plugin and put it inside the folder of the built platform package

nartc:
@goetzrobin Several steps:

- First, make sure the `outputPath` of the plugin is pointing to the desired location https://github.com/angular-threejs/angular-three/blob/main/libs/plugin/project.json#L11
- Add `generators` (and `schematics`) to `platform` (aka the parent lib) package.json https://github.com/angular-threejs/angular-three/blob/main/libs/angular-three/package.json#L35
- Add a target to build both the `platform` and the plugin https://github.com/angular-threejs/angular-three/blob/main/libs/angular-three/project.json#L24. Make sure to name the target accordingly with our Nx setup.

packages/platform/generators.json
nartc: You can use `convertNxGenerator()` to do this. https://github.com/angular-threejs/angular-three/blob/main/libs/plugin/src/generators/init/compat.ts#L4
then point the `schematics` to the compat file https://github.com/angular-threejs/angular-three/blob/main/libs/plugin/generators.json#L14

packages/nx-plugin/src/generators/app/generator.ts
```diff
@@ -0,0 +1,172 @@
...
+}
+
+function addFiles(
+  tree: Tree,
+  options: NormalizedOptions,
+  majorAngularVersion: number
+) {
+  const templateOptions = {
+    ...options,
+    offsetFromRoot: options.offsetFromRoot,
```
nartc, line 131: suggestion: we can just spread `...options` right?

nartc (approved):

packages/nx-plugin/src/generators/app/generator.ts
goetzrobin: haha what in the world was I doing there 😄 thank you!

brandonroberts (approved):

brandonroberts:
Thanks all!

@allcontributors add @MDyrcz5 for code
@allcontributors add @nartc for review

brandonroberts:
Nx generator support has been released in `@analogjs/platform@0.1.0-beta.20`!

PR #310, opened 2023-03-27
feat(nx-plugin): moved add-project-config to seperate func

MDyrcz5:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] vite-angular-plugin
- [ ] astro-angular
- [ ] create-analog
- [ ] router
- [ ] platform
- [ ] content

Moved a `addProjectConfiguration` to a seperate func

Issue Number: N/A

- [ ] Yes
- [x] No

### DIFF
--- a/packages/nx-plugin/src/generators/app/lib/create-project.ts
+++ b/packages/nx-plugin/src/generators/app/lib/create-project.ts
@@ -1,0 +1,150 @@
+import { NormalizedOptions } from '../generator';
+import type { ProjectConfiguration, Tree } from '@nrwl/devkit';
+import { addProjectConfiguration } from '@nrwl/devkit';
+
+export function createProject(
+  tree: Tree,
+  normalizedOptions: NormalizedOptions
+) {
+  const projectConfiguration: ProjectConfiguration = {
+    root: normalizedOptions.projectRoot,
+    projectType: 'application',
+    sourceRoot: `${normalizedOptions.projectRoot}/src`,
+    // todo: do we want a custom executor hiding most of this?
+    // targets: {
+    //   build: {
+    //     executor: '@analogjs/nx:build',
+    //   },
+    // },
+    targets: {
+      build: {
+        executor: '@nrwl/vite:build',
+        outputs: [
+          '{options.outputPath}',
+          `dist/apps/${normalizedOptions.projectName}/.nitro`,
+          `dist/apps/${normalizedOptions.projectName}/ssr`,
+          `dist/apps/${normalizedOptions.projectName}/analog`,
+        ],
+        options: {
+          configFile: 'vite.config.ts',
+          outputPath: `dist/apps/${normalizedOptions.projectName}/client`,
+        },
+        defaultConfiguration: 'production',
+        configurations: {
+          development: {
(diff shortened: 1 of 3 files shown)


packages/nx-plugin/src/generators/app/lib/create-project.ts
```diff
@@ -0,0 +1,150 @@
+import { NormalizedOptions } from '../generator';
+import type { ProjectConfiguration, Tree } from '@nrwl/devkit';
+import { addProjectConfiguration } from '@nrwl/devkit';
+
+export function createProject(
+  tree: Tree,
+  normalizedOptions: NormalizedOptions
```
goetzrobin, line 7: Can we only pass in the properties we need from normalized options. This makes it easier to identify what this function actually depends on. We really only need: `projectRoot`,  `projectName`, and `parsedTags`

yarn.lock
goetzrobin, line 1: Let me double check if this is something I accidentally committed in the previous commit. I tested the generator locally so there is a good chance... 🙈

packages/nx-plugin/src/generators/app/lib/create-project.ts
goetzrobin, line 1: Please adjust the file name according to the name you pick :)

goetzrobin (changes requested):
Thanks for this commit! I left a couple minor remarks but you read my mind! I had planned to move this in it's own function! 👍

packages/nx-plugin/src/generators/app/lib/create-project.ts
```diff
@@ -0,0 +1,150 @@
...
+          development: {
+            buildTarget: `${normalizedOptions.projectName}:build:development`,
+            hmr: true,
+          },
+          production: {
+            buildTarget: `${normalizedOptions.projectName}:build:production`,
+          },
+        },
+      },
+      'build-ng': {
```
brandonroberts, line 60: `build-ng` was just for initial testing and should not be included

packages/nx-plugin/src/generators/app/lib/create-project.ts
```diff
@@ -0,0 +1,150 @@
...
+        executor: '@nrwl/linter:eslint',
+        outputs: ['{options.outputFile}'],
+        options: {
+          lintFilePatterns: [
+            `apps/${normalizedOptions.projectName}/**/*.ts`,
+            `apps/${normalizedOptions.projectName}/**/*.html`,
+          ],
+        },
+      },
+      test: {
```
brandonroberts, line 131: This should use the Vitest setup

MDyrcz5: Added

packages/nx-plugin/src/generators/app/lib/create-project.ts
MDyrcz5: Removed

MDyrcz5:
@goetzrobin I also made changes, in the template files so that the tests would start passing now

goetzrobin (approved):

Issue #320, opened 2023-04-05
Docs: Nx Plugin


yassernasc:
### Please provide the environment you discovered this bug in.

not quite a bug 😅 

### Which area/package is the issue in?

Docs

### Description

At `packages/nx-plugin/README.md`, instead:
> nx g @analogjs/**nx**:app analog-app

The correct command seems to be 
> nx g @analogjs/**platform**:app analog-app

As well, we can create a new page inside the Integration section at the official docs with the `packages/nx-plugin/README.md` content.

### I would be willing to submit a PR to fix this issue

- [X] Yes
- [ ] No

brandonroberts:
Thanks @yassernasc.  That fix would be good. I'd like to have a more detailed docs page for usage with Nx under the Integrations section, so I think it would be its own document of how to:

- Install the package
- Run the generator for a new application
- Run the nx build, nx test commands
- Generate an E2E application (future)

brandonroberts:
I think its ok to test out with. I ran into an issue with a completely empty workspace also. There is no `tsconfig.base.json` generated with an empty workspace which caused an error when trying to generate a new app.

These steps worked for me:

1. npx create-nx-workspace@latest my-workspace (Choose integrated)
2. (Choose angular workspace preset)
3. npm i @analogjs/platform --save-dev
4. nx g @analogjs/platform:app analog-app
5. nx build analog-app
6. nx test analog-app

brandonroberts:
We should support starting from an empty workspace with no preset also @goetzrobin

brandonroberts:
Not quite yet. Nx is for "stable" projects, and we aren't quite there yet version-wise.

goetzrobin:
I'll give this some love the next couple days/weeks and make sure you can get analog up and running in Nx with a single step

DonaldMurillo:
> I think its ok to test out with. I ran into an issue with a completely empty workspace also. There is no tsconfig.base.json generated with an empty workspace which caused an error when trying to generate a new app.
> These steps worked for me:
> npx create-nx-workspace@latest my-workspace (Choose integrated)
(Choose angular workspace preset)
npm i @analogjs/platform --save-dev
nx g @analogjs/platform:app analog-app
nx build analog-app
nx test analog-app

I tried the steps mentioned above but got the following error
![image](https://user-images.githubusercontent.com/54650861/235367322-fe4b5156-9913-4b7c-bde2-053a8da76c9f.png)
[...]

PR #370, opened 2023-05-04
feat: make nx plugin work without angular preset and add tailwind option

goetzrobin:
The nx plugin now works without any manual addition of @nx/angular. You can create an empty workspace and simply add the @analogjs/platform dependency. After installing it you can run one command: nx g @analogjs/platform:app. This will automatically install all angular and analog dependencies. This plugin should also work as a preset for npx create-workspace. Since e2e testing for nx plugins is still a bit messy I tested this manually for now. It worked as expected for an empty package based and an empty integrated workspace as well as for an already preconfigured angular workspace.

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] vite-plugin-angular
- [ ] vite-plugin-nitro
- [ ] astro-angular
- [ ] create-analog
- [ ] router
- [x] platform
- [ ] content

Nx plugin required you to have your workspace already configured with angular.

Issue Number: #320

## What is the new behavior?

This PR makes the analog plugin take care of all the config. It also makes the Nx plugin
compatible with npx create-nx-workspace presets. (need to verify this after publishing to npm)
It also adds support for setting up tailwind

- [ ] Yes
- [x] No

## Other information

I wouldn't consider this completing #320 because we still need to update the Docs and integrate a SPARTAN setup.
I'd like to have all this working before writing the docs to make them be in sync from the get go :)

### DIFF
--- a/packages/nx-plugin/src/generators/app/generator.ts
+++ b/packages/nx-plugin/src/generators/app/generator.ts
@@ -7,7 +7,6 @@ import {
   stripIndents,
   Tree,
 } from '@nx/devkit';
-import * as path from 'path';
 import { AnalogNxApplicationGeneratorOptions } from './schema';
 import { major } from 'semver';
 import { getInstalledPackageVersion } from '../../utils/version-utils';
--- a/packages/nx-plugin/src/generators/app/generator.ts
+++ b/packages/nx-plugin/src/generators/app/generator.ts
@@ -1,7 +1,5 @@
 import {
-  addDependenciesToPackageJson,
   formatFiles,
-  generateFiles,
   getWorkspaceLayout,
   installPackagesTask,
   names,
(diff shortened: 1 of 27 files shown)


packages/nx-plugin/src/generators/app/generator.ts
```diff
@@ -51,24 +25,25 @@ export interface NormalizedOptions
...
+  const projectDirectory = allNames.fileName;
   const projectName = projectDirectory.replace(new RegExp('/', 'g'), '-');
-  const projectRoot = `${getWorkspaceLayout(tree).appsDir}/${projectDirectory}`;
+  const projectRoot = `${appsDir}/${projectDirectory}`;
   const parsedTags = options.tags
     ? options.tags.split(',').map((s) => s.trim())
     : [];
   const offsetFromRoot = determineOffsetFromRoot(projectRoot);
-
+  const nxPackageNamespace = major(nxVersion) >= 16 ? '@nx' : '@nrwl';
```
brandonroberts, line 46: Let's just use `@nx` as we'll pin everything to Nx 16.1.0 to support Angular >= v15

packages/nx-plugin/src/generators/app/versions.ts
```diff
@@ -1,5 +1,9 @@
+export const MINIMUM_SUPPORTED_ANGULAR_VERSION = '15.0.0';
 // V16
 // dependencies
+export const V16_ANGULAR = '16.0.0';
+export const V16_NX_DEVKIT = '16.0.3';
```
brandonroberts, line 5: Pin all Nx deps to ^16.0.0 so it will pick up the latest release

```suggestion
export const V16_NX_DEVKIT = '^16.0.0';
```

packages/nx-plugin/src/generators/preset/generator.spec.ts
```diff
@@ -0,0 +1,28 @@
+import { createTreeWithEmptyWorkspace } from '@nrwl/devkit/testing';
+import {
+  Tree,
+  readProjectConfiguration,
+  addDependenciesToPackageJson,
+} from '@nrwl/devkit';
```
brandonroberts, line 6:
```suggestion
} from '@nx/devkit';
```

packages/platform/package.json
```diff
@@ -26,7 +26,8 @@
   "dependencies": {
     "nitropack": "^1.0.0",
     "@analogjs/vite-plugin-angular": "^0.2.0-beta.3",
-    "@analogjs/vite-plugin-nitro": "^0.2.0-beta.3"
+    "@analogjs/vite-plugin-nitro": "^0.2.0-beta.3",
+    "@nx/devkit": "16.0.1"
```
brandonroberts, line 30:
```suggestion
    "@nx/devkit": "^16.0.0"
```

packages/nx-plugin/src/generators/app/generator.ts
goetzrobin: The one thing I want to mention is that this is version comes from the user's Nx workspace. If I understand this correctly removing this logic would mean that we do not support workspaces that have an Nx version of 15

brandonroberts: Good point. Let's keep them separate

packages/nx-plugin/src/generators/app/versions.ts
goetzrobin: Should be addressed now

packages/nx-plugin/src/generators/preset/generator.spec.ts
goetzrobin: Fixed

packages/platform/package.json
goetzrobin: Changed 👍

packages/nx-plugin/src/generators/app/generator.ts
goetzrobin: I tested with a v15.9.2 workspace and the current  approach & versions for V15 worked 👍

brandonroberts (approved):

PR #374, opened 2023-05-05
feat(vite-plugin-angular): add support for JIT mode

### LINKED ISSUE #368: Feature: Add support for JIT mode for testing

vite-plugin-angular

### Information

Currently when running tests, the TypeScript files are compiled with AOT using the plugin. Angular compilation supports JIT mode also which could be faster in terms of compilation, but would require the `@angular/compiler` to be imported.

This would add an option to the plugin for `enableJit` that is enabled when `process.env['NODE_ENV'] === 'test'`

_No response_

- [ ] Yes
- [ ] No

brandonroberts:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [x] vite-plugin-angular
- [ ] vite-plugin-nitro
- [ ] astro-angular
- [ ] create-analog
- [ ] router
- [ ] platform
- [ ] content
- [ ] nx-plugin
- [ ] trpc

Closes #368

## What is the new behavior?

Enabling JIT mode through the Analog Platform and Vite Angular Plugin is an option.

```js
/// <reference types="vitest" />

import { offsetFromRoot } from '@nx/devkit';
import angular from '@analogjs/vite-plugin-angular';

// https://vitejs.dev/config/
export default defineConfig(({ mode }) => {
  return {
    root: 'src',
    plugins: [
      angular({ jit: boolean }), // enabled automatically when running tests with Vitest
    ],
    test: {
      globals: true,
      environment: 'jsdom',
      setupFiles: ['src/test-setup.ts'],
      include: ['**/*.spec.ts'],
    },
    define: {
      'import.meta.vitest': mode !== 'production',
    },
  };
});
```

When running tests, JIT mode is used instead of AOT which provides faster compilation of files, while still supporting internal and external styles and templates. This will speed up compilation time with libraries such as Vitest, Playwright, Storybook, etc.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/platform/src/lib/platform-plugin.ts
+++ b/packages/platform/src/lib/platform-plugin.ts
@@ -31,6 +31,6 @@ export function platformPlugin(opts: Options = {}): Plugin[] {
     (platformOptions.ssr
       ? devServerPlugin({ entryServer: opts.entryServer })
       : false) as Plugin,
-    ...angular(opts?.vite),
+    ...angular({ jit: platformOptions.jit, ...opts?.vite }),
   ];
 }
--- a/packages/platform/src/lib/platform-plugin.ts
+++ b/packages/platform/src/lib/platform-plugin.ts
@@ -31,6 +31,6 @@ export function platformPlugin(opts: Options = {}): Plugin[] {
     (platformOptions.ssr
       ? devServerPlugin({ entryServer: opts.entryServer })
       : false) as Plugin,
-    ...angular({ jit: platformOptions.jit, ...opts?.vite }),
+    ...angular({ jit: platformOptions.jit, ...(opts?.vite ?? {}) }),
   ];
 }
--- a/packages/vite-plugin-angular/src/lib/angular-jit-plugin.ts
+++ b/packages/vite-plugin-angular/src/lib/angular-jit-plugin.ts
@@ -1,0 +1,51 @@
+import { Plugin, PluginContainer } from 'vite';
+import { readFileSync } from 'fs';
+
+export function jitPlugin({
+  styleTransform,
+  inlineStylesExtension,
+}: {
+  styleTransform: PluginContainer['transform'];
+  inlineStylesExtension: string;
+}): Plugin {
+  return {
+    name: '@analogjs/vite-plugin-angular-jit',
+    resolveId(id: string) {
+      if (id.startsWith('virtual:angular')) {
+        return `\0${id}`;
+      }
+
+      return;
+    },
+    async load(id: string) {
+      if (id.includes('virtual:angular:jit:template:file;')) {
+        const contents = readFileSync(id.split('file;')[1], 'utf-8');
+
+        return `export default \`${contents}\`;`;
+      } else if (id.includes('virtual:angular:jit:style:inline;')) {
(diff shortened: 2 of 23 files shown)


brandonroberts:
@sand4rt this should help with Playwright compilation speed also.

sand4rt:
@brandonroberts thanks! Gonna try this asap!

packages/platform/src/lib/platform-plugin.ts
```diff
@@ -31,6 +31,6 @@ export function platformPlugin(opts: Options = {}): Plugin[] {
     (platformOptions.ssr
       ? devServerPlugin({ entryServer: opts.entryServer })
       : false) as Plugin,
-    ...angular(opts?.vite),
+    ...angular({ jit: platformOptions.jit, ...opts?.vite }),
```
yharaskrik, line 34: If `opts?.vite` ends up being undefined this spread will error. I think you want `...(opts?.vite ?? {})`?

brandonroberts: Good catch, thanks

goetzrobin (approved):

packages/vite-plugin-angular/src/lib/angular-jit-plugin.ts
```diff
@@ -0,0 +1,69 @@
+import { Plugin, PluginContainer, ViteDevServer } from 'vite';
+import { readFileSync } from 'fs';
+
+export function jitPlugin({
+  inlineStylesExtension,
+}: {
+  inlineStylesExtension: string;
+}): Plugin {
+  let styleTransform: PluginContainer['transform'] | undefined;
```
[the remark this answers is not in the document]
marcus-sa: @gergobergo the correct TS solution is already being used. There's no point in abstracting something that doesn't need to be abstracted.

sand4rt:
@brandonroberts thought the `templateUrl` would still work with `angular({ jit: true })` somehow but unfortunately that is not the case. I wonder if this is still a useful feature for testing because of this?

brandonroberts:
@sand4rt it supports `templateUrl` and `styleUrls` with Vitests. I'll have to check and see what the issue is with Playwright. Will you open a new issue with a sample repo?

KlausEverWalkingDev:
After some tests I came with these weirdly conclusions:
image

brandonroberts:
@KlausEverWalkingDev this is the same as disabling `jit` mode altogether. The `vite` object overrides the `analog` plugin options for the Vite plugin.

KlausEverWalkingDev:
Oh, thanks. I added it while fiddling with analog source file, but I admit I didn't understand the real meaning of it. By the way, what's the difference between analog and angular plugin? I didn't get when to use only one of them or both.

brandonroberts:
> Oh, thanks. I added it while reading the analog source files, but I admit I didn't understand the real meaning of it. By the way, what's the difference between analog and angular plugin? I didn't get when to use only one of them or both.

The analog plugin is the "glue" between the analog packages.

The angular plugin is what handles connecting Angular compilation to Vite. It can be used on its own and is used for Astro, Qwik, and other Vite integration

PR #382, opened 2023-05-10
feat(platform): improve nx plugin and add tRPC support

goetzrobin:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] vite-plugin-angular
- [ ] vite-plugin-nitro
- [ ] astro-angular
- [ ] create-analog
- [ ] router
- [ ] platform
- [ ] content
- [x] nx-plugin
- [x] trpc

So far you had to manually install and integrate the tRPC client.

Issue Number: N/A

## What is the new behavior?

This adds a prompt that allows you to
configure the tRPC client out of the box.

- [ ] Yes
- [x] No

## Other information

The next step is to come up with much better looking templates for the starters.
Things to consider are:
- App with Tailwind, no tRPC
- App with Tailwind + tRPC
- App without Tailwind, no tRPC
- App without Tailwind + tRPC

Also worth adding might be the blog starter


packages/nx-plugin/src/generators/app/files/template-angular-v16/src/app.config.server.ts__template__
```diff
@@ -0,0 +1,10 @@
+import { ApplicationConfig, mergeApplicationConfig } from '@angular/core';
+import { provideServerRendering } from '@angular/platform-server';
+import { provideClientHydration } from '@angular/platform-browser';
+import { appConfig } from './app.config';
+
+const serverConfig: ApplicationConfig = {
+  providers: [provideServerRendering(), provideClientHydration()],
```
brandonroberts, line 7: provideClientHydration() is already included in the app.config.ts. It doesn't need to be included here

```suggestion
  providers: [provideServerRendering()],
```

packages/nx-plugin/src/generators/app/lib/add-analog-project-config.ts
```diff
@@ -43,7 +43,7 @@ export function addAnalogProjectConfig(
         defaultConfiguration: 'development',
         options: {
           buildTarget: `${projectName}:build`,
-          port: 4200,
+          port: 4205,
```
brandonroberts, line 46: This should remain `4200`

```suggestion
          port: 4200,
```

packages/nx-plugin/src/generators/app/files/template-angular-v16/src/wait-for.ts__template__
```diff
@@ -0,0 +1,19 @@
+import { firstValueFrom, isObservable, Observable } from 'rxjs';
```
brandonroberts, line 1: Should this be moved into the `@analogjs/trpc` package?

brandonroberts: Also, this file should not be generated if tRPC is not added

brandonroberts (changes requested):

brandonroberts:
Lots of great changes! Some thoughts

- The generated app looks great but needs more color.
- I think it would look even better possibly as the homepage for analogjs.org
- Let's keep the generated app simple so there's less to do to customize. Maybe just have Analog logo, docs link, and the todo list for the tRPC app, and the counter for the regular app.

brandonroberts:
Also, I think branching into two sets of template files would be simpler then lots of if/else weaved into the generator templates

packages/nx-plugin/src/generators/app/files/template-angular-v16/src/wait-for.ts__template__
goetzrobin: I am fine with moving it to the `@analog/trpc` package. Or should we move it to `@analogjs/platform` since it might also be helpful for people using the http client

packages/nx-plugin/src/generators/app/files/template-angular-v16/src/app.config.server.ts__template__
goetzrobin: fixed

packages/nx-plugin/src/generators/app/lib/add-analog-project-config.ts
goetzrobin: fixed

packages/nx-plugin/src/generators/app/files/template-angular-v16/src/wait-for.ts__template__
goetzrobin: moved wait for utility to trpc package

goetzrobin:
@brandonroberts I removed the features and open-source sections of each template.
I think you are right that they would be nice on the home page but add unnecessary code for the template.

I added some color and the logo at the top of the page.

I moved each page in its own template instead of doing the if/else logic inside a single one.

Thanks for the feedback and let me know if it looks good now 😃

brandonroberts (approved):

PR #441, opened 2023-05-26
fix(trpc): allow to pass custom headers to trpc client

### LINKED ISSUE #394: Allow for tRPC authentication through headers/custom fetch, or whatever makes sense

Don't known / other

### Information

Currently the tRPC client does not allow you to pass in custom headers or a custom fetch function to support authentication of the request through e.g. an JWT

Don’t use auth 😔

- [X] Yes
- [ ] No

goetzrobin:
A tRPCHeaders signal is exposed to set/get the headers added to each tRPC call. This allows us to, e.g. add an authorization header that we can then use to create protected procedures that can only be executed when the user is authorized.

closes #394

## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] vite-plugin-angular
- [ ] vite-plugin-nitro
- [ ] astro-angular
- [ ] create-analog
- [ ] router
- [ ] platform
- [ ] content
- [ ] nx-plugin
- [x] trpc

Closes #394 

## What is the new behavior?

You can provide custom headers to the trpc client by setting the tRPCHeaders signal

- [ ] Yes
- [x] No

## Other information

It works but I feel like the way we configure the tRPC client could be much better.
I think we should aim for a way to configure the tRPC client like @markostanimirovic allows the ngrx signal store to be configured. I want to get this merged because it is essential for anyone who wants to use the client for anything that is 
not public and the DX is alright but we should definitely rethink the way we allow people to configure the tRPC client.

```ts
const { tRPCClient, provideTRPCClient, tRPCHeaders } = createTrpcClient(
  withTransformer(superjson),
  withLinks([yourCustomLink()]),
  withHttpBatchLink(
      withURL(''),
      withHeaders()
  )
);
```

### DIFF
--- a/apps/trpc-app-e2e-playwright/tests/app.spec.ts
+++ b/apps/trpc-app-e2e-playwright/tests/app.spec.ts
@@ -43,14 +43,25 @@ describe('tRPC Demo App', () => {
     ).toContain(/Analog + tRPC/i);
   });
 
-  test<TRPCTestContext>(`If user enters the first note the note should be stored
-    successfully and listed in the notes array`, async (ctx) => {
+  test<TRPCTestContext>(`
+  If user enters the first note the note should be storedsuccessfully and listed in the notes array.
+  Still unauthorized the user should not be able to delete the note and the error should be displayed.
+  After the users clicks the Login button and gets authorized, deleting the note again should work successfully,
+  and the error should disappear.
+     `, async (ctx) => {
     await ctx.notesPage.typeNote(notes.first.note);
 
     await ctx.notesPage.addNote();
     expect(await ctx.notesPage.notes().elementHandles()).toHaveLength(1);
 
     await ctx.notesPage.removeNote(0);
+    expect(await ctx.notesPage.notes().elementHandles()).toHaveLength(1);
+    expect(await ctx.notesPage.getDeleteErrorCount()).toBe(1);
+
+    await ctx.notesPage.toggleLogin();
+    await ctx.notesPage.removeNote(0);
+    await page.waitForSelector('.no-notes');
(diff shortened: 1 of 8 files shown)


goetzrobin:
@brandonroberts @Dafnik @markostanimirovic please take a look at the `Other information` section and let me know what you guys think. I might need some help from you guys as I took a look at https://github.com/markostanimirovic/ngrx-signal-store-playground/blob/main/src/lib/signal-store.ts and there's definitely a lot going on that I don't understand yet!

brandonroberts (approved):

brandonroberts:
@goetzrobin its worth a look. I don't think we need to build an abstraction on top of tRPC configuration if we give them direct access to configure it already. If we don't maybe we should consider that first and not make this too Angular-specific.

markostanimirovic:
Hey guys 👋 

I think that function composition would not bring many benefits in this case.

SignalStore APIs are designed in a composable way mostly to allow building features that can be reused in multiple stores, overcome limitations of the class-based state mgmt solutions, and enable tree-shaking of unused stuff.

In the case of the `createTrpcClient` function, I'd keep the current signature. It looks simple and intuitive. 👌

goetzrobin:
Yeah I might have jumped ahead to some premature optimization on that one. I was just trying to think ahead of we'd ever allow for a Web-socket link and all that. Maybe the way to go is what you suggested @brandonroberts: to expose more of the config directly

brandonroberts:
> Yeah I might have jumped ahead to some premature optimization on that one. I was just trying to think ahead of we'd ever allow for a Web-socket link and all that. Maybe the way to go is what you suggested @brandonroberts: to expose more of the config directly

Its less for you to maintain in the long run 🙂

PR #481, opened 2023-06-25
feat(nx-plugin): fix angular 15.x and 16.x install support

jeremyhofer:
## PR Checklist

## PR Type

- [x] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] vite-plugin-angular
- [ ] vite-plugin-nitro
- [ ] astro-angular
- [ ] create-analog
- [ ] router
- [ ] platform
- [ ] content
- [x] nx-plugin
- [ ] trpc

Before this update the plugin would force the latest Angular version to be used within a Nx workspace regardless of the Angular version(s) supported by the user's current version of Nx. This led to some inconsistencies at time with the dependency versions used and installed, causing various install, build, serve, and similar issues.

This also addresses some installation and build related issues due to the `@nrwl` to `@nx` scope transition, and `@angular-devkit/build-angular` internal project path changes.

Potentially addresses https://github.com/analogjs/analog/issues/486 which was recently opened.

## What is the new behavior?

This version will install consistent dependency versions for Analog and Angular.

For Nx workspaces on Nx versions older than 15.2.0 installation will not be supported as Nx did not add Angular 15 support until 15.2.0.

For Nx workspaces on Nx versions 15.2.0 through 16.0.0 the latest 15.X version of Angular will be installed (currently ~15.2.0 as of this PR). The Analog dependency versions that work for Angular 15.X are installed.

For Nx workspaces on Nx version 16.1.0 and newer the latest 16.X version of Angular will be installed (currently ~16.1.0 as of this PR). Latest Analog dependency versions are installed as Angular 16 is the current major version.

- [ ] Yes
- [X] No


goetzrobin (changes requested):
I tested it locally and we need one more change to make it work for Nx 15:

Where we add the analog dependencies we also need to remove the initially installed
platform version. This is kind of counter intuitive, but we are basically replacing the latest
platform with the one that works for the Nx workspace version. If we don't do that the newest plugin will kept and we get a file path error. Making this changes will take care of the file
changes in the angular builder.

```ts
import { addDependenciesToPackageJson, removeDependenciesFromPackageJson, Tree } from '@nx/devkit';
import { getAnalogDependencies } from '../versions/dependencies';
import { getAnalogDevDependencies } from '../versions/dev-dependencies';

export async function addAnalogDependencies(
  tree: Tree,
  nxVersion: string,
  angularVersion: string
) {
  const dependencies = getAnalogDependencies(nxVersion, angularVersion);
  const devDependencies = getAnalogDevDependencies(nxVersion);
  removeDependenciesFromPackageJson(tree, [], ['@analogjs/platform'])
  addDependenciesToPackageJson(tree, dependencies, devDependencies);
}
```

Another thing I noticed is that I don't believe Nx follows their own version matrix. Once I have Nx 16.0.0 Angular 16.1 gets installed (with a brand new workspace)

goetzrobin (approved):
Very excited for this to land! 🚀 🔥

goetzrobin:
@brandonroberts code changes look good from my side and locally tests pass for me. not sure why CI is timing out :(

brandonroberts (approved):

brandonroberts:
Awesome work! @allcontributors add @jeremyhofer for code, infra

PR #496, opened 2023-07-02
feat(content): add support for using slug from markdown frontmatter

### LINKED ISSUE #488: Add Ability to Set the Browser's URL for a Content Post

content

### Information

The path for a blog post in `src/content/blog` is the filename that comes after `/blog`. So, `src/content/blog/2023-06-27-test.md` generates a path of `www.mysite.com/blog/2023-06-27-test`. 

However, I have an attribute in the frontmatter of my blog posts called custom_slug. I need that slug to be used for the path in the browser's URL. I was able to get that working, basically, but the file could not be found because the custom slug doesn't match the file path to any file.

I created a branch to demo what I was thinking could work, but I'm not sure if it's correct. Here's a link to it:

https://github.com/pjlamb12/analog/commit/80e04b47949fa3942bd8326395f81587885483e6

The other option is to allow you to pass a parameter for the slug (in my case I passed `custom_slug` and it found the correct attribute) and then also a filename parameter so that the default slug can be passed and the file can be found.

- [X] Yes
- [ ] No

pjlamb12:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] vite-plugin-angular
- [ ] vite-plugin-nitro
- [ ] astro-angular
- [ ] create-analog
- [ ] router
- [ ] platform
- [x] content
- [ ] nx-plugin
- [ ] trpc

The slug for the blog content page (like a blog post) is always the filename (minus the md extension)

Closes #488

## What is the new behavior?
You can provide a custom slug attribute for posts so that you can define the slug in the frontmatter of the markdown file

- [ ] Yes
- [x] No

https://media.giphy.com/media/l41Ye5dhLPqILtT2w/giphy.gif


brandonroberts:
Thanks @pjlamb12. The solution you have does work, but it's somewhat less efficient because you have to loop through all the files to find the custom slug for the filename, whereas currently, it's a 1-to-1 lookup based on the filename. 

Maybe we can build the list using the custom slugs earlier in here [CONTENT_FILES_LIST_TOKEN](https://github.com/analogjs/analog/blob/main/packages/content/src/lib/content-files-list-token.ts)

And build the dictionary of slugs to files earlier here

https://github.com/analogjs/analog/blob/main/packages/content/src/lib/content-files-token.ts

We could build a map of filename -> content and merge it with custom_slug -> content and keep the same 1-to-1 lookup.

Thoughts?

pjlamb12:
I have no strong feelings about anything I've done so if there are better ways I'll definitely look into how to improve it.

I'll look at it and see if I can put it together the way you mentioned.

brandonroberts:
I think a first step could be setting the `slug` to the one provided in the markdown file here if it exists.

https://github.com/analogjs/analog/blob/main/packages/content/src/lib/content-files-list-token.ts#L24

The other step is more involved. We could use the slug lookup as a fallback in case the filename lookup doesn't return anything from the file map returned by Vite. I'll think some more on that side also

pjlamb12:
I asked Bard for ideas, and came up with the following:

> Sure, here is the JavaScript code that you can use to change the keys of the object to use the slug attribute from the array where the filename matches the key from the object:

```js
const object = {
  "filename1": "file1.txt",
  "filename2": "file2.txt",
  "filename3": "file3.txt",
};

const array = [
  {
    "filename": "file1.txt",
    "slug": "file-1-txt",
  },
  {
    "filename": "file2.txt",
    "slug": "file-2-txt",
  },
  {
    "filename": "file3.txt",
    "slug": "file-3-txt",
  },
];

const lookup = {};
array.forEach((item) => {
  lookup[item.filename] = item.slug;
});

const newObject = {};
Object.entries(object).forEach((entry) => {
  const filename = entry[0];
  const value = entry[1];

  const slug = lookup[filename];
  if (slug !== undefined) {
    newObject[slug] = value;
  }
});

console.log(newObject);
```

> This code will first create a lookup object that maps each filename to its corresponding slug. Then, it will iterate over the object and the array, and for each entry, it will check if the filename from the object matches the filename from the array. If it does, it will use the slug from the array as the new key for the value in the object. Finally, it will print the new object to the console.

> This code is efficient because it only iterates over the object and the array once. It also uses the lookup object to avoid having to search the entire array for each filename.

I then asked it about the efficiency:

> The above code is relatively efficient. The main bottleneck is the Object.entries() method, which iterates over the object and returns an array of key-value pairs. However, this method is typically optimized by JavaScript engines, so the performance impact should be minimal.

> The other part of the code that could be optimized is the lookup object. This object is created by iterating over the array, which can be slow if the array is large. However, this object can be reused, so the performance impact should be minimal if the code is called multiple times with the same array.

> Overall, the above code is efficient and should not cause any performance problems in most cases. However, if you are working with a large object or array, you may want to consider using a different approach, such as using a hash table or a dictionary.

> Here are some other possible approaches that you could use to change the keys of the object:

* You could use the Object.assign() method to create a new object with the new keys. This method is relatively efficient, but it does not preserve the original object.
* You could use the for...in loop to iterate over the object and change the keys one by one. This method is not as efficient as the Object.entries() method, but it is easier to understand.
* You could use the Map object to create a mapping between the old keys and the new keys. This object is very efficient for lookup operations, but it can be more complex to use than other approaches.
> The best approach to use will depend on your specific needs and requirements. If you need to change the keys of an object quickly and efficiently, the Object.entries() method is a good option. If you need to preserve the original object, the Object.assign() method is a good option. If you need to perform lookup operations on the new keys frequently, the Map object is a good option.

Based on that feedback, I took a stab at implementing it in the way you can see from the latest commit.

The only thing I couldn't solve with this method was returning the real `filename`. But, other than that, I confirmed that it worked by using a custom slug attribute, or the default slug (the filename without the extension).

brandonroberts:
Nice. The returned `filename` can be the `slug` used for the lookup. I left some feedback also about simplifying the custom slug

brandonroberts (changes requested):

packages/content/src/lib/content-files-token.ts
```diff
@@ -1,14 +1,32 @@
...
+    });
+
+    const newObject: Record<string, () => Promise<string>> = {};
+    Object.entries(contentFiles).forEach((entry) => {
+      const filename = entry[0];
+      const value = entry[1];
+
+      const slug = lookup[filename];
+      if (slug !== undefined) {
+        newObject[slug] = value;
```
brandonroberts, line 26: Should we check for collisions? Just a thought that can be done later

pjlamb12:
Sounds good. I should have some time tonight that I can work on these revisions.

pjlamb12:
Alright I got those things implemented that you mentioned above, and it is still working. If this looks good, I'll also look at the tests and update them as needed.

brandonroberts (changes requested):

packages/content/src/lib/content-files-token.ts
```diff
@@ -1,14 +1,32 @@
...
 >('@analogjs/content Content Files', {
   providedIn: 'root',
   factory() {
     const contentFiles = getContentFiles();
+    const contentFilesList = inject(CONTENT_FILES_LIST_TOKEN);
 
-    return contentFiles;
+    const lookup: Record<string, string> = {};
+    contentFilesList.forEach((item) => {
+      lookup[item.filename] = item.slug || item.filename;
```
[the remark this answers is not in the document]
pjlamb12: So, I'm a little confused on what we should return for the content files then. I reverted this back to what's in the main branch and it can't find the content for the page any more when using the `slug` from the frontmatter.

packages/content/src/lib/content.ts
```diff
@@ -19,23 +19,28 @@ import { parseRawContentFile } from './parse-raw-content-file';
...
 ): Observable<ContentFile<Attributes | Record<string, never>>> {
   const route = inject(ActivatedRoute);
   const contentFiles = inject(CONTENT_FILES_TOKEN);
-  const prefix = typeof param === 'string' ? '' : `${param.subdirectory}/`;
+
   const paramKey = typeof param === 'string' ? param : param.param;
   return route.paramMap.pipe(
     map((params) => params.get(paramKey)),
     switchMap((slug) => {
-      const filename = `/src/content/${prefix}${slug}.md`;
```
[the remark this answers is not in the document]
pjlamb12: Where are we matching up the filename and the `slug` so that the file is found when using the `slug` for the route?

brandonroberts:
@pjlamb12 to provide some clarity on the slug vs filename, the `filename` is a full path to the file, while the slug is just the param. So the structure you used earlier looks more like this:

```js
const object = {
  "/Users/username/projects/analog-app/src/content/filename1.txt": "file1.txt",
  "/Users/username/projects/analog-app/src/content/filename2.txt": "file2.txt",
  "/Users/username/projects/analog-app/src/content/filename3.txt": "file3.txt"
};

const array = [
  {
    "filename": "/Users/username/projects/analog-app/src/content/file1.txt",
    "slug": "file-1-txt",
  },
  {
    "filename": "/Users/username/projects/analog-app/src/content/file2.txt",
    "slug": "file-2-txt",
  },
  {
    "filename": "/Users/username/projects/analog-app/src/content/file3.txt",
    "slug": "file-3-txt",
  },
];

const contentFiles = {
  "/Users/username/projects/analog-app/src/content/filename1.txt": () => Promise<string>,
  "/Users/username/projects/analog-app/src/content/filename2.txt": () => Promise<string>,
  "/Users/username/projects/analog-app/src/content/filename3.txt": () => Promise<string>
};
```

So that's how if you use the `slug`, you lose the rest of the filename, which is why `contentFiles[slug]` won't work.

pjlamb12:
I wasn't able to get to this today but hopefully will tomorrow.

pjlamb12:
@brandonroberts I think I got this implemented correctly...at least it's working again.

packages/content/src/lib/content-files-list-token.ts
```diff
@@ -17,11 +16,12 @@ export const CONTENT_FILES_LIST_TOKEN = new InjectionToken<ContentFile[]>(
 
       return Object.keys(contentFiles).map((filename) => {
         const attributes = contentFiles[filename];
+        const slug = attributes['slug'];
 
         return {
           filename,
           attributes,
-          slug: encodeURI(getSlug(filename)),
+          slug: encodeURI(slug) ?? getSlug(filename),
```
brandonroberts, line 24: This should fix the unit and e2e tests

```suggestion
          slug: slug ? encodeURI(slug) : encodeURI(getSlug(filename)),
```

apps/blog-app/src/content/2022-12-27-my-first-post.md
```diff
@@ -1,6 +1,6 @@
 ---
 title: My First Post
-slug: 2022-12-27-my-first-post
+slug: my-first-post
```
brandonroberts, line 3: Let's keep one with the slug the same as the filename

```suggestion
slug: 2022-12-27-my-first-post
```

packages/content/src/lib/content.ts
```diff
@@ -35,7 +41,7 @@ export function injectContent<
       if (!contentFile) {
         return of({
           attributes: {},
-          filename,
+          filename: filename || '',
```
brandonroberts, line 44: Revert

```suggestion
          filename,
```

packages/content/src/lib/content.ts
```diff
@@ -57,12 +63,14 @@ export function injectContent<
         const { content, attributes } =
           parseRawContentFile<Attributes>(rawContentFile);
 
-        return {
+        const returnObj = {
```
brandonroberts, line 66: Revert

brandonroberts (approved):
Minor tweaks, everything else looks good!

pjlamb12:
Sounds good! Got the last set of edits pushed.

brandonroberts:
Awesome, thanks @pjlamb12!

pjlamb12:
@brandonroberts do you know when this will be published by chance?

brandonroberts:
Just published under the `0.2.0-beta.25` release https://github.com/analogjs/analog/commit/99eecb712504c81dc5429c295f9b1ddd94af42a8

PR #533, opened 2023-07-09
fix(create-analog): restrain node 16 to `16.17`.

### LINKED ISSUE #532: Fix: engines in create-analog should be updated to reflect compatible versions of Node

create-analog

### Information

We have known working versions of Node that are compatible when creating new Analog projects. Currently in the `engines` property in the `package.json` we allow for incompatible versions to be used.

https://github.com/analogjs/analog/blob/main/packages/create-analog/package.json#L17-L19

The `engines` should be updated to:

```json
  "engines": {
    "node": "~16.17.0 || >=18.13.0"
  },
```

_No response_

- [ ] Yes
- [ ] No

JeanMeche:
Using `16.18` & `16.19` breaks Analog, let's retrains the version to the latest working Node16: `16.17`

Fixes #532

### DIFF
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "prettify": "prettier --write ."
   },
   "engines": {
-    "node": "^16.14.0 || >=18.10.0",
+    "node": "~16.17.0 || >=18.13.0",
     "pnpm": "^8.0.0"
   },
   "packageManager": "pnpm@8.0.0",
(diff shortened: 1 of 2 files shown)


brandonroberts:
This one should be updated also for the create-analog package.

https://github.com/analogjs/analog/blob/main/packages/create-analog/package.json

brandonroberts (approved):

brandonroberts:
@allcontributors add @JeanMeche for infra

Issue #517, opened 2023-07-06
RFC: Switch to Nx for create-analog package


brandonroberts:
We are basically maintaining two separate templating systems at this point. Idea. What do you think about switching `create-analog` to basically be a wrapper around:

`npx create-nx-workspace --preset=@analogjs/platform`

Pros
- Simplifies maintenance
- Takes advantage of standalone Nx CLI project.json structure for Angular apps
- Works the same inside/outside of Nx monorepo

Potential Cons
- It would move us off the Angular CLI and `angular.json`
- Lose `ng update`, `ng add`

Update:
Nx allows you to create your own CLI that can do both
            
_Originally posted by @brandonroberts in https://github.com/analogjs/analog/issues/388#issuecomment-1546143623_

brandonroberts:
> I'm probably only going to use analog with nx so functionally it's pure benefit. But if it were in a work project, id be cautious of a coupling to a non-strictly-necessary dependency :)

Which is the `non-strictly-necessary` dependency? Nx?

luishcastroc:
I would probably embrace NX if the focus of Analog will be to be as versatile as possible, if the idea is the framework to be used in both simple and large projects NX is for sure the best for it's extensibility.

justinrassier:
I am a bit of an Nx fan, so I am biased,  but I say the upsides of embracing Nx would outweigh the downsides for sure.

mihajm:
> > I'm probably only going to use analog with nx so functionally it's pure benefit. But if it were in a work project, id be cautious of a coupling to a non-strictly-necessary dependency :)
> 
> Which is the `non-strictly-necessary` dependency? Nx?

Yup nx :) the way I see it as an angular meta-framework that one is very obviously required whereas nx as a monorepo tool is an addition to the stack that benefits dx but might not be what the user wants, especially if alternatives such as turborepo take off.

I could be misunderstanding the original question tho & as said, love nx myself and will pretty much always use it for analog/angular projects :)

tbogard:
I would like to see this happening. Nx is robust enough to handle the needed generators for the analog package for it to happen. you got +1 on it.

dalenguyen:
+1

Nx would help to offload a lot of maintenance work if this means Analogjs will reuse Nx infrastructure.

Hazzajenko:
100% nx

stianmorsund:
I personally think Nx is superior, and have been using it alongside Angular almost since it was released. 

[...]

Issue #413, opened 2023-05-22
Add --minimal flag to generate a minimalist application


nartc:
### Which scope/s are relevant/related to the feature request?

create-analog

### Information

I want to see a `--minimal` flag when running the Application Generator. With `--minimal` flag, the generator will generate an application:
- With no test setup
- No server routes?
- A simpler `index.page.ts`

### Describe any alternatives/workarounds you're currently using

Generate a full application then remove things manually

### I would be willing to submit a PR to fix this issue

- [X] Yes
- [ ] No

brandonroberts:
I like minimal setups also

goetzrobin:
@nartc I can pick this up if you don’t have the bandwidth right now

nartc:
@goetzrobin go for it

PR #546, opened 2023-07-13
feat(create-analog): adding an option to select the ui framework

cskiwi:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] vite-plugin-angular
- [ ] vite-plugin-nitro
- [ ] astro-angular
- [x] create-analog
- [ ] router
- [ ] platform
- [ ] content
- [ ] nx-plugin
- [ ] trpc

the option `--skipTailwind` has been changed with an `--ui` option that allows the selection of the UI framework the developer wants to use

- [x] Yes
- [ ] No

existing commands using the `--skipTailwind` won't work anymore


brandonroberts:
Thanks @cskiwi. I need to think about this more. Not certain we should go down this route though with so many options out there. Today, it's Material and Tailwind. Tomorrow it's Material, Tailwind, Taiga UI, Bootstrap, Clarity, etc. We want to keep `create-analog` minimal and leave it up to the developer to tailor it further. I also don't see Tailwind/Material as mutually exclusive.

I could see the CLI go in the direction of picking what type of app you'd like to build. 

- Blog
- Todos (Fullstack Example)
- Barebones

We also have #517 which would change the work being done here also.

cskiwi:
@brandonroberts Sounds like a good direction to navigate to.

I've added the material stuff mostly because that's mostly what I've been using for the last 8-ish years. So creating new projects directly with Material included would save me some time.

Of course if it was fully integrated it with nx, then we would only have to validate that the normal way of adding angular material to a nx app works, and we'd be golden

goetzrobin:
@brandonroberts seems like what you are describing is along the lines of this effort, too: 
https://github.com/analogjs/analog/issues/413

Are you suggesting we'd continue to ask for tailwind like we currently do and then ask for the type of application?

brandonroberts:
@goetzrobin yes, I think the type of app you want to build and using Tailwind are two separate questions. Asking what type of app will generate a starter app from a template we have.

goetzrobin:
Awesome! I like that approach a lot!

PR #557, opened 2023-07-23
fix(router): do not run CD cycles unnecessary when markdown modules are loaded

arturovt:
Hey, I've just started experimenting with Analog and counting change detection cycles occurring in different scenarios due to the various asynchronous operations used in different parts of the code. I noticed that rendering a simple page, like `about.md` from the tutorial, causes 8 ticks immediately. This current commit updates the implementation of the `toMarkdownModule`, which loads `@analogjs/content` and `markdownFileFactory`, scheduling tasks in the root zone.

### DIFF
--- a/packages/router/src/lib/markdown-helpers.ts
+++ b/packages/router/src/lib/markdown-helpers.ts
@@ -1,32 +1,49 @@
 import { inject } from '@angular/core';
 import { RouteExport } from './models';
 
+declare const Zone: any;
+
+// The Zone is currently enabled by default, so we wouldn't need this check.
+// However, leaving this open space will be useful if zone.js becomes optional
+// in the future. This means we won't have to modify the current code, and it will
+// continue to work seamlessly.
+const isNgZoneEnabled = typeof Zone !== 'undefined' && !!Zone.root;
+
 export function toMarkdownModule(
   markdownFileFactory: () => Promise<string>
 ): () => Promise<RouteExport> {
-  return () =>
-    Promise.all([import('@analogjs/content'), markdownFileFactory()]).then(
-      ([
-        { parseRawContentFile, MarkdownRouteComponent, ContentRenderer },
-        markdownFile,
-      ]) => {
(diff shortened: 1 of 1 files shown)


brandonroberts (approved):

brandonroberts:
Thanks @arturovt. Will you share how you track the CD cycles?

arturovt:
Sure. Firstly, I start with patching `tick()`, which is the simplest way to track the number of change detections:

```ts
const tick = ApplicationRef.prototype.tick;
let ticksCalled = 0;
ApplicationRef.prototype.tick = function () {
  console.log(`Ticks called: ${ticksCalled++}`);
  return tick.call(this);
};
```

But this would only tell me the number of CDs happening.

Then I track asynchronous tasks which are invoked in the Angular zone. I usually just add a console logging to `node_modules/@angular/core/fesm2022/core.mjs -> forkInnerZoneWithAngularBehavior -> onInvokeTask: (delegate, current, target, task, applyThis, applyArgs)`, I add the call before `onEnter(zone)`:

```js
console.log(task);
onEnter(zone);
```

It'is easy to track event tasks since they have a `source` property (for example, `HTMLAnchorElement:click`).

It's always challenging to track microtasks because zone.js intercepts all calls to `new Promise`, `Promise.prototype.then`, and `Promise.resolve`. `then` and `resolve` typically lead to the creation of new `Promise` instances. zone.js replaces `globalThis.Promise` with its own implementation called `ZoneAwarePromise`, which affects debugging capabilities. In zone.js, there is a function called `scheduleResolveOrReject` where we may find the following code:

```js
zone.scheduleMicroTask(source, () => {
    try {
        const parentPromiseValue = promise[symbolValue];
```

The `promise` may be the original promise that resulted in a microtasks chain. However, this is just a surface-level observation, as it often takes some time to identify the root cause of the change detection triggering.

See also this code as an example how tasks can be easily tracked: https://ng-guru.io/guide/change-detection-cycles-per-different-async-apis/#tracing-asynchronous-tasks

PR #558, opened 2023-07-23
fix(vite-plugin-angular): do not execute hot module update if host is not set

arturovt:
This commit updates the implementation of the `handleHotUpdate` Vite plugin hook and includes a check to return if the `host` is not set. The `handleHotUpdate` hook may be called before the `buildStart` hook, which sets the compilation and the compiler host.

I've been having this error frequently:

Screenshot from 2023-07-23 20-59-40


brandonroberts:
Awesome find @arturovt. This has been an intermittent issue on Windows machines and some environments we hadn't tracked down yet.

brandonroberts (approved):

brandonroberts:
Thanks! @allcontributors add @arturovt for code.

JeanMeche:
Awesome, thx for the fix 👍

PR #561, opened 2023-07-23
fix(platform): cache content attributes if they have not changed

arturovt:


packages/platform/src/lib/content-plugin.ts
```diff
@@ -1,21 +1,46 @@
 import { Plugin } from 'vite';
 import * as fs from 'fs';
 
+interface Content {
+  code: string;
+  attributes: string;
+}
+
+const cache = new Map<string, Content>();
```
brandonroberts, line 9: This can be moved inside the plugin on line 11 before the array return

brandonroberts (approved):

PR #562, opened 2023-07-23
fix(content): do not run change detection when loading mermaid

arturovt:


brandonroberts:
@arturovt got some merge conflicts from the last fix that was merged. @JeanMeche will you take a look also?

brandonroberts (approved):

PR #566, opened 2023-07-24
fix(vite-plugin-angular): cache already resolved style and template URLs

arturovt:


brandonroberts:
Thanks @arturovt. I'm not sure how much we gain here as we're already reading from the transformed source, and not reading again from the filesystem to parse the external styles and templates.

arturovt:
@brandonroberts aren't we running `path.dirname` and `path.resolve` when resolving the absolute file path?

brandonroberts:
@arturovt that's true. I didn't treat those as filesystem read/write calls but resolve does use the `process.cwd()`

brandonroberts (approved):

Issue #468, opened 2023-06-19
using tRPC in constructor or ngOnInit causes build getting stuck


goetzrobin:
### Please provide the environment you discovered this bug in.

0.2.0-beta.17

### Which area/package is the issue in?

trpc

### Description

After upgrading to Angular 16.1 running `nx build trpc-app` get's stuck.
The following error is thrown:
```
ERROR TRPCClientError: fetch failed
    at TRPCClientError.from (file:///Users/robingotz/Developer/goetzrobin/analog/node_modules/.pnpm/@trpc+client@10.25.0_@trpc+server@10.25.0/node_modules/@trpc/client/dist/transformResult-403dc1d4.mjs:15:16)
....
     at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  meta: undefined,
  shape: undefined,
  data: undefined,
  [cause]: TypeError: fetch failed
      at Object.fetch (node:internal/deps/undici/undici:11457:11)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
    cause: Error: connect ECONNREFUSED 127.0.0.1:4205
        at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1494:16) {
      errno: -61,
      code: 'ECONNREFUSED',
      syscall: 'connect',
      address: '127.0.0.1',
      port: 4205
    }
  }
}
```

### Please provide the exception or error you saw

```true
ERROR TRPCClientError: fetch failed
    at TRPCClientError.from (file:///Users/robingotz/Developer/goetzrobin/analog/node_modules/.pnpm/@trpc+client@10.25.0_@trpc+server@10.25.0/node_modules/@trpc/client/dist/transformResult-403dc1d4.mjs:15:16)
....
     at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
  meta: undefined,
  shape: undefined,
  data: undefined,
  [cause]: TypeError: fetch failed
      at Object.fetch (node:internal/deps/undici/undici:11457:11)
      at process.processTicksAndRejections (node:internal/process/task_queues:95:5) {
    cause: Error: connect ECONNREFUSED 127.0.0.1:4205
        at TCPConnectWrap.afterConnect [as oncomplete] (node:net:1494:16) {
      errno: -61,
      code: 'ECONNREFUSED',
      syscall: 'connect',
      address: '127.0.0.1',
      port: 4205
    }
  }
}
```

### I would be willing to submit a PR to fix this issue

[...]

PR #574, opened 2023-07-26
feat(nx-plugin): TRPC using base url depending on the provider

cskiwi:
## PR Checklist

- [ ] The commit message follows our guidelines: https://github.com/analogjs/analog/blob/main/CONTRIBUTING.md#-commit-message-guidelines

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] vite-plugin-angular
- [ ] vite-plugin-nitro
- [ ] astro-angular
- [ ] create-analog
- [ ] router
- [ ] platform
- [ ] content
- [x] nx-plugin
- [ ] trpc

In the current implementation, the app defaults to running trpc on http://127.0.0.1:4200/. To enhance flexibility and accommodate different hosting providers, this pull request introduces a getBaseUrl function. With this update, the trpc endpoint can be easily adjusted based on the hosting environment. Changing the port or any other configurations will now be automatically resolved, providing a more adaptable and seamless experience.

This change aims to improve the app's configurability and ensure smoother deployments across various hosting setups. Your review and feedback on this pull request would be highly appreciated. Thank you!

Closes #

- [ ] Yes
- [x] No


goetzrobin:
I really like this! 
In an ideal world the user does not have to worry about setting an absolute path at all and can simply put "/api/trpc", "trpc", etc.

Maybe this should be exposed as a function of the trpc package instead of the Nx plugin template?
@brandonroberts @cskiwi what do you guys think?

cskiwi:
> I really like this! In an ideal world the user does not have to worry about setting an absolute path at all and can simply put "/api/trpc", "trpc", etc.
> 
> Maybe this should be exposed as a function of the trpc package instead of the Nx plugin template? @brandonroberts @cskiwi what do you guys think?

@goetzrobin That sounds like a good idea. However, before implementing it into the 'trpc' package, we should take a look at resolving the following issue: https://github.com/analogjs/analog/issues/468. 

This precautionary step is necessary to ensure that we don't introduce any breaking changes, especially if addressing the issue requires making additional changes to the way the API is passed along. By doing this, we can assist the users while still retaining the flexibility to make changes to the 'trpc' package in the future.

goetzrobin:
Hey there @cskiwi! Check out this PR: https://github.com/analogjs/analog/pull/639 
There might still be some gotchas I am missing, but it seems like it would make a `getBaseUrl` unnecessary and the TrpcClient would "just" work!

cskiwi:
> Hey there @cskiwi! Check out this PR: #639 There might still be some gotchas I am missing, but it seems like it would make a `getBaseUrl` unnecessary and the TrpcClient would "just" work!

Oh nice! this would be beautifull for the developer experience!

brandonroberts:
Closing as resolved via #639 which will go out with the next release

PR #575, opened 2023-07-26
docs(astro-angular): remove commas from list of `yarn add` dependencies

arturovt:
`,` (commas) are not allowed when listing dependencies.


brandonroberts:
@allcontributors add @arturovt for doc

PR #582, opened 2023-07-28
build(vite-plugin-angular): emit both CJS and ESM formats

arturovt:


benmccann:
I think you could convert `vite-plugin-angular` and `vite-plugin-nitro` to ESM only and leave the Nx plugin as CJS only. Is there any reason to dual publish this package? Vite works just fine with ESM plugins, so you really shouldn't need to provide a CJS plugin. Especially since user code is compiled to ESM for both the client-side and SSR. It seems like it just adds more complication and maintenance burden to keep CJS here and will eventually cause another migration to ESM only, so you might as well just go straight to the end state and be done with dealing with CJS in these packages for good

brandonroberts:
Closing in favor of #598

PR #659, opened 2023-09-19
feat(content): `ContentRenderer` is able to return a TOC

### LINKED ISSUE #658: The content renderer should be able to provide the TOC

content

### Information

It can be usefull to have a table of content for long and structured content. 

The content renderer should be able to provide it. 

_No response_

- [X] Yes
- [ ] No

JeanMeche:
this commits add the support for generating a table of content to the `MarkdownContentRenderer`

fixes #658

## PR Type

- [x] Feature

- [x] content

- [x] Yes


JeanMeche:
Should I maybe introduce a `injectTableOfContent` to get allong with `injectContent` so we'd have 2 way to retrieve the TOC ?

brandonroberts:
Yea I think it would be convenient to have. Can it not be provided as part of injectContent()? 

Unless injectContentHeaders() doesn't need any arguments and can figure it out.

brandonroberts:
Also need to add the new dependency here for upgrades

https://github.com/analogjs/analog/blob/main/packages/content/migrations/migration.json#L22

Is there no way to do it with just marked?

JeanMeche:
What do you think about the latest changes @brandonroberts ? 

Browser side we use the native `DomParser` and on the serve `JSDOM` handles the parsing.

brandonroberts:
@JeanMeche I think it's a decent solution. It looks like `jsdom` is getting bundled in the client though from looking at the `api-[hash].js` asset. 

image

Maybe we wrap it in `!import.meta.env.SSR` instead so the code gets tree-shaken from the browser build.

- Do we still need the `"github-slugger": "^2.0.0",` dependency?
- What happens if there are other headers on the page? Is up to the consumer to filter those out?

JeanMeche:
Yeah I dropped the `github-slugger`, since I'm parsing the generated DOM directly. I'm only feeding the DOM output to the parser, to external TOCs won't be an issue, it will be 100% content base.

packages/content/src/lib/markdown-content-renderer.service.ts
```diff
@@ -6,18 +6,42 @@ import {
...
     return this.#marked.getMarkedInstance().parse(content);
   }
 
+  async getTableOfContents(content: string): Promise<TableOfContentItem[]> {
+    const htmlText = await this.render(content);
+    let doc: Document;
+    if (typeof DOMParser === 'function') {
+      const parser = new DOMParser();
+      doc = parser.parseFromString(htmlText, 'text/html');
+    } else if (!import.meta.env.SSR) {
```
brandonroberts, line 27:
```suggestion
    } else if (import.meta.env.SSR) {
```

brandonroberts:
@JeanMeche ok cool. What do you think about making it available through `injectContent()` as a `toc` property?

JeanMeche:
Correct me if I'm wrong bug `injectContent()` returns unrendered text and we need `html` to generate the TOC. 

Should we assume the content is markdown and generate the toc from this assumption ?

brandonroberts:
> Correct me if I'm wrong bug `injectContent()` returns unrendered text and we need `html` to generate the TOC.
> 
> Should we assume the content is markdown and generate the toc from this assumption ?

No, you're correct. Let's keep it as-is.

JeanMeche:
Actually @brandonroberts, can we wait before merging this. 

I'd like to explore the solution proposed by @UziTech  https://github.com/markedjs/marked-gfm-heading-id/issues/412#issuecomment-1732043237 👍 (which would be way better)

JeanMeche:
I've opened a PR : 

* markedjs/marked-gfm-heading-id#413

brandonroberts (approved):

brandonroberts:
Nice! Ok, cool

JeanMeche:
The `marked-gfm-heading-id` PR has been merge and I've updated the implementation. 

Let me know what you think of it.

We could go one step further by removing the `parse` in the implementation and assume it will be called after the render method.

brandonroberts:
Much cleaner now. I think removing the `parse` is reasonable also. What about `getContentHeadings()` for the name?

brandonroberts:
Also needs an Nx package migration added to https://github.com/analogjs/analog/blob/main/packages/content/migrations/migration.json

JeanMeche:
@brandonroberts Did I get the migration right ?

brandonroberts:
@JeanMeche yep, its good.

brandonroberts:
A few remaining updates for new apps/workspaces

https://github.com/analogjs/analog/blob/main/packages/create-analog/template-angular-v16/package.json#L31

https://github.com/analogjs/analog/blob/main/packages/nx-plugin/src/generators/app/versions/nx_16_X/versions.ts#L15

https://github.com/analogjs/analog/blob/main/packages/nx-plugin/src/generators/app/generator.spec.ts#L73

brandonroberts (approved):
📑

PR #792, opened 2023-12-12
feat(vite-plugin-nitro): add support for typescript paths import resolution

jeremyhofer:
## PR Checklist

- [] Tests for the changes have been added (for bug fixes / features)
- [] Docs have been added / updated (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] vite-plugin-angular
- [X] vite-plugin-nitro
- [ ] astro-angular
- [ ] create-analog
- [ ] router
- [X] platform
- [ ] content
- [X] nx-plugin
- [ ] trpc

Currently typescript imports using the `compilerOptions.paths` aliases set in `tsconfig.json` are not automatically handled for Nitro. This causes the nitro server to fail to run when executing the imports. This is especially evident in a Nx monorepo leveraging multiple in-built libraries.

Closes # [Discord Thread](https://discordapp.com/channels/994618831987290112/1183254913086656718)

## What is the new behavior?

This PR adds [rollup-plugin-typescript-paths](https://www.npmjs.com/package/rollup-plugin-typescript-paths) as a dependency of `vite-plugin-nitro`. This plugin is included as one of the `plugins` within the `rollupConfig` of the `NitroConfig`.

The plugin is configured to default to loading / reading from a `tsconfig.json` file in the root of the repo. However, to support other setups an option was added to both the `Options` for `vite-plugin-nitro` and the `@analog/platform` plugin to allow a user to override the tsconfig file / path.

Related to the 2nd point, the `nx-plugin` was update to specify `tsconfig.base.json` as the tsconfig file to load. This was added as a default for all current `vite.config.ts` templates in the plugin.

- [ ] Yes
- [X] No

### DIFF
--- a/package.json
+++ b/package.json
@@ -150,6 +150,7 @@
     "prettier": "^2.8.3",
     "prismjs": "^1.29.0",
     "prompts": "^2.4.2",
+    "rollup-plugin-typescript-paths": "^1.4.0",
     "rollup-plugin-visualizer": "^5.9.0",
     "start-server-and-test": "^1.15.4",
     "tailwindcss": "^3.0.2",
--- a/packages/nx-plugin/src/generators/app/files/template-angular-v15/vite.config.ts__template__
+++ b/packages/nx-plugin/src/generators/app/files/template-angular-v15/vite.config.ts__template__
@@ -48,6 +48,7 @@ export default defineConfig(({ mode }) => {
         prerender: {
           routes: ['/'],
         },
+        tsConfigPath: 'tsconfig.base.json',
       }),
       tsConfigPaths({
         root: '<%= offsetFromRoot %>',
--- a/packages/nx-plugin/src/generators/app/files/template-angular-v16/vite.config.ts__template__
+++ b/packages/nx-plugin/src/generators/app/files/template-angular-v16/vite.config.ts__template__
@@ -25,10 +25,13 @@ export default defineConfig(({ mode }) => {
               prerender: false,
             }
           }
-        }
+        },
+        tsConfigPath: 'tsconfig.base.json',
       }),
       <% } else { %>
-      analog(),
+      analog({
+        tsConfigPath: 'tsconfig.base.json',
+      }),
       <% } %>
       tsConfigPaths({
         root: '<%= offsetFromRoot %>',
--- a/packages/nx-plugin/src/generators/app/files/template-angular-v17/vite.config.ts__template__
+++ b/packages/nx-plugin/src/generators/app/files/template-angular-v17/vite.config.ts__template__
@@ -25,10 +25,13 @@ export default defineConfig(({ mode }) => {
               prerender: false,
             }
           }
-        }
+        },
+        tsConfigPath: 'tsconfig.base.json',
       }),
       <% } else { %>
-      analog(),
+      analog({
+        tsConfigPath: 'tsconfig.base.json',
+      }),
       <% } %>
       nxViteTsPaths(),
       splitVendorChunkPlugin(),
--- a/packages/platform/src/lib/options.ts
+++ b/packages/platform/src/lib/options.ts
@@ -32,4 +32,5 @@ export interface Options {
   jit?: boolean;
   index?: string;
   workspaceRoot?: string;
+  tsConfigPath?: string;
 }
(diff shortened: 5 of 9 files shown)


jeremyhofer:
@brandonroberts wanted to follow up on this PR. Anything further I should look to address?

brandonroberts:
@jeremyhofer looks good in general. Just a little stuck on the naming because we already have a `tsconfig` property the `vite` object, which is different from this one and may cause some confusion.

```ts
analog({
  vite: { tsconfig: 'path/to/tsconfig.json' },
  tsConfigPath: 'tsconfig.base.json'
})
```

Maybe it should be `workspaceTsConfig` to go along with `workspaceRoot`.

Thoughts?

jeremyhofer:
@brandonroberts been experimenting with this and thinking more about how to best handle this, and want your thoughts.

First, some additional things I've determined regarding tsconfig paths support in analog today:
1. The import issue the PR is currently targeting only impacts running the Nitro server (`run`, `serve`, `build` and execute, and similar targets). It specifically does not have an impact on the `test` target.
2. Using tsconfig path imports within components / pages and unit tests cannot resolve imports either. However, this is resolved via `vite-tsconfig-paths` instead of the rollup/nitro plugin. Nx has migrations that checks for and adds `vite-tsconfig-paths` to your `vite-config.ts` file if it is not present, so I was not seeing the issue in my repo. However, removing that plugin caused the paths to not be resolved. Similarly, in a new standlone analog app, if you configure paths in `tsconfig.json` they are not recognized.

Regarding the current approach in this PR targeting a fix for the Nitro server - I think we can get by without configuring a `tsconfig` file, instead just referencing `tsconfig.app.json`. This ultimately extends the root `tsconfig.json` in a standalone analog project, or the `tsconfig.base.json` in a Nx repo. Unfortunately the `rollup-plugin-typescript-paths` dependency I added does not use typescript to fully resolve the tsconfig, so `extends` inheritance is not respected. This will have to be manually tooled.

I'm not having luck finding a rollup plugin that handles the tsconfig paths in a way that works with analog. Rollup provides the [@rollup/plugin-typescript](https://github.com/rollup/plugins/tree/master/packages/typescript), but it requires specific `tsconfig` configurations, especially related to the `outDir`, that do not work with how analog bundles the application. Other plugins similarly either are doing too much or not enough to fit the analog usecase.

Given all the above, I'm currently considering the following:
1. Manually implementing a rollup plugin to handle the tsconfig paths. Functionally it be like `rollup-plugin-typescript-paths`, but will use the typescript compiler to resolve the tsconfig and handle all inheritance, as the `@rollup/plugin-typescript` plugin does. This plugin will reference `tsconfig.app.json`.
2. Adding `vite-tsconfig-paths` as a dependency of `@analogjs/platform` and including the plugin in the set of plugins provided via the [platform plugin](https://github.com/analogjs/analog/blob/main/packages/platform/src/lib/platform-plugin.ts#L28), as this is needed for all project targets.

Thoughts or concerns on any of the above?

brandonroberts:
Option 2 is a better solution from a long-term maintenance standpoint. Nx uses their own tsconfig paths plugin, will this cause any conflicts with that?

jeremyhofer:
In this case what I recommended weren't different options, I'd have to do both items 1 and 2 from the bottom of my last comment to resolve tsconfig path support for all use cases.

For item 1 I called out the question is around if we're ok implementing the rollup plugin to support tsconfig paths for nitro / rollup, which is needed for bundling / running the server. Alternative would be to work the `@rollup/plugin-typescript` plugin into vite-plugin-nitro, but this may drive refactoring in vite-plugin-nitro.

For item 2, good callout on Nx's tsconfig path plugin. I just saw that this evening while experimenting with my test related posts. In my current project Nx had added the `vite-tsconfig-paths` plugin and configured it in a migration, their own plugin wasn't introduced yet. I'll experiment with this and see if there's any issues with analog including a vite tsconfig path plugin by default.

brandonroberts:
I've seen a couple others run into this also, but I think we would be better off adding this as a section in the Nx guide where its most likely to happen.

https://analogjs.org/docs/integrations/nx

Thoughts?

PR #793, opened 2023-12-12
feat(trpc): allow install of superjson version 2.2.1 and newer

jeremyhofer:
## PR Checklist

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] vite-plugin-angular
- [ ] vite-plugin-nitro
- [ ] astro-angular
- [ ] create-analog
- [ ] router
- [ ] platform
- [ ] content
- [ ] nx-plugin
- [X] trpc

Currently superjson 1.x version >= 1.12.3 may be installed. Latest superjson version is 2.2.1, which - minus them internally changing to an esmodule package format - is backwards compatible with v1.

Releases: https://github.com/blitz-js/superjson/releases

Closes #

## What is the new behavior?

Adds support for installing latest 2.x versions >= 2.2.1 in addition to 1.x versions.

- [ ] Yes
- [X] No


jeremyhofer:
@goetzrobin @brandonroberts following up on this PR. Any concerns with this change?

goetzrobin:
@jeremyhofer @brandonroberts I believe we can simply migrate to this new version instead of allowing both the old and the new. Thoughts?

jeremyhofer:
Updating solely to the 2.x version would work as well. I went with this approach initially since superjson is defined as a peer dependency and there are no compatibility issues between the 1.x and 2.x versions today. It would probably be more correct to allow any compatible 2.x version instead of requiring a version >=2.2.1, but that's the only 2.x version I've explicitly tested and know works.

Brings the question - if we change to a 2.x version only do we want to consider changing superjson to be an actual dependency instead of a peer? Peer makes me think full flexibility and allow any versions that are compatible. If we change to a dependency then we can remove the overhead of maintaining the version in a project's package.json.

brandonroberts:
I'm good with bumping to v2 if it's fully compatible. I think keeping it as a peer dep is fine and more generally preferred

jeremyhofer:
Just updated the branch to fully switch to v2.

brandonroberts:
We should add a `packageJsonUpdates` migration to the newer version also. Similar to here

https://github.com/analogjs/analog/blob/main/packages/content/migrations/migration.json#L4

jeremyhofer:
:+1: Will get this added!

packages/trpc/migrations/migration.json
```diff
@@ -1,4 +1,16 @@
 {
   "$schema": "../../../node_modules/@angular-devkit/schematics/collection-schema.json",
-  "schematics": {}
+  "schematics": {},
+  "packageJsonUpdates": {
+    "0.2.30": {
```
jeremyhofer, line 5: Question - do we need to define additional updates to handle the `1.0.0-beta` versions?

Also, let me know if these targeted versions are correct. Based them off the latest release version.

brandonroberts: No, we don't need to target the beta versions also.

brandonroberts (approved):

PR #806, opened 2023-12-18
feat(vite-plugin-nitro): provide req and res to renderer

### LINKED ISSUE #586: Feature: Accessing request and response objects within component class

router

### Information

Adding an ability to access request and response objects, would allow for server side authentication. As an example next js provides an exported function:

```ts
export const getServerSideProps = withSession(async function ({ req, res }) {
  const { user } = req.session
 
  if (!user) {
    return {
      redirect: {
        destination: '/login',
        permanent: false,
      },
    }
  }
 
  return {
    props: { user },
  }
})
 ```

It follows a similar pattern to [component.server.ts](https://github.com/analogjs/analog/blob/main/apps/analog-app/src/app/pages/(home).server.ts) files. The signature of the `load` function could be updated to return redirects, or another function to be exposed in order to not break the compatability.

No alternative right now.

- [X] Yes
- [ ] No

### LINKED ISSUE #802: Provide full request object to renderer
To provide cookies to your Angular app, it is sometimes necessary to provide it to your Angular from the server entrypoint prior to the Angular app taking over. Cookies, for example, are sometimes omitted due to Angular's xhr2's `strictCookies` setting. 

I have observed this both in my local analogjs app and pure Angular apps with ssr and express. To avoid this in an express-angular project, I simply provide the REQUEST to my app using the nguniversal token:

```
  server.get('*', (req, res, next) => {
    const { protocol, originalUrl, baseUrl, headers } = req;
    commonEngine
      .render({
        bootstrap,
        documentFilePath: indexHtml,
        url: `${protocol}://${headers.host}${originalUrl}`,
        publicPath: browserDistFolder,
        providers: [{ provide: APP_BASE_HREF, useValue: req.baseUrl }, { provide: REQUEST, useValue: req } ], // <--- provided here
      })
      .then((html) => res.send(html))
      .catch((err) => next(err));
  });

```

Analog does not yet provide a way to do this. The request object handled by a load hook, for example, which is really just a route resolver, is already sanitized by the framework; it is not a raw request and hence, in my case, lacked cookies.

I noticed that modifying the lines highlighted below gave me the full request with the cookies I needed:

https://github.com/analogjs/analog/blob/main/packages/vite-plugin-nitro/src/lib/runtime/renderer.js#L14-L15

https://github.com/analogjs/analog/blob/c812260168f86fd6e5fee2287e5e1f9848a63008/packages/vite-plugin-nitro/src/lib/plugins/dev-server-plugin.ts#L46

I am looking for either advice on how to make a backwards compatible PR that will work with different deployments, or from an existing contributor to take point on a solution. This is unfortunately a blocker for me :-(

### PR DESCRIPTION
**Pull request by Jefftopia**
PR adds full req and res object from h3 to the analog event handler for  app rendering

closes #802
closes #586 

## PR Checklist

- [ ] The commit message follows our guidelines: https://github.com/analogjs/analog/blob/main/CONTRIBUTING.md#-commit-message-guidelines

## PR Type

- [ ] Bugfix
- [ x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] vite-plugin-angular
- [ x] vite-plugin-nitro
- [ ] astro-angular
- [ ] create-analog
- [ ] router
- [ ] platform
- [ ] content
- [ ] nx-plugin
- [ ] trpc

Current behavior is that the full, raw request and response are not provided to Analog, which limits the ability to use cookies from the incoming request. 

Closes #802 

## What is the new behavior?

Now both the runtime renderer event handler and the nitro dev server have access to the request and response, which can then be provided to the application.

- [ ] Yes
- [x ] No

### DIFF
--- a/apps/analog-app/src/app/app.config.ts
+++ b/apps/analog-app/src/app/app.config.ts
@@ -7,12 +7,12 @@ import { ApplicationConfig } from '@angular/core';
 import { provideClientHydration } from '@angular/platform-browser';
 import { provideFileRouter } from '@analogjs/router';
 import { withNavigationErrorHandler } from '@angular/router';
-import { cookieInterceotor } from './interceptors/cookies.interceptor';
+import { cookieInterceptor } from './interceptors/cookies.interceptor';
 
 export const appConfig: ApplicationConfig = {
   providers: [
     provideFileRouter(withNavigationErrorHandler(console.error)),
-    provideHttpClient(withFetch(), withInterceptors([cookieInterceotor])),
+    provideHttpClient(withFetch(), withInterceptors([cookieInterceptor])),
     provideClientHydration(),
   ],
 };
--- a/apps/analog-app/src/app/app.config.ts
+++ b/apps/analog-app/src/app/app.config.ts
@@ -1,13 +1,18 @@
-import { provideHttpClient } from '@angular/common/http';
+import {
+  provideHttpClient,
+  withFetch,
+  withInterceptors,
+} from '@angular/common/http';
 import { ApplicationConfig } from '@angular/core';
 import { provideClientHydration } from '@angular/platform-browser';
 import { provideFileRouter } from '@analogjs/router';
 import { withNavigationErrorHandler } from '@angular/router';
+import { cookieInterceotor } from './interceptors/cookies.interceptor';
 
 export const appConfig: ApplicationConfig = {
   providers: [
     provideFileRouter(withNavigationErrorHandler(console.error)),
-    provideHttpClient(),
+    provideHttpClient(withFetch(), withInterceptors([cookieInterceotor])),
     provideClientHydration(),
   ],
 };
--- a/apps/analog-app/src/app/interceptors/cookies.interceptor.ts
+++ b/apps/analog-app/src/app/interceptors/cookies.interceptor.ts
@@ -1,0 +1,23 @@
+import { isPlatformServer } from '@angular/common';
+import { HttpHandlerFn, HttpHeaders, HttpRequest } from '@angular/common/http';
+import { PLATFORM_ID, inject } from '@angular/core';
+
+export function cookieInterceotor(
+  req: HttpRequest<unknown>,
+  next: HttpHandlerFn,
+  location = inject(PLATFORM_ID)
+) {
+  if (isPlatformServer(location)) {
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Review by brandonroberts — CHANGES_REQUESTED** [review]

File: apps/analog-app/src/app/app.config.ts [hunk 1]
```diff
@@ -1,13 +1,18 @@
...
+import {
+  provideHttpClient,
+  withFetch,
+  withInterceptors,
+} from '@angular/common/http';
 import { ApplicationConfig } from '@angular/core';
 import { provideClientHydration } from '@angular/platform-browser';
 import { provideFileRouter } from '@analogjs/router';
 import { withNavigationErrorHandler } from '@angular/router';
+import { cookieInterceotor } from './interceptors/cookies.interceptor';
```
brandonroberts, line 10:
```suggestion
import { cookieInterceptor } from './interceptors/cookies.interceptor';
```

apps/analog-app/src/app/app.config.ts
```diff
@@ -1,13 +1,18 @@
...
 
 export const appConfig: ApplicationConfig = {
   providers: [
     provideFileRouter(withNavigationErrorHandler(console.error)),
-    provideHttpClient(),
+    provideHttpClient(withFetch(), withInterceptors([cookieInterceotor])),
```
brandonroberts, line 15:
```suggestion
    provideHttpClient(withFetch(), withInterceptors([cookieInterceptor])),
```

apps/analog-app/src/app/interceptors/cookies.interceptor.ts
```diff
@@ -0,0 +1,23 @@
+import { isPlatformServer } from '@angular/common';
+import { HttpHandlerFn, HttpHeaders, HttpRequest } from '@angular/common/http';
+import { PLATFORM_ID, inject } from '@angular/core';
+
+export function cookieInterceotor(
```
brandonroberts, line 5:
```suggestion
export function cookieInterceptor(
```

brandonroberts (approved):
🍪

brandonroberts:
Looks great @Jefftopia! One small cleanup item on the interceptor naming

brandonroberts:
Thanks! @allcontributors add @Jefftopia for code

PR #856, opened 2024-01-19
feat(vite-plugin-angular): add ability to export from .ng files

joshuamorony:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [x] vite-plugin-angular
- [ ] vite-plugin-nitro
- [ ] astro-angular
- [ ] create-analog
- [ ] router
- [ ] platform
- [ ] content
- [ ] nx-plugin
- [ ] trpc

.ng format does not support defining a routeMeta

## What is the new behavior?

This allows adding a `route` configuration to `defineMetadata` e.g:

```ts
  defineMetadata({
    route: {
      title: 'My page',
    },
  });
```

Happy to modify how this is supplied - this seemed the most natural to me (I have also extended the window type to include the `RouteMeta` type for `route`).

I assumed we didn't want to `export` the routeMeta in .ng files. Another option I considered was having a separate `defineRouteMetadata`

- [ ] Yes
- [x] No

### DIFF
--- a/packages/vite-plugin-angular/src/lib/authoring/ng.ts
+++ b/packages/vite-plugin-angular/src/lib/authoring/ng.ts
@@ -165,6 +165,16 @@ function processNgScript(
 
     const nodeFullText = node.getText();
 
+    if (
+      Node.isTypeAliasDeclaration(node) ||
+      Node.isInterfaceDeclaration(node) ||
+      Node.isEnumDeclaration(node)
+    ) {
+      if (node.hasExportKeyword()) {
+        targetSourceFile.addStatements(nodeFullText);
+      }
+    }
+
     // for VariableStatement (e.g: const ... = ..., let ... = ...)
     if (Node.isVariableStatement(node)) {
       // NOTE: we do not support multiple declarations (i.e: const a, b, c)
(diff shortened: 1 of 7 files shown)


packages/vite-plugin-angular/src/lib/authoring/ng.ts
```diff
@@ -368,7 +373,8 @@ ${gettersSetters
 function processMetadata(
   metadataObject: ObjectLiteralExpression,
   targetMetadataArguments: ObjectLiteralExpression,
-  targetClass: ClassDeclaration
+  targetClass: ClassDeclaration,
+  targetSourceFile: any
```
nartc, line 377: nit: the type is `SourceFile` imported from `ts-morph`

nartc (commented):
I personally would go with `export const routeMeta` and use `ts-morph` to bring this to the `targetSourceFile` instead of using `defineMetadata` for it.

packages/vite-plugin-angular/src/lib/authoring/ng.ts
joshuamorony: Thanks fixed up the type - as for `export const routeMeta` wasn't there a plan to utilise exports for outputs?

But it does feel a bit strange to me to have two different ways to handle metadata, e.g:

```ts
export const routeMeta = {
  title: 'My page'
}

defineMetadata({
  selector: 'my-component'
})
```

Anyway happy to follow whatever you or Brandon think here, feel free to give me the final word and I'll implement the change

nartc: re:output No, we're using `new EventEmitter` directly.

I'm ok either way. cc @brandonroberts

brandonroberts: We do need to find a good pattern for defining code that's inside an ng file that's not component class code.

I was thinking about using `export` but because we can statically find inputs/outputs I don't think its necessary. 

I think for consistency with Analog routes in non-ng files, we should go with the `export const routeMeta` that gets put outside the transformed component class.

joshuamorony: What about rather than making this change just for `routeMeta` we use `export` as a way to export anything outside of the component class. I've had people complain that this .ng format wouldn't allow doing things like exporting and interface.

So perhaps rather than specifically grabbing the `routeMeta` I make a change to do this:

```html
<script lang="ts">
  export const routeMeta: RouteMeta = {
    title: 'My page'
  }

  export interface MyInterface {
    title: string
  }
</script>

etc
```

```ts
@Component({
  etc
})
export class VirtualAnalogComopnent { 

}

export const routeMeta: RouteMeta = {
  title: 'My page'
}

export interface MyInterface {
  title: string
}
```

brandonroberts: Yes, I think that's reasonable.

nartc: I'm also good with that.

joshuamorony: Pushed an implementation for this - just need to give it a proper look over and will also add an example to the ng-app later today. Let me know if there's anything you want changed.

joshuamorony: This should be good to go now

nartc (approved):

brandonroberts (approved):

brandonroberts:
Thanks! @allcontributors add @joshuamorony for code

PR #872, opened 2024-01-30
feat(vite-plugin-angular): add inline markdown support to analog template

brandonroberts:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [x] vite-plugin-angular
- [ ] vite-plugin-nitro
- [ ] astro-angular
- [ ] create-analog
- [ ] router
- [ ] platform
- [ ] content
- [ ] nx-plugin
- [ ] trpc

Closes #

## What is the new behavior?

An inline template for a `.analog` file supports authoring with markdown. This template is compiled into HTML and used as the component template, allowing for Angular syntax and components to be used.

```html
<script lang="ts">
  import Hello from '../hello.analog';
</script>

<template lang="md">
  # About me works!

  ## Subheader

  '''sh
  npm create analog
  '''

  <Hello />
</template>
```

Result:

image

NOTE: The same constraints exist for an Angular component template, such as using `@` directly, interpolation tags, and more.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/vite-plugin-angular/src/lib/authoring/analog.spec.ts
+++ b/packages/vite-plugin-angular/src/lib/authoring/analog.spec.ts
@@ -109,6 +109,12 @@ effect(() => {
 </script>
 `;
 
+const MARKDOWN_CONTENT = `
+<template lang="md">
+  # Hello World
+</template>
+`;
+
 describe('authoring ng file', () => {
   it('should process component as ng file', () => {
     const source = compileAnalogFile('virtual.analog.ts', COMPONENT_CONTENT);
@@ -119,6 +125,12 @@ describe('authoring ng file', () => {
   it('should process directive as ng file', () => {
     const source = compileAnalogFile('virtual.analog.ts', DIRECTIVE_CONTENT);
     expect(source).toContain('Directive');
+    expect(source).toMatchSnapshot();
+  });
+
+  it('should process component with markdown as analog file with a virtual templateUrl', () => {
+    const source = compileAnalogFile('virtual.analog.ts', MARKDOWN_CONTENT);
+    expect(source).toContain('Component');
     expect(source).toMatchSnapshot();
   });
 });
--- a/packages/vite-plugin-angular/src/lib/authoring/analog.spec.ts
+++ b/packages/vite-plugin-angular/src/lib/authoring/analog.spec.ts
@@ -116,13 +116,13 @@ const MARKDOWN_CONTENT = `
 `;
 
 describe('authoring ng file', () => {
-  it('should process component as ng file', () => {
+  it('should process component as an analog file', () => {
     const source = compileAnalogFile('virtual.analog.ts', COMPONENT_CONTENT);
     expect(source).toContain('Component');
     expect(source).toMatchSnapshot();
   });
 
-  it('should process directive as ng file', () => {
+  it('should process directive as an analog file', () => {
     const source = compileAnalogFile('virtual.analog.ts', DIRECTIVE_CONTENT);
     expect(source).toContain('Directive');
     expect(source).toMatchSnapshot();
--- a/packages/vite-plugin-angular/src/lib/authoring/analog.ts
+++ b/packages/vite-plugin-angular/src/lib/authoring/analog.ts
@@ -27,7 +27,8 @@ const INVALID_METADATA_PROPERTIES = [
 ];
 
 const SCRIPT_TAG_REGEX = /<script lang="ts">([\s\S]*?)<\/script>/i;
-const TEMPLATE_TAG_REGEX = /<template>([\s\S]*?)<\/template>/i;
+const TEMPLATE_TAG_REGEX =
+  /(<template>|<template lang="md">)([\s\S]*?)<\/template>/i;
 const STYLE_TAG_REGEX = /<style>([\s\S]*?)<\/style>/i;
 
 const ON_INIT = 'onInit';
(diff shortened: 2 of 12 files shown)


goetzrobin (commented):
Love love love this PR! Just so I understand what is going on. If you use the `<template lang="md">` tag, the template get's turned into a virtual file, which is recognized by it's "virtual" path and parsed through the `marked` service while it is read. Which means the template contains "regular" html. Importantly this "regular" html can contain Angular elements, because it will then be again parsed through the default `.analog` readFile function, which means we end up with markdown that supports Angular components. Is that about right? 😍

packages/vite-plugin-angular/src/lib/host.ts
```diff
@@ -53,16 +54,60 @@ export function augmentHostWithResources(
         ...parameters
       );
     };
+
+    const baseReadFile = (resourceHost as ts.CompilerHost).readFile;
+    const TEMPLATE_TAG_REGEX =
+      /(<template>|<template lang="md">)([\s\S]*?)<\/template>/i;
```
nartc, line 60: suggestion: let's make these constants in a separate file

packages/vite-plugin-angular/src/lib/authoring/analog.ts
```diff
@@ -77,7 +78,12 @@ export function compileAnalogFile(
   const componentMetadata = (() => {
     if (ngType === 'Component') {
       const items = ['changeDetection: ChangeDetectionStrategy.OnPush'];
-      if (templateContent) {
+
+      if (fileContent.includes('lang="md"')) {
```
nartc, line 82: suggestion: I'd compute this flag ahead and check if this is true then skip the `TEMPLATE_TAG_REGEX.exec()` in this file.

luishcastroc:
> Love love love this PR! Just so I understand what is going on. If you use the `<template lang="md">` tag, the template get's turned into a virtual file, which is recognized by it's "virtual" path and parsed through the `marked` service while it is read. Which means the template contains "regular" html. Importantly this "regular" html can contain Angular elements, because it will then be again parsed through the default `.analog` readFile function, which means we end up with markdown that supports Angular components. Is that about right? 😍 

Pretty much what I see as well

goetzrobin:
> > Love love love this PR! Just so I understand what is going on. If you use the `<template lang="md">` tag, the template get's turned into a virtual file, which is recognized by it's "virtual" path and parsed through the `marked` service while it is read. Which means the template contains "regular" html. Importantly this "regular" html can contain Angular elements, because it will then be again parsed through the default `.analog` readFile function, which means we end up with markdown that supports Angular components. Is that about right? 😍
> 
> Yep, that's about right.

Love it! One thing that comes to mind as a potential future enhancement is allowing developers to pass their own custom marked instance through the vite config to the `MarkedSetupService`.

brandonroberts:
> > > Love love love this PR! Just so I understand what is going on. If you use the `<template lang="md">` tag, the template get's turned into a virtual file, which is recognized by it's "virtual" path and parsed through the `marked` service while it is read. Which means the template contains "regular" html. Importantly this "regular" html can contain Angular elements, because it will then be again parsed through the default `.analog` readFile function, which means we end up with markdown that supports Angular components. Is that about right? 😍
> > 
> > 
> > Yep, that's about right.
> 
> Love it! One thing that comes to mind as a potential future enhancement is allowing developers to pass their own custom marked instance through the vite config to the `MarkedSetupService`.

Definitely! Developers may want custom highlighting also instead of Prism as mentioned by @nartc

packages/vite-plugin-angular/src/lib/authoring/analog.spec.ts
```diff
@@ -109,6 +109,12 @@ effect(() => {
 </script>
 `;
 
+const MARKDOWN_CONTENT = `
+<template lang="md">
+  # Hello World
+</template>
+`;
+
 describe('authoring ng file', () => {
```
luishcastroc, line 118: should we re-name this for consistency?

luishcastroc (approved):

joshuamorony:
Absolutely loving seeing this work, thank you and I can't wait to get my hands on it.

Do you have any opinion on using remark/rehype for parsing instead of marked (not sure if there is some technical limitation here)? I have a custom remark plugin for example that relies on using the Markdown AST, and it would be nice to have access to that ecosystem of plugins too.

I imagine we could then allow people to pass in whatever plugins they like with something like:

```ts
provideContent(withMarkdownRenderer({
   remarkPlugins: [[preview, { percentage: 0.4 }], relativeImages, getHeadings, getWordCount]
   rehypePlugins: [rehypeSlug, rehypeAutolinkHeadings, rehypePrism],
}))
```

brandonroberts:
> Absolutely loving seeing this work, thank you and I can't wait to get my hands on it.
> 
> Do you have any opinion on using remark/rehype for parsing instead of marked (not sure if there is some technical limitation here)? I have a custom remark plugin for example that relies on using the Markdown AST, and it would be nice to have access to that ecosystem of plugins too.
> 
> I imagine we could then allow people to pass in whatever plugins they like with something like:
> 
> ```ts
> provideContent(withMarkdownRenderer({
>    remarkPlugins: [[preview, { percentage: 0.4 }], relativeImages, getHeadings, getWordCount]
>    rehypePlugins: [rehypeSlug, rehypeAutolinkHeadings, rehypePrism],
> }))
> ```

I don't think there's a technical limitation here, but more of a default option. I think this would be more of documenting how to provide your own renderer with `remark` for example.

As long as you implement a service that implements the [ContentRenderer](https://github.com/analogjs/analog/blob/main/packages/content/src/lib/content-renderer.ts) abstract class and provide it, you can provide your own implementation.

nartc (approved):

PR #905, opened 2024-02-25
feat(vite-plugin-angular): add support for import attributes

joshuamorony:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [x] vite-plugin-angular
- [ ] vite-plugin-nitro
- [ ] astro-angular
- [ ] create-analog
- [ ] router
- [ ] platform
- [ ] content
- [ ] nx-plugin
- [ ] trpc

See: https://github.com/analogjs/analog/discussions/901

## What is the new behavior?

Allows mapping `providers`, `viewProviders`, `imports`, and `exposes` to component/directive metadata via import attributes (`exposes` is a special case that has special handling and is not added to the metadata)

- [x] Yes
- [ ] No

This removes the existing functionality that supported adding analog components to `imports` if the import was directly a `.analog` file. We could leave this functionality in place also if we want.


packages/vite-plugin-angular/src/lib/authoring/analog.ts
```diff
@@ -160,10 +167,36 @@ function processAnalogScript(
...
+        for (const attribute of attributes) {
+          const value = attribute.value.replaceAll("'", '');
+
+          if (!(value in meta)) {
+            throw new Error(
+              `[Analog] meta property "${value}" is not supported`
+            );
+          }
+
+          if (attribute.name === 'meta') {
```
nartc, line 182: nit: we should probably check this along with the above condition?

packages/vite-plugin-angular/src/lib/authoring/analog.ts
```diff
@@ -160,10 +167,36 @@ function processAnalogScript(
...
+            const defaultImport = node.getDefaultImport();
```
nartc, line 183: nit: the `getStructure()` above should have the default import in there.

packages/vite-plugin-angular/src/lib/authoring/analog.ts
```diff
@@ -160,10 +167,36 @@ function processAnalogScript(
...
+            if (defaultImport) {
+              meta[value].push(defaultImport.getText());
+            }
+
+            const namedImports = node.getNamedImports();
```
nartc, line 188: nit: same as above.the `getStructure()` should have this information already

nartc (commented):
A couple of nits and a question/suggestion

Long shot but what if an `ImportAttribute` is used purposefully from the userland that don't have anything to do with our `meta` checking? We probably should still keep those `ImportAttribute` around for their import statements when we move those imports to a `.ts` file.

joshuamorony:
> Long shot but what if an `ImportAttribute` is used purposefully from the userland that don't have anything to do with our `meta` checking? We probably should still keep those `ImportAttribute` around for their import statements when we move those imports to a `.ts` file.

Yes this makes sense - I've removed the error what was thrown if it doesn't match what Analog expects so non-Analog import attributes can still pass through. I replaced the error with a `console.warn`, is there a better approach here?

nartc (approved):

PR #912, opened 2024-02-28
feat(vite-plugin-angular): allow analog imports with no name

joshuamorony:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [x] vite-plugin-angular
- [ ] vite-plugin-nitro
- [ ] astro-angular
- [ ] create-analog
- [ ] router
- [ ] platform
- [ ] content
- [ ] nx-plugin
- [ ] trpc

Analog components need to be imported like this:

```ts
import Hello from './hello.analog' with { analog: 'imports' };
```

and can then be used like this:

```html
<Hello />
```

I think this creates the false perception that the name of the default import is relevant, when it isn't actually used at all to determine the components selector. If we change the default import name:

```ts
import HelloComponent from './hello.analog' with { analog: 'imports' };
```

```html
<HelloComponent />
```

This will not work. I think this is surprising behaviour, and second to that since we don't actually need the default import name it is unnecessary boilerplate.

## What is the new behavior?

This PR allows importing analog components like this:

```ts
import './hello.analog' with { analog: 'imports' };
```

Which I think makes for nicer DX and removes the confusion around the default import name having anything to do with the selector.

- [ ] Yes
- [x] No

### DIFF
--- a/apps/ng-app/src/app/app.component.analog
+++ b/apps/ng-app/src/app/app.component.analog
@@ -5,10 +5,10 @@
   import { RouterOutlet, RouterLink } from '@angular/router' with { analog: 'imports'};
   import { delay } from 'rxjs';
 
-  import Hello from './hello.analog' with { analog: 'imports'};
-  import AnotherOne from './another-one.analog' with { analog: 'imports' };
-  import Highlight from './highlight.analog' with { analog: 'imports' };
-  import External from './external/external.analog' with { analog: 'imports' };
+  import './hello.analog' with { analog: 'imports'};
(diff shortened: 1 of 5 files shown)


brandonroberts:
I think it could be an option, but not "the way". I like the convention of importing the `ComponentName` and using it even if its only symbolic. You still have to use a "selector" for it to work.

Feels less magical than just importing a file and thinking about what the component tag is based on the filename.

And if/when direct component imports become a thing upstream, we won't need to change anything.

nartc (approved):

Issue #688, opened 2023-10-04
Fresh project on Windows fails


DmitryEfimenko:
### Please provide the environment you discovered this bug in.

Windows

### Which area/package is the issue in?

create-analog

### Description

I'm on a Windows as well. Tried starting a fresh project following ["Getting Started" docs](https://analogjs.org/docs/getting-started). Immediately received an error when attempted to start the project:

### Please provide the exception or error you saw

```true
$ npm run build

> app@0.0.0 build
> ng build

vite v4.4.10 building for production...
✓ 341 modules transformed.
dist/client/index.html                             1.75 kB │ gzip:  1.31 kB
dist/client/assets/index-05bc7378.css              5.95 kB │ gzip:  1.86 kB
dist/client/assets/index.page-af7a8736.js          1.31 kB │ gzip:  0.72 kB
dist/client/assets/analogjs-content-a8651e95.js  138.45 kB │ gzip: 44.83 kB
dist/client/assets/index-29aeeebf.js             260.23 kB │ gzip: 82.55 kB
Building SSR application...
✓ built in 5.95s
vite v4.4.10 building SSR bundle for production...
✓ 31 modules transformed.
dist/ssr/assets/index.page-60cc8828.mjs            2.48 kB
dist/ssr/assets/analogjs-content-72cce11b.mjs     18.53 kB
dist/ssr/assets/xhr2-516fb8e9.mjs                 36.46 kB
dist/ssr/main.server.mjs                       1,305.53 kB
✓ built in 5.00s
Prerendering static pages...
[nitro] [request error] [unhandled] Only URLs with a scheme in: file, data are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:' 
  at new NodeError (node:internal/errors:393:5)  
  at throwIfUnsupportedURLScheme (node:internal/modules/esm/resolve:1026:11)  
  at defaultResolve (node:internal/modules/esm/resolve:1106:3)  
  at nextResolve (node:internal/modules/esm/loader:163:28)  
  at ESMLoader.resolve (node:internal/modules/esm/loader:837:30)  
  at ESMLoader.getModuleJob (node:internal/modules/esm/loader:424:18)  
  at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:76:40)  
  at link (node:internal/modules/esm/module_job:75:36)
Building Server...


The '@analogjs/platform' server has been successfully built.

[...]

PR #915, opened 2024-02-29
fix: add file protocol to solve windows pre rendering issues

angelfraga:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] vite-plugin-angular
- [x] vite-plugin-nitro
- [ ] astro-angular
- [ ] create-analog
- [ ] router
- [ ] platform
- [ ] content
- [ ] nx-plugin
- [ ] trpc

In windows pre rendering fails due next error

```
[nitro] [request error] [unhandled] Only URLs with a scheme in: file, data are supported by the default ESM loader. On Windows, absolute paths must be valid file:// URLs. Received protocol 'c:' 
```

Closes https://github.com/analogjs/analog/issues/688

- [ ] Yes
- [ ] No

### DIFF
--- a/packages/vite-plugin-nitro/src/lib/vite-plugin-nitro.ts
+++ b/packages/vite-plugin-nitro/src/lib/vite-plugin-nitro.ts
@@ -3,6 +3,7 @@ import { App, toNodeListener } from 'h3';
 import type { Plugin, UserConfig } from 'vite';
 import { normalizePath, ViteDevServer } from 'vite';
 import * as path from 'path';
+import { platform } from 'node:os';
 
 import { buildServer } from './build-server';
 import { buildSSRApp } from './build-ssr';
@@ -15,6 +16,7 @@ import { devServerPlugin } from './plugins/dev-server-plugin';
 import { loadEsmModule } from './utils/load-esm';
 import { getMatchingContentFilesWithFrontMatter } from './utils/get-content-files';
 
+const isWindows = platform() === 'win32';
 let clientOutputPath = '';
 
 export function nitro(options?: Options, nitroOptions?: NitroConfig): Plugin[] {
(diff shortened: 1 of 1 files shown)


angelfraga:
@brandonroberts  while this change actually fixes the build in windows, I am aware this might not be the final fix.
I did not even tried out in WSL or Mac and probably we might need an additional OS check before prepending that.

packages/vite-plugin-nitro/src/lib/vite-plugin-nitro.ts
```diff
@@ -104,9 +104,12 @@ export function nitro(options?: Options, nitroOptions?: NitroConfig): Plugin[] {
         }
 
         nitroConfig.alias = {
-          '#analog/ssr': normalizePath(
-            path.resolve(workspaceRoot, 'dist', rootDir, 'ssr/main.server')
-          ),
+          // This is not the final fix but start point to discuss a fix for windows
+          '#analog/ssr':
+            'file://' +
```
brandonroberts, line 109: You can use this to check for windows

```ts
import os from 'node:os'

const isWindows = os.platform() === 'win32'
```

```suggestion
            (isWindows ? 'file://' : '') +
```

angelfraga: Thanks, i will add it and check tests in Windows after work

angelfraga:
@brandonroberts after running the test under windows with the fix for SSG actually does not fix all of them. In addition test snapshot have differencies between windows and other linux/mac.

But in the other hand this fix seems to solve the issue. So it would be great to have somebody else checking the fix under windows. 

In order to test the fix, just running the blog app in the repo is enough.

```
pnpm exec nx run blog-app:build --configuration=production
```

Before the fix

image

After the fix

image


Here is my env: 

Windows 11
Node 20.10.0

PR #965, opened 2024-03-25
fix(nx-plugin): rename Vite config and correct Vitest version for Vitest schematic

brandonroberts:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] vite-plugin-angular
- [ ] vite-plugin-nitro
- [ ] astro-angular
- [ ] create-analog
- [ ] router
- [ ] platform
- [ ] content
- [x] nx-plugin
- [ ] trpc

The `setup-vitest` schematic/generator installs an incorrect version of Vitest (1.31.1), and uses a `vite.config.ts` which fails because an existing project is not set to use ESM by default.

Closes #

## What is the new behavior?

- The `setup-vitest` schematic/generator installs the correct version of Vitest (^1.3.1).
- A `vite.config.mts` file is used instead which uses ESM for Vitest only.
- The `^1.0.0` version of Analog packages is installed.

- [ ] Yes
- [x] No


Micha-Richter:
Do the vite config files in the templates of create-analog also need to be renamed? 🤔

brandonroberts:
@Micha-Richter no, they don't. This is for the standalone Vitest setup. If you want to migrate completely to Analog, more changes are required, including updating the package.json, angular.json, and possibly renaming vite.config.mts to vite.config.ts. Both filenames will work though. We're using vite.config.mts in this case so Vite resolves it to ESM, because the plugins are ESM-only now. The Angular CLI is still using CommonJS though.

I don't want to force the full update to ESM for existing workspaces as its not as fully tested with the Angular CLI.

PR #1044, opened 2024-04-18
fix(content): ensure content has been rendered before SSR serialization

### LINKED ISSUE #1038: Blog Pages Generated By Content Path Have Flicker

1. Clone the Analog Project
2. Run `pnpm i`
3. Run `nx serve blog-app`

Navigate here: http://localhost:3000/archived/post1-2024

You will see the page load empty first, and then it loads. 

## First Flicker Problem

There are actually 2 different flicker problems.

The first flicker is due to the Observable Not Resolving in time for the server to load the data. It does not `await` the markdown comopnent to render. At first I thought this was a problem with the `of('')` on [this line](https://github.com/analogjs/analog/blob/abe129e6ef0b80d36c1317b25c617e589e310be5/packages/content/src/lib/markdown.component.ts#L51), but even when I set the variable directly to equal the router observable, it does not complete before the server page is generated.

Generally, I would use the [waitFor](https://github.com/analogjs/analog/blob/abe129e6ef0b80d36c1317b25c617e589e310be5/packages/content/src/lib/utils/zone-wait-for.ts) function to force the component to resolve the observable (as a promise) on the server in a different zone before rendering, but that does not work either.  This is a dire problem as it could effect loading meta tags or schema for SEO purposes. We need to generate on the server.

**We need a Zone.js expert for this.**

## Second Flicker Problem
If you set the `content$` directly and comment out the markdown rendering on [this line](https://github.com/analogjs/analog/blob/abe129e6ef0b80d36c1317b25c617e589e310be5/packages/content/src/lib/markdown.component.ts#L84), you will see it loads (without markdown parsing) on the server in time. However, there is a second flicker.

This problem is not as serious, but I believe the second flicker is from the observable running 2x. Generally speaking the server should fetch the data and hydrate the data to the browser so it doesn't have to reparse the markdown component. It should not re-run the observable on the client unless there is a route change. 

Should we just remove the observables all together here? I think signals could help simplify some of this code.

content

### Description

Here is a direct blog example you can see the problem:

- https://www.prestonlamb.com/blog/strongly-typed-reactive-forms-in-angular
- https://perko.dev/blog/post/2023-08-31-quickly-setup-drupal-on-windows-with-laragon

I can't tell if this is a second flicker problem or transition, so may be unrelated:
- https://mrrobot.dev/blog/the-importance-of-softskills

I just looked a few blogs from Brandon's post:
- https://dev.to/analogjs/how-to-build-a-blog-with-analog-and-angular-4pk2

```true
Flicker.

One way to fix the problem completely would be to remove the ability to pass the markdown code directly to the markdown component and handle it server side only. This would keep the client code bundle smaller. It adds a better UX but worse DX (arguably).

We may not need to worry about the ZoneJS issue as @atscott points out `Pending Tasks` will take over ZoneJS's ability to run async tasks outside a component.

https://github.com/analogjs/analog/issues/1033

Either way, Flicker Problem #1 needs immediate attention.
```

### Other information

If someone figures out the fix and doesn't want to do the PR, I am willing to do it.

- [X] Yes
- [ ] No

brandonroberts:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] vite-plugin-angular
- [ ] vite-plugin-nitro
- [ ] astro-angular
- [ ] create-analog
- [ ] router
- [ ] platform
- [x] content
- [ ] nx-plugin
- [ ] trpc

When rendering markdown content using the `analog-markdown` component, there is a flicker when the content renders. This happens because the markdown is compiled asynchronously, and the SSR serialization has already happened before the content is rendered.

This causes the content to be re-rendered on the client instead of hydrated after SSR.

Closes #1038 

## What is the new behavior?

This uses the `ɵPendingTasks` (internal) API to add a pending task as soon as possible to delay SSR serialization until after the content is rendered. As a fallback, if nothing is rendered within 100ms, the pending task is cleared. Open to suggestions on a better solution here. Using the router navigation events as a fallback happens too quickly.

AFTER/BEFORE:

https://github.com/analogjs/analog/assets/42211/a5849b12-898c-4159-9ef7-93906fd83bd9

- [ ] Yes
- [x] No

### DIFF
--- a/packages/content/src/lib/render-task.service.ts
+++ b/packages/content/src/lib/render-task.service.ts
@@ -1,0 +1,31 @@
+import {
+  Injectable,
+  inject,
+  ɵPendingTasks as PendingTasks,
+} from '@angular/core';
+import { filter, first, timeout } from 'rxjs';
+
+@Injectable()
+export class RenderTaskService {
+  pendingTasks = inject(PendingTasks);
+  private readonly renderTask = this.pendingTasks.add();
+
+  constructor() {
+    this.pendingTasks.hasPendingTasks
+      .pipe(
+        filter((isStable) => !isStable),
+        timeout(100),
+        first()
+      )
+      .subscribe({
+        next: () => {},
+        error: () => {
+          this.clearRenderTask();
+        },
+      });
+  }
+
+  clearRenderTask() {
+    this.pendingTasks.remove(this.renderTask);
+  }
+}
--- a/packages/content/src/lib/markdown-content-renderer.service.spec.ts
+++ b/packages/content/src/lib/markdown-content-renderer.service.spec.ts
@@ -2,11 +2,16 @@ import { TestBed } from '@angular/core/testing';
 
 import { MarkdownContentRendererService } from './markdown-content-renderer.service';
 import { MarkedSetupService } from './marked-setup.service';
+import { RenderTaskService } from './render-task.service';
 
 describe('MarkdownContentRendererService', () => {
   function setup() {
     TestBed.configureTestingModule({
-      providers: [MarkdownContentRendererService, MarkedSetupService],
+      providers: [
+        MarkdownContentRendererService,
+        MarkedSetupService,
+        RenderTaskService,
+      ],
     });
     return { service: TestBed.inject(MarkdownContentRendererService) };
   }
--- a/packages/content/src/lib/markdown-content-renderer.service.ts
+++ b/packages/content/src/lib/markdown-content-renderer.service.ts
@@ -1,22 +1,25 @@
 import {
+  ENVIRONMENT_INITIALIZER,
   inject,
   Injectable,
   InjectionToken,
-  PLATFORM_ID,
   Provider,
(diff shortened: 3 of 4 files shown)


packages/content/src/lib/render-task.service.ts
```diff
@@ -0,0 +1,31 @@
+import {
+  Injectable,
+  inject,
+  ɵPendingTasks as PendingTasks,
+} from '@angular/core';
+import { filter, first, timeout } from 'rxjs';
+
+@Injectable()
+export class RenderTaskService {
+  pendingTasks = inject(PendingTasks);
```
dalenguyen, line 10: Thanks, It's good to know about how to use `PendingTasks`! 

Nit: `#pendingTasks = inject(PendingTasks)`?

brandonroberts: Fixed, thanks!

PR #1158, opened 2024-06-12
feat: upgrade to pnpm v9.x and nx v19.2.x

benpsnyder:
Upgrades to latest Nx v19.2.x with migrations having been run. I added some minor, explicit devDependencies that PNPM v9 required. Core changes are minimal and builds pass.


brandonroberts:
@allcontributors add @benpsnyder for infra

PR #1188, opened 2024-07-01
docs: make vitest test code compile for easy copy-paste

niklas-wortmann:
I tried to copy the code and noticed that it didn't compile, technically it is marked with `...` but still this makes it somewhat easier to use.


brandonroberts (approved):

brandonroberts:
Thanks! @allcontributors add @niklas-wortmann for doc

PR #1190, opened 2024-07-01
fix(vite-plugin-angular): prevent context loss for "each" tests

### LINKED ISSUE #1115: Can't test with `it/describe.each` and `fakeAsync`

https://stackblitz.com/~/github.com/jahusa02/analog-vitest

vitest-angular

### Description

When a testfile contains `it/describe.each` and `fakeAsync` the tests errors out.

I described the error in this Issue https://github.com/analogjs/analog/issues/841#issuecomment-2110296265|

When I use the patched function (e.g. not importing it via `import { it } from 'vitest'`), the `it.each` tests will fail.

```typescript
TypeError: Cannot read properties of undefined (reading 'withContext')
 ❯ ../../../packages/vite-plugin-angular/setup-vitest.ts:247:31
 ❯ src/lib/+state/effects/add-pending-contract/add-pending-contract.effects.spec.ts:98:6
     96|       { formStateValid: false, contractId: 'foo', satznummer: null },
     97|       { formStateValid: false, contractId: null, satznummer: 'bar' },
     98|     ])(
       |      ^
     99|       'should not post form values if ContractIds are %j',
    100|       ({ formStateValid, contractId, satznummer }) => {
 ❯ _ZoneDelegate.invoke ../../../node_modules/zone.js/fesm2015/zone.js:365:28
 ❯ ZoneImpl.run ../../../node_modules/zone.js/fesm2015/zone.js:111:43
 ❯ ../../../packages/vite-plugin-angular/setup-vitest.ts:41:21
``` 

When I importing `it`, the `fakeAsync` tests will fail with the mentioned `Error: Expected to be running in 'ProxyZone', but it was not found.`, but the `it.each` will pass.

```true
Either

TypeError: Cannot read properties of undefined (reading 'withContext')

or

Error: Expected to be running in 'ProxyZone', but it was not found.
```

### Other information

_No response_

- [ ] Yes
- [ ] No

gultyayev:
## PR Checklist

## PR Type

- [x] Bugfix
- [ ] Feature
- [ ] Build related changes
- [ ] CI related changes

- [x] vite-plugin-angular
- [ ] vite-plugin-nitro
- [ ] vitest-angular
- [ ] astro-angular
- [ ] create-analog
- [ ] router
- [ ] platform
- [ ] content
- [ ] nx-plugin
- [ ] trpc

Currently, when zones are wrapping the test function the context is lost causing an error "Cannot read `withContext()` of undefined".

Closes #1115

## What is the new behavior?

The original fn contains the mentioned `withContext()` method. I added reference to the initial method to the bounding function in order to preserve context.

- [ ] Yes
- [x] No

## Other information

I'm not sure how to properly test this change. I did not find tests for `setup-vitest`. I also failed to test this package using `npm link` as it failed complaining 

```
failed to load config from /Users/crush/Projects/analog-vitest-repro/apps/analog-vitest/vite.config.mts
Error: Cannot find module '@angular/build/private'
```

The original fix was found by tinkering the `setup-vitest.js` file from node_modules from the repro in the issue. I applied the same changes to the .ts file.

### DIFF
--- a/packages/vite-plugin-angular/setup-vitest.ts
+++ b/packages/vite-plugin-angular/setup-vitest.ts
@@ -229,22 +229,25 @@ const bindDescribe = (originalVitestFn: {
 /**
  * bind test method to wrap test.each function
  */
-const bindTest = (originalVitestFn: {
-  apply: (
-    arg0: any,
-    arg1: any[]
-  ) => {
-    (): any;
-    new (): any;
-    apply: { (arg0: any, arg1: any[]): any; new (): any };
-  };
-}) =>
+const bindTest = (
+  self: any,
+  originalVitestFn: {
+    apply: (
+      arg0: any,
+      arg1: any[]
+    ) => {
+      (): any;
+      new (): any;
+      apply: { (arg0: any, arg1: any[]): any; new (): any };
+    };
+  }
+) =>
   function (...eachArgs: any) {
     return function (...args: any[]) {
       args[1] = wrapTestInZone(args[1]);
 
       // @ts-ignore
-      return originalVitestFn.apply(this, eachArgs).apply(this, args);
+      return originalVitestFn.apply(self, eachArgs).apply(self, args);
     };
   };
 
@@ -269,9 +272,9 @@ const bindTest = (originalVitestFn: {
 
     return originalvitestFn.apply(this, args);
   };
-  env[methodName].each = bindTest(originalvitestFn.each);
-  env[methodName].only = bindTest(originalvitestFn.only);
-  env[methodName].skip = bindTest(originalvitestFn.skip);
+  env[methodName].each = bindTest(originalvitestFn, originalvitestFn.each);
+  env[methodName].only = bindTest(originalvitestFn, originalvitestFn.only);
+  env[methodName].skip = bindTest(originalvitestFn, originalvitestFn.skip);
 
   if (methodName === 'test' || methodName === 'it') {
     env[methodName].todo = function (...args: any) {
--- a/packages/vitest-angular/setup-zone.ts
+++ b/packages/vitest-angular/setup-zone.ts
@@ -229,22 +229,25 @@ const bindDescribe = (originalVitestFn: {
 /**
  * bind test method to wrap test.each function
  */
-const bindTest = (originalVitestFn: {
-  apply: (
-    arg0: any,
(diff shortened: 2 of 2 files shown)


brandonroberts:
Thanks @gultyayev! Will you apply the change here also? 

https://github.com/analogjs/analog/blob/beta/packages/vitest-angular/setup-zone.ts

gultyayev:
> Thanks @gultyayev! Will you apply the change here also?
> 
> https://github.com/analogjs/analog/blob/beta/packages/vitest-angular/setup-zone.ts

Sure.

However, I just realised that somehow my fix works only when a test is run via WebStorm 🤔 If I run `nx run analog-vitest:test` it still gives the same error...

Now I feel even more related to the Gif

brandonroberts:
> > Thanks @gultyayev! Will you apply the change here also?
> > https://github.com/analogjs/analog/blob/beta/packages/vitest-angular/setup-zone.ts
> 
> Sure.
> 
> However, I just realised that somehow my fix works only when a test is run via WebStorm 🤔 If I run `nx run analog-vitest:test` it still gives the same error...
> 
> Now I feel even more related to the Gif

Oh, interesting 🤔

gultyayev:
Oh... I just realized why this is happening.

The two errors, are actually one from each test suite.

image

My fix fixes the global variables patching (test.service.spec.ts).
However, the second case, when we import helper functions directly from `vitest` is unhandled. I suppose, it's not handled initially by the plugin.

brandonroberts:
That's correct. This assumes you're using the patched globals

gultyayev:
> That's correct. This assumes you're using the patched globals

Is there a way for us to also patch the imports?

gultyayev:
I can see two points for improvements now:

1. Update docs to mention this limitatin
2. Update the schematics & docs to also provide global types for TypeScript language service not to complain

I guess I could raise two separate MRs to followup.

brandonroberts:
> I can see two points for improvements now:
> 
> 1. Update docs to mention this limitation
> 2. Update the schematics & docs to also provide global types for TypeScript language service not to complain
> 
> I guess I could raise two separate MRs to followup.

That would be helpful, thanks

brandonroberts (approved):

brandonroberts:
Thanks! @allcontributors add @gultyayev for code

PR #1194, opened 2024-07-02
fix(router): add error message when default export is missing from component

### LINKED ISSUE #1129: Add error message when `xx.page.ts` does not have a default export

router

### Information

I pulled my hair out trying to figure out why my routing does not work. Turns out the component needs to be a `default` export.

There are two improvements that could be made here:

1. Make that requirement more visible in the documentation, right now it's not immediately visible when skimming the docs. I would make it a bigger "alert" with maybe a yellow background.
2. Add a check and error message before resolving the routes, instead of propagating the default Angular error that a `component` needs to be provided in the `Route` definition.

_No response_

- [x] Yes, if pointed in the right direction
- [ ] No

andersonmfjr:
## PR Checklist

- [X] Tests for the changes have been added (for bug fixes / features)

## PR Type

- [ ] Bugfix
- [X] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] vite-plugin-angular
- [ ] vite-plugin-nitro
- [ ] vitest-angular
- [ ] astro-angular
- [ ] create-analog
- [X] router
- [ ] platform
- [ ] content
- [ ] nx-plugin
- [ ] trpc

Closes #1129 

## What is the new behavior?

Show a console.warn when default export is missing from component

- [ ] Yes
- [X] No

### DIFF
--- a/packages/router/src/lib/routes.ts
+++ b/packages/router/src/lib/routes.ts
@@ -182,15 +182,27 @@ function toRoutes(rawRoutes: RawRoute[], files: Files): Route[] {
       ? {
           path: rawRoute.segment,
           loadChildren: () =>
-            module!().then((m) => [
-              {
-                path: '',
-                component: m.default,
-                ...toRouteConfig(m.routeMeta as RouteMeta | undefined),
-                children,
-                [ANALOG_META_KEY]: analogMeta,
-              },
-            ]),
+            module!().then((m) => {
+              const hasModuleDefault = !!m.default;
+              if (
+                process.env['NODE_ENV'] !== 'production' &&
+                !hasModuleDefault
+              ) {
+                throw new Error(
+                  `[Analog] Missing default export at ${rawRoute.filename}`
+                );
+              }
+
+              return [
+                {
+                  path: '',
+                  component: m.default,
+                  ...toRouteConfig(m.routeMeta as RouteMeta | undefined),
+                  children,
+                  [ANALOG_META_KEY]: analogMeta,
+                },
+              ];
+            }),
         }
       : { path: rawRoute.segment, children };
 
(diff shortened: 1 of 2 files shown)


packages/router/src/lib/routes.ts
```diff
@@ -182,15 +182,27 @@ function toRoutes(rawRoutes: RawRoute[], files: Files): Route[] {
...
-                [ANALOG_META_KEY]: analogMeta,
-              },
-            ]),
+            module!().then((m) => {
+              const hasModuleDefault = !!m.default;
+              if (
+                process.env['NODE_ENV'] !== 'production' &&
+                !hasModuleDefault
+              ) {
+                throw new Error(
```
brandonroberts, line 191: Let's just use `console.warn` here instead of throwing an error

packages/router/src/lib/routes.ts
```diff
@@ -182,15 +182,27 @@ function toRoutes(rawRoutes: RawRoute[], files: Files): Route[] {
...
-                children,
-                [ANALOG_META_KEY]: analogMeta,
-              },
-            ]),
+            module!().then((m) => {
+              const hasModuleDefault = !!m.default;
+              if (
+                process.env['NODE_ENV'] !== 'production' &&
+                !hasModuleDefault
+              ) {
```
brandonroberts, line 190: Let's use the Vite environment variable so this gets tree-shaken on a prod build

```suggestion
              if (!import.meta.env.PROD) {
                const hasModuleDefault = !!m.default;
                
                if (!hasModuleDefault) {
```

brandonroberts (changes requested):

packages/router/src/lib/routes.ts
andersonmfjr: thanks a bunch for your time and consideration

brandonroberts (approved):

brandonroberts:
Thanks! @allcontributors add @andersonmfjr for code

PR #1202, opened 2024-07-03
feat(vite-plugin-nitro): add internal deployment support for Netlify

pieh:
## PR Checklist

## PR Type

- [ ] Bugfix
- [x] Feature
- [ ] Build related changes
- [ ] CI related changes

- [ ] vite-plugin-angular
- [x] vite-plugin-nitro
- [ ] vitest-angular
- [ ] astro-angular
- [ ] create-analog
- [ ] router
- [ ] platform
- [ ] content
- [ ] nx-plugin
- [ ] trpc

Users deploying to Netlify need to additionally configure location of produced serverless functions

Closes #

## What is the new behavior?

Serverless functions are generated at location that Netlify platform automatically handle removing the need to specify ~custom serverless function directory when configuring deployments to Netlify.

Additional note is that hopefully soon at least non-nx variant would require zero configuration. We recently added Analog to frameworks we automatically detect ( https://github.com/netlify/build/pull/5751 ) which soon will suggest appropiate settings when creating Analog site in Netlify web ui and there is work in-progress to use same detection automatically configure deployment when using Netlify CLI ( https://github.com/netlify/cli/pull/6729 )

- [ ] Yes
- [x] No

Explanation: users defining Netlify Functions directory with current instructions will still be able to deploy successfully because Netlify always look for functions in adjusted directory, so even if we point functions directory to one that no longer exist (or at least doesn't have Analog function) it will work correctly:

```
Packaging Functions from .netlify/functions-internal directory:
 - server/server.mjs
No Functions were found in dist/analog directory
```

### DIFF
--- a/packages/vite-plugin-nitro/src/lib/vite-plugin-nitro.ts
+++ b/packages/vite-plugin-nitro/src/lib/vite-plugin-nitro.ts
@@ -132,6 +132,10 @@ export function nitro(options?: Options, nitroOptions?: NitroConfig): Plugin[] {
           nitroConfig = withCloudflareOutput(nitroConfig);
         }
 
+        if (isNetlifyPreset(buildPreset)) {
+          nitroConfig = withNetlifyOutput(nitroConfig);
+        }
+
         if (!ssrBuild && !isTest) {
           // store the client output path for the SSR build config
           clientOutputPath = resolve(
@@ -366,5 +370,18 @@ const withCloudflareOutput = (nitroConfig: NitroConfig | undefined) => ({
   output: {
     ...nitroConfig?.output,
     serverDir: '{{ output.publicDir }}/_worker.js',
+  },
+});
+
+const isNetlifyPreset = (buildPreset: string | undefined) =>
+  process.env['NETLIFY'] ||
+  process.env['NETLIFY_LOCAL'] ||
+  (buildPreset && buildPreset.toLowerCase().includes('netlify'));
+
+const withNetlifyOutput = (nitroConfig: NitroConfig | undefined) => ({
+  ...nitroConfig,
+  output: {
+    ...nitroConfig?.output,
+    dir: '{{ rootDir }}/.netlify/functions-internal',
   },
 });
--- a/apps/docs-app/docs/features/deployment/providers.md
+++ b/apps/docs-app/docs/features/deployment/providers.md
@@ -13,14 +13,13 @@ Analog supports deploying on [Netlify](https://netlify.com/) with minimal config
 
 <Tabs groupId="porject-type">
   <TabItem label="Create analog" value="create-analog">
-In the build settings of your Netlify project, set the [publish directory](https://docs.netlify.com/configure-builds/overview/#definitions) to `dist/analog/public` to deploy the static assets and the [functions directory](https://docs.netlify.com/configure-builds/overview/#definitions) to `dist/analog` to deploy the server.
+In the build settings of your Netlify project, set the [publish directory](https://docs.netlify.com/configure-builds/overview/#definitions) to `dist/analog/public` to deploy the static assets.
   </TabItem>
 
   <TabItem label="Nx" value="nx">
 In the build settings of your Netlify project on the web UI, do the following.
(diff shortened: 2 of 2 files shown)


brandonroberts:
Thanks @pieh! Can this be zero-config for static assets also?

pieh:
> Can this be zero-config for static assets also?

With https://github.com/netlify/build/pull/5751 we will soon automatically pre-fill configuration when creating a new site on Netlify web app (once that change actually end up in production), so this flow will be ~somewhat zero conf for static assets soon. 

For CLI usage (`netlify deploy --build`) for now users will still need to define deploy settings in `netlify.toml` configuration file, but this no longer will be needed when https://github.com/netlify/cli/pull/6729 is finished, so eventually it will be fully zero conf, just not yet.

Netlify doesn't have equivalent to something like Vercel's output api for static assets (https://github.com/analogjs/analog/blob/beta/packages/vite-plugin-nitro/src/lib/vite-plugin-nitro.ts#L354-L356) where there is always specific directory from which static assets are uploaded - Netlify relies on "publish directory" for those things, so instead of frameworks putting those files in deployment specific directory, either user or our framework detection tell us where framework outputs static assets

With that said, I think something is not exactly right with either this change or configuration of test Netlify sites/apps because preview deploy for `analog-app` 404s for API route ( https://deploy-preview-1202--analog-app.netlify.app/api/v1/products ) which wasn't happening for previous pull requests ( https://deploy-preview-1201--analog-app.netlify.app/api/v1/products ), so this probably should not be merged yet until this is figured out what's wrong.

I did attempt to deploy this app on my side of things locally with following configuration (so `functions` dir is not set, relying on the change I did):
```
# netlify.toml
[build]
command = "nx build analog-app"
publish = "dist/apps/analog-app/analog/public"
```
and I did get working deploy ( https://66855a09ed43201e349b11f4--analog-app-no-functions-defined-test.netlify.app/api/v1/products ), so some debugging is needed why deploy preview on this PR is not actually working correctly

brandonroberts:
Thanks for getting better Analog support merged into Netlify!

Ok cool. It could be my manual deployment settings that are conflicting.

image

pieh:
I think I have clue given the site settings - the "Package path" is not set here while mine was equivalent to `apps/analog-app` and likely the function was generated in `apps/analog-app/.netlify/functions-internal`, while Netlify would look for `/.netlify/functions-internal` when "Package path" is not set.

So while I do think it would be beneficial for site to specify "correct" package path, any change in this PR should NOT break the sites with currently working setting, so I'll look if I can figure out setup that would just work for both scenarios with Package Path set and not set.

I don't have much Nitro knowledge and the `dir` I set here is just straight copied from Nitro's Netlify preset ( https://github.com/unjs/nitro/blob/63526f24b03e5e5de65934c697c92dd00f1e1ab1/src/presets/netlify/preset.ts#L20 ), so this change likely "inherited" the same problem from the original setting

packages/vite-plugin-nitro/src/lib/vite-plugin-nitro.ts
```diff
@@ -368,3 +372,16 @@ const withCloudflareOutput = (nitroConfig: NitroConfig | undefined) => ({
...
+const isNetlifyPreset = (buildPreset: string | undefined) =>
+  process.env['NETLIFY'] ||
+  process.env['NETLIFY_LOCAL'] ||
+  (buildPreset && buildPreset.toLowerCase().includes('netlify'));
+
+const withNetlifyOutput = (nitroConfig: NitroConfig | undefined) => ({
+  ...nitroConfig,
+  output: {
+    ...nitroConfig?.output,
+    dir: '{{ rootDir }}/.netlify/functions-internal',
```
brandonroberts, line 385: I think what we need to do here is set the `serverDir`. The `dir` sets the output for everything else, including the `publicDir` and `.nitro` build directory.

```suggestion
    serverDir: '{{ rootDir }}/.netlify/functions-internal',
```

brandonroberts (changes requested):

brandonroberts:
Yea, I don't usually use the package path because even in a monorepo setup like Nx, I build from the root and use the project path to build where the output goes in the `dist` folder, which I think is more common.

brandonroberts:
@pieh were you able to find anything else on this?

Issue #904, opened 2024-02-24
Enhancing Router Support for Shared Pages in Monorepo Setups


n-can-k:
### Which scope/s are relevant/related to the feature request?

router

### Information

I'm working on a monorepo (NX) that includes multiple applications, and I've encountered a challenge with routing for pages shared across these applications. Currently, the router doesn't seem to support monitoring pages outside of the app folder, which is crucial for my setup since the apps are essentially just shells, and all modules and pages are imported from the corresponding folder under the `libs`.

To address this, I wanted to implement a custom router provider.
This required me to access to some internal functions like `createRoutes` and `updateMetaTagsOnRouteChange`, so I exported them in my fork. And it now works on my apps.
( I submitted a merge request to expose those internals: https://github.com/analogjs/analog/pull/903 )

Alternatively, I considered extending `provideFileRouter` with a new `files` parameter, but I think this isn't the right approach since it won't be backwards compatible.

Do you think the custom router provider path is the way to go, or is there another solution?

Thanks a ton for considering my two cents.

### I would be willing to submit a PR to fix this issue

- [X] Yes
- [ ] No

brandonroberts:
I think exposing `createRoutes` is fine, but I'd rather expose the `ENVIROMENT_INITIALIZER` that `updateMetaTagsOnRouteChange` uses as a function than that function directly. Something like `provideRouteMetaTagsListener()` or something similar.

n-can-k:
i think adding new provider function like this and exposing the `createRoutes` function to create new routes is also an option.

```ts
export function provideFileRouterWithCustomRoutes(
  customRoutes : Route[],
  ...features: RouterFeatures[]
): EnvironmentProviders {
  return makeEnvironmentProviders([
    // TODO: remove type casting after Angular >=15.1.1 upgrade
    // https://github.com/angular/angular/pull/48720
    (
      provideRouter([ ...customRoutes, ...routes ], ...features) as unknown as {
        ɵproviders: Provider[];
      }
    ).ɵproviders,
    {
      provide: ENVIRONMENT_INITIALIZER,
      multi: true,
      useValue: () => updateMetaTagsOnRouteChange(),
    },
  ]);
}
```


this way exposed internals would be minimal

brandonroberts:
You can do that already without anything custom using `provideRoutes`. The `ROUTES` token is uses is a multi-token.

```ts
import {
  provideHttpClient,
  withFetch,
  withInterceptors,
} from '@angular/common/http';
[...]

PR #1228, opened 2024-07-24
feat: add support for configuring additional page/content route sources

brandonroberts:
## PR Checklist

Related to #904

## What is the new behavior?

- Page and content routes can be sourced from directories outside the main `src/app/pages` and `src/content` directories
- Adds support for discovering page routes and content from projects in a monorepo
- Adds support to the Vite plugin for Angular to include additional files in compilation
- Allows the `createRoutes` function to be used with the Angular application builder without overrides

```ts
/// <reference types="vitest" />

import analog from '@analogjs/platform';

// https://vitejs.dev/config/
  build: {
    target: ['es2020'],
  },
  resolve: {
    mainFields: ['module'],
  },
  plugins: [
    analog({
      // scans for `**/*.page.{ts,analog}`. Must be inside a `pages` directory
      additionalPagesDirs: ['/libs/shared/feature'],

      // scans for `**/*.{md,agx}`. Must be inside a `content` directory
      additionalContentDirs: ['/libs/shared/feature'],

      // scans for `**/*.{ts}`. Must be inside a `routes` directory
      additionalAPIDirs: ['/libs/shared/feature/src/api']

    })
  ],
  test: {
    globals: true,
    environment: 'jsdom',
    setupFiles: ['src/test-setup.ts'],
    include: ['**/*.spec.ts'],
    reporters: ['default'],
  },
  define: {
    'import.meta.vitest': mode !== 'production',
  },
}));
```

- [ ] Yes
- [x] No


nckirik:
we can also add a config option for API route dirs. this is currently available through nitro.scanDirs config. just an alias would be enough I think

brandonroberts:
> we can also add a config option for API route dirs. this is currently available through nitro.scanDirs config. just an alias would be enough I think

Added the `additionalAPIDirs` option for scanning API routes

PR #1276, opened 2024-08-14
fix(router): handle cookie header internally

osnoser1:
## PR Checklist

Closes: None.

## What is the new behavior?

* Cookies will be available on `load` function of `.server.ts` files without implementing a custom interceptor.

- [ ] Yes
- [x] No

## Other information

Very small test:
* In the `apps/analog-app/src/app/pages/(home).server.ts` file, add a console.log.
image

* Without this change, if you remove the cookies interceptor of `apps/analog-app/src/app/app.config.ts`, the console log will print an empty object.


brandonroberts:
What if they already have a "cookie" header? I think the header should have a unique name at least, such as `X-Analog-Cookie`. 

Is `cookie` one value or a dictionary of key/value cookies?

osnoser1:
> What if they already have a "cookie" header? I think the header should have a unique name at least, such as `X-Analog-Cookie`.
> 
> Is `cookie` one value or a dictionary of key/value cookies?

I was thinking that the cookie header should only be added if the request URL has the format `/api/_analog/whatever`; I believe it can work in that case.

osnoser1:
The Idea is, if we make a request to by example: `http://localhost:4200/posts/my-slug`, when the `load` function of the `.server.ts` file runs, I can get the cookies using the `h3` functions (`getCookie`, `parseCoikies`, etc). If we set custom headers, we won't be able to use the h3 cookies functions.

osnoser1:
> cookie

`cookie` is a complete cookie string, the server with the `h3` functions parse internally for us that string

packages/router/src/lib/provide-file-router.ts
```diff
@@ -26,5 +28,10 @@ export function provideFileRouter(
       multi: true,
       useValue: () => updateMetaTagsOnRouteChange(),
     },
+    {
```
brandonroberts, line 31: I didn't know there was an internal token for adding root interceptor functions. I'd like to use this for the `requestContextInterceptor` also, but that's separate from this PR.

brandonroberts:
> The idea is that if we make a request to, for example, `http://localhost:4200/posts/my-slug`, when the `load` function of the `.server.ts` file runs, we can get the cookies using the `h3` functions (`getCookie`, `parseCookies`, etc.). If we set custom headers, we won't be able to use the h3 cookies functions, we would have to use the headers functions (that it is valid).
> 
> The only cookies this request https://github.com/analogjs/analog/blob/beta/packages%2Frouter%2Fsrc%2Flib%2Froute-config.ts#L71 should have are the server request cookies

I see, that makes sense. We can keep it scoped to start as its only for internal Analog requests.

brandonroberts (approved):

PR #1307, opened 2024-08-30
feat(vite-plugin-angular): add support for .ag extension

joshuamorony:
## PR Checklist

Allow `.ag` extension as a shorthand alternative to `.analog`

- [ ] Yes
- [x] No


packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
```diff
@@ -368,7 +368,9 @@ export function angular(options?: PluginOptions): Plugin[] {
           }
 
           if (
-            (id.endsWith('.analog') || id.endsWith('.agx')) &&
+            (id.endsWith('.analog') ||
+              id.endsWith('.agx') ||
```
ajitzero, line 372: question: What exactly is `*.agx`? `agx` isn't set up in all the places where `ag` is set up now. Looks like an `mdx` equivalent with a markdown focus?

joshuamorony: Essentially yes, it allows using analog components as content files with front matter

ajitzero: Ah, okay. I was under the impression that they were all just aliases of each other. Thanks!

brandonroberts (approved):

PR #1326, opened 2024-09-06
fix(router): add responseType in requestContextInterceptor for HttpClient

### LINKED ISSUE #1322: `Nitro` throws an error when using `HttpClient ` with non-JSON `responseType`

https://stackblitz.com/edit/github-sccmpf?file=src%2Fapp%2Fpages%2Findex.page.ts

router

### Description

I've encountered an issue where `Nitro` throws an error during the server rendering step when using the `HttpClient` with the new Server Request Context feature, if the `responseType` is set to anything other than `json`. This error occurs when attempting to request data in formats such as `text`, `blob`, or `arraybuffer`.

```true
Response is not a string
```

### Other information

The issue originates from the `request-context.ts` file in the router package, where support for responseType needs to be added in Nitro's fetch client.

```js
    return from(
      global.$fetch
        .raw(fetchUrl, {
          method: req.method as any,
          params: requestUrl.searchParams,
          responseType: req.responseType, // add responseType support
          headers: req.headers.keys().reduce((hdrs, current) => {
            return {
              ...hdrs,
              [current]: req.headers.get(current),
            };
          }, {}),
        })
```

- [X] Yes
- [ ] No

lars0n:
… responseType from httpclient

## PR Checklist

Closes #1322

## What is the new behavior?
Allow HttpClient to support responses other than JSON in SSR

- [ ] Yes
- [x] No

### DIFF
--- a/packages/router/src/lib/request-context.ts
+++ b/packages/router/src/lib/request-context.ts
@@ -39,12 +39,15 @@ export function requestContextInterceptor(
     const fetchUrl = req.url.includes('/api/')
       ? requestUrl.pathname
       : requestUrl.href;
+    const responseType =
+      req.responseType === 'arraybuffer' ? 'arrayBuffer' : req.responseType;
 
     return from(
       global.$fetch
         .raw(fetchUrl, {
           method: req.method as any,
           params: requestUrl.searchParams,
+          responseType,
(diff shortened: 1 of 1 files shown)


brandonroberts (approved):

brandonroberts:
Thanks! @allcontributors add @lars0n for code

PR #1344, opened 2024-09-11
fix(vite-plugin-angular): allow supplying and exporting vfile data for agx files

joshuamorony:
I've taken the liberty of experimenting with a solution for this, so of course feel free to disregard if the approach/API isn't appropriate. Happy to make changes as required too.

Currently with `agx` files the frontmatter export is created before the markdown template transforms run, which is fine, but when the transform functions run they might want to add `vfile` data (as is the case with some remark/rehype plugins e.g. I have created some plugins that track word count/headings through vfile data)

At the moment when a markdown template transform is run it might return all the data as a vfile, e.g:

```ts
VFile {
  cwd: '/Users/joshuamorony/Dev/projects/oss/analog-stuff',
  data: {},
  history: [],
  messages: [],
  value: '<h2>Test</h2>\n' +
    '<p>Hello</p>\n' +
    '<ul>\n' +
    '<li>one</li>\n' +
    '<li>two</li>\n' +
    '<li>three</li>\n' +
    '</ul>\n' +
    '<ul>\n' +
    '<li>one</li>\n' +
    '<li>two</li>\n' +
    '<li>three</li>\n' +
    '</ul>'
}
```

But then the rest of the data needs to be dropped to be compatible with the markdown template transform format, and just the value as the content string is returned, e.g:

```ts
export const agxRemarkRehype =
  (options: RemarkRehypeOptions = {}): MarkdownTemplateTransform =>
  async (content: string) => {
    const processor = setupUnified(options);

    // Contains all the vFile data
    const mdContent = await processor.process(content);

    // Contains only the value string from the vFile data
    return String(mdContent);
  };
```

So all of the vfile data, including any data plugins might have added, will be lost. Currently there is no way to add metadata to the resulting content (e.g. through markdown transform/plugins), only the original frontmatter defined in the file will make it through when actually using the agx component.

e.g. given this plugin set up:

```ts
markdownTemplateTransforms: [
  agxRemarkRehype({
    remarkPlugins: [
      [preview, { percentage: 0.4 }],
      getHeadings,
      getWordCount,
    ],
    rehypePlugins: [rehypeSlug, rehypePrettyCode],
  }),
],
```

Everything works just as it does in the svx ecosystem, except `getHeadings` and `getWordCount` because they don't actually transform the AST, they just create metadata about the tree via the VFile.

## What is the new behavior?

This PR allows the markdown template transforms to optionally return a `vfile` instead of a content string. A `vfile` defines its own internal `toString` method that can be used to retrieve the normal content string, but now we also have a way to make any additional metadata available.

The existing markdown transform loop will now use the content string directly as normal if just a string is supplied, but if a vfile is supplied it will invoke the vfiles `toString` method to get the transformed content.

But the markdown template transforms are now also being run a second time, in order to generate the vfile data at the same time that the `metadata` export is created for the agx file. If any vfile data is returned it is added to an additional `vfile` export.

This way you would be able to access any `vfile` metadata data along with the frontmatter `metadata` when using `agx` files.

UPDATE: To make the vfile data easier to access with the existing `injectContent` approach, I modified this to instead merge the vfile data into the metadata export instead of making it its own export. This means with the existing `injectContent` functionality you can just access the vfile data through `attributes.vfile`.

### Example attributes for standard transform that returns a string

Screenshot 2024-09-16 at 12 07 06 pm

### Example attributes for transform that returns a vfile

Screenshot 2024-09-16 at 12 09 36 pm

- [ ] Yes
- [x] No

No, although technically on the off chance that someone happened to already be naming one of their front matter properties `vfile` this update would cause that property to be overwritten

### DIFF
--- a/package.json
+++ b/package.json
@@ -172,6 +172,7 @@
     "ts-morph": "^21.0.1",
     "ts-node": "10.9.1",
     "typescript": "5.4.3",
+    "vfile": "^6.0.3",
     "vite": "^5.3.0",
     "vite-plugin-eslint": "^1.8.1",
     "vite-tsconfig-paths": "4.2.0",
--- a/apps/ng-app/src/app/models.ts
+++ b/apps/ng-app/src/app/models.ts
@@ -1,0 +1,6 @@
+export interface PostAttributes {
+  title: string;
+  slug: string;
+  description: string;
+  coverImage: string;
+}
--- a/apps/ng-app/src/app/pages/posts.[slug].page.analog
+++ b/apps/ng-app/src/app/pages/posts.[slug].page.analog
@@ -5,31 +5,25 @@ import {
 } from '@analogjs/content';
 import { MarkdownComponent } from '@analogjs/content' with { analog: 'imports' }
 import { RouteMeta } from '@analogjs/router';
-import { AsyncPipe, JsonPipe, NgFor, NgIf } from '@angular/common' with { analog: 'imports'};
-import { Component, inject } from '@angular/core';
+import { effect, inject } from '@angular/core';
+import { toSignal } from '@angular/core/rxjs-interop';
 import { map } from 'rxjs';
 
-import { PostAttributes } from './models';
+import { PostAttributes } from '../models';
 
 const renderer = inject(ContentRenderer);
 const post$ = injectContent<PostAttributes>();
 
-const toc$ = post$.pipe(
-  map(() => {
-    return renderer.getContentHeadings();
-  })
-);
+const postSignal = toSignal(post$);
+
+effect(() => {
+  console.log(postSignal());
+})
 </script>
 
-<template><ng-container *ngIf="post$ | async as post">
-  <h1>{{ post.attributes.title }}</h1>
-  <div *ngIf="toc$ | async as toc">
-    <ul>
-      <li *ngFor="let item of toc">
-        <a href="#{{ item.id }}">{{ item.text }}</a>
-      </li>
-    </ul>
-  </div>
-
-  <analog-markdown [content]="post.content"></analog-markdown>
-</ng-container></template>
+<template>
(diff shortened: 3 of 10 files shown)


joshuamorony:
Note: I updated the PR to instead merge the VFile data with the existing metadata export so that it can more easily be accessed through the existing `injectContent` functionality

package.json
```diff
@@ -172,6 +172,7 @@
     "ts-morph": "^21.0.1",
     "ts-node": "10.9.1",
     "typescript": "5.4.3",
+    "vfile": "^6.0.3",
```
brandonroberts, line 175: Does this need to be added as a dependency in the `package.json` for `vite-plugin-angular`?

joshuamorony: Yes I think so, either I could add that or I could just remove the dependency entirely - we don't necessarily care that it's a VFile specifically (just that a VFile would be allowed since it is what would likely be used in most cases), we could just supply our own type like:

```ts
export interface VFileLike {
  toString(encoding?: string | null | undefined): string;
  path?: any;
  data?: any;
  messages?: any;
  history?: any;
  cwd?: any;
}
```

The only potential gotcha here I think is that this `VFileLike` can't really enforce (I don't think) that a custom `toString` method is supplied in the object, since the base `toString` will be inherited if none is supplied and will still satisfy this type.

VFile has a custom `toString` built in that will return the appropriate value.

So if the user isn't using a VFile, and they don't supply their own `toString` implementation within their object, it is going to result in the content string being `[object Object]`.

If we add the `vfile` dependency and force VFile for this perhaps it's a bit safer.

Do you have a preference?

brandonroberts: I think adding it as a dependency is fine

joshuamorony: Ok cool, done

brandonroberts (approved):

PR #1346, opened 2024-09-12
feat(router): introduce server support for form actions

### LINKED ISSUE #935: Feature: Server Form Actions

trpc

### Information

Not sure if this would be a TRPC thing, or router thing.

It would be cool to have server actions. They are executed from a form action element. The goal is not just to submit to a form, but to all the form to have "progressive enhancements" so that they could still be submitted without JavaScript (if the internet is slow), and you get the same results as when you use JavaScript.

https://nextjs.org/docs/app/building-your-application/data-fetching/server-actions-and-mutations#behavior
https://kit.svelte.dev/docs/form-actions#progressive-enhancement
https://start.solidjs.com/core-concepts/actions
https://remix.run/docs/en/main/route/action
https://qwik.dev/docs/action/
https://form-actions-nuxt.pages.dev/get-started/introduction

Just manually send a form to an endpoint without progressive enhancements.  Just throwing this out there to have it on the backlog.

- [ ] Yes
- [ ] No

brandonroberts:
## PR Checklist

Closes #935 

## What is the new behavior?

In `.server.ts` files, an `action` function can be defined to be used with forms to progressively enhance forms and build the foundation for handling form actions on the server.

```ts
// newsletter.server.ts
import {
  type PageServerAction,
  redirect,
  json,
  fail,
} from '@analogjs/router/server/actions';
import { readFormData } from 'h3';

export function load() {
  return {
    loaded: true,
  };
}

export async function action({ event }: PageServerAction) {
  const body = await readFormData(event);
  const email = body.get('email') as string;

  if (!email) {
    return fail(422, { email: 'Email is required' });
  }

  if (email.length < 10) {
    return redirect('/');
  }

  console.log({ email: body.get('email') });

  return json({ type: 'success' });
}
```

A follow-up PR will land to add a form directive to bridge the form to the form server action.

- [ ] Yes
- [x] No


apps/analog-app/src/app/pages/form-action.directive.ts
```diff
@@ -0,0 +1,40 @@
...
+export class FormActionDirective {
+  action = input<string>('');
+  onSuccess = output<unknown>();
+  onError = output<unknown>();
+  state = output<'submitting' | 'error'>();
+  router = inject(Router);
+
+  submitted($event: { target: HTMLFormElement } & Event) {
+    $event.preventDefault();
+    const body = new FormData($event.target);
```
ajitzero, line 24: I didn't know we could do that 🤦 I was using viewChild for this as I couldn't find this info within `$event` 🤷

apps/analog-app/src/app/pages/form-action.directive.ts
```diff
@@ -0,0 +1,40 @@
...
+
+    fetch(this.action() || `/api/_analog/pages${window.location.pathname}`, {
+      method: $event.target.method,
+      body,
+    }).then((res) => {
```
ajitzero, line 29: question: I'm not against `fetch`, but would it be possible to support/use HttpClient? Does the internal implementation even matter? I genuinely don't know.

apps/analog-app/src/app/pages/newsletter.server.ts
```diff
@@ -0,0 +1,53 @@
...
+}
+
+export async function action({
+  event,
+}: PageServerAction): Promise<ActionResult> {
+  const body = await readFormData(event);
+  const email = body.get('email') as string;
+
+  if (!email) {
+    return fail(422, { errors: { email: 'Email is required' } });
```
ajitzero, line 43: question: If JavaScript is disabled, what happens in this case? Can we set up something like `[...not-found].page.ts` but like `[...422].page.ts` to handle this?

ajitzero: I tried handling it manually by adding a `format: 'json'` field if the call is made via Js:
- https://github.com/ajitzero/spartan-repro/blob/feat/form-action/analog-app/src/server/routes/v1/subscribe.ts
- https://github.com/ajitzero/spartan-repro/blob/feat/form-action/analog-app/src/app/pages/subscribe/(home).page.ts#L54

ajitzero (approved):

apps/analog-app/src/app/pages/form-action.directive.ts
brandonroberts: Its possible to use HttpClient, but I don't think it matters. fetch is more flexible out of the box in terms of determining what to do with the response

apps/analog-app/src/app/pages/newsletter.server.ts
brandonroberts: This is still TBD. Ideally, if JS is off, you come back to this page with errors displayed as part of the rendered HTML

jdgamble555:
whahoo!

apps/analog-app/src/app/pages/search.server.ts
```diff
@@ -0,0 +1,35 @@
...
+  const query = getQuery(event);
+  console.log('loaded search', query['search']);
+
+  return {
+    loaded: true,
+    searchTerm: `${query['search']}`,
+  };
+}
+
+export async function action({ event }: PageServerAction) {
```
nartc, line 20: nit: this is copy-pasta so not needed? If this is for demo purposes, can we change the body to something related to the search page instead?

nartc (approved):
Approved with one small nit

apps/analog-app/src/app/pages/search.server.ts
brandonroberts: Oh yea, copy-pasta. Will remove

apps/analog-app/src/app/pages/form-action.directive.ts
```diff
@@ -0,0 +1,78 @@
...
+              this.router.navigate([redirectUrl]);
+            } else if (res.headers.get('Content-type') === 'application/json') {
+              res.json().then((result) => {
+                this.onSuccess.emit(result);
+                this.state.emit('success');
+              });
+            } else {
+              res.text().then((result) => {
+                this.onSuccess.emit(result);
+                this.onSuccess.emit('success');
```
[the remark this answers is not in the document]
brandonroberts: Yep

Jevgenijs92 (approved):

PR #1378, opened 2024-10-02
fix(vite-plugin-nitro): use proxy for API requests instead of custom middleware

brandonroberts:
## PR Checklist

Closes #

## What is the new behavior?

Introduces the `useAPIMiddleware` flag for the `platform` package.

```ts
import analog from '@analogjs/platform';

// https://vitejs.dev/config/
  plugins: [analog({
    useAPIMiddleware: false,
  })],

}));
```
API routes are built under `/api` during development, and `/` in the production build, and requests made to `/api` routes are configured using custom internal middleware. This flag disables the custom internal middleware and uses `routeRules` to proxy the request instead, which allows the API routes to properly preserve cookies and headers with direct requests and redirects. 

- Existing `routeRules` configuration is merged with the internal configuration if provided.
- The prevents a breaking change when used with Cloudflare

- [ ] Yes
- [x] No


andreascorti:
Thanks a lot @brandonroberts for this fix! Really appreciated!

osnoser1:
I was checking what was happening with the headers that weren't being set in prod using Vercel for a couple of days. Initially I thought it was a problem with h3 in Vercel, but in the end, I figured out the problem could be in the nitro vite plugin but I hadn't a chance to keep digging.

I really appreciate this fix @brandonroberts, thanks 😄

PR #1419, opened 2024-10-25
fix(vite-plugin-nitro): enable websocket support & add docs

### LINKED ISSUE #1398: Websocket Support

this is the repo i've tested the `ws` support: [Analog Nitro Websocket](https://github.com/nckirik/analog-nitro-websocket)

vite-plugin-nitro

### Description

I've been working on getting `WebSocket` connections to work on `AnalogJS` for a while.

### The Problem
In development environment, `WebSockets` close unexpectedly on the server side without informing the client. The hooks for `upgrade` and `open` in the `/api/ws/chat` event handler gets executed; however, the connection is then immediately closed with code `1006`, causing the client to timeout.

In production environment, `Websockets` connections get established but there are **no communication** between the client and server.

### Findings
I've found the root problem of this issue and wanted to share some thoughts and get some feedback before diving into a PR.

1. Development Mode Upgrade Handling
    - In dev mode, `vite-nitro-plugin` attaches the Nitro app as middleware to Vite using 
      ```ts
      viteServer.middlewares.use(
        apiPrefix,
        toNodeListener(server.app as unknown as App)
      );
      ```
    - This causes WebSocket `upgrade` requests to be treated as regular API requests *(since there is no `upgrade` handler now)*, leading to the WebSocket hook for the requested route being called, but the socket connection itself never getting fully established. The connection looks like it's immediately closed on the server. But the client never gets notified.
    - To fix this, I made Vite's HTTP server pass the upgrade event back to Nitro:
      ```ts
      viteServer.httpServer?.on("upgrade", server.upgrade);
      ```
    - With this change, the WebSocket connection is successfully established, but the hook for the requested route still doesn't get called *(same as in production)*.  

2. API Prefix
    - In dev, Vite handles all routes (Angular SSR and Nitro) and registers Nitro with an API prefix (by default `/api`). This makes server routes like `/v1/hello` available under `/api/v1/hello`.
    - In production, Nitro handles everything, and the `apiMiddlewareHandler` proxies prefixed routes to their appropriate handlers. However, this middleware doesn't get called for `upgrade` requests, which means routes like `/api/ws/chat` fail to find the WebSocket handler. The WebSocket works fine if accessed directly without the prefix (e.g., `/ws/chat`).
    - Nitro's `routeRules` doesn't affect this behavior either.
    - With the new `upgrade` handler *proposed above*, the behavior is same as in production. Websocket routes won't get the prefix. I think we can live without them for websockets, as long as we document the limitation and provide examples to avoid confusion

3. Dev Environment HMR Conflicts
    - In dev mode, Vite uses the same port for `HMR WebSocket` as the `dev server`. With the new `upgrade` handler *proposed above*, it blocks Vite's own WebSocket connection, causing repeated reconnection attempts and a broken dev environment.
    - To avoid this conflict, we have two options;
      - Change Vite's `HMR port` by default in `vite.config.ts` in the templates
        ```ts
          ...
          server: {
            hmr: {
              // different port than the dev server port
              // which is defined in project.json or angular.json
              port: 5001, 
              path: 'vite-hmr',
            },
          }
          ...
        });
        ```
      - Check if `nitro.experimental.websocket` is `true` to only apply the upgrade fix and document the need of `HMR port change`
        ```ts
        // Websockets are enabled, handle upgrade events
        if (nitroOptions?.experimental?.websocket) {
          viteServer.httpServer?.on("upgrade", server.upgrade);
        }
        ```
    - I think rather than changing the default port, we should check for experimenal websocket support to apply the upgrade handler. We should also document that the `HMR Port` should be changed if websocket support is desired. 

Looking forward to your feedback!

_No response_

### Other information

_No response_

- [X] Yes
- [ ] No

nckirik:
## PR Checklist

Closes #1398 

## What is the new behavior?

* Enables WebSocket support for development environment
* Adds WebSocket documentation

- [ ] Yes
- [x] No

### DIFF
--- a/apps/docs-app/docs/features/api/websockets.md
+++ b/apps/docs-app/docs/features/api/websockets.md
@@ -1,0 +1,116 @@
+# WebSocket
+
+Analog also supports `WebSockets` and `Server-Sent Events` through Nitro.
+
+## Enable WebSockets
+
+Currently, WebSocket support in [Nitro](https://nitro.unjs.io/guide/websocket) is experimental and it should be enabled:
+
+`vite.config.ts`
+
+```typescript
+import { defineConfig } from 'vite';
+import analog from '@analogjs/platform';
+
+export default defineConfig({
+  // ...
+  plugins: [
+    analog({
+      // ...
+      nitro: {
+        experimental: {
+          websocket: true,
+        },
+      },
+    }),
+  ],
+  // ...
+});
+```
+
+**Note:** In development, the Vite HMR WebSocket server runs on the same port as the dev server by default. To prevent conflicts, you need to change this port. The dev server port is usually defined in `project.json`/`angular.json`, which takes precedence over `vite.config.ts`. To allow the port settings in `vite.config.ts` to take effect, remove the port definition from `project.json`/`angular.json`. Additionally, you can specify an optional path to easily differentiate connections in the browser dev tools:
+
+`vite.config.ts`
+
+```typescript
+import { defineConfig } from 'vite';
+import analog from '@analogjs/platform';
+
+export default defineConfig({
+  // ...
+  server: {
+    port: 3000, // dev-server port
+    hmr: {
+      port: 3002, // hmr ws port
+      path: 'vite-hmr', // optional
+    },
+  },
+  // ...
+});
+```
+
+## Defining a WebSocket Handler
+
+Similar to [API routes](/docs/features/api/overview), WebSocket Handlers are defined in the `src/server/routes` folder.
+
+```typescript
+// src/server/routes/ws/chat.ts
+import { defineWebSocketHandler } from 'h3';
+
+export default defineWebSocketHandler({
+  open(peer) {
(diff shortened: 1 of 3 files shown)


apps/docs-app/docs/features/api/websockets.md
```diff
@@ -0,0 +1,116 @@
+# WebSocket
+
+Analog also supports `WebSockets` and `Server-Sent Events` through Nitro.
+
+## Enable WebSockets
```
brandonroberts, line 5:
```suggestion
## Enabling WebSockets
```

brandonroberts (changes requested):

nckirik:
@brandonroberts i just applied the changes but github still says there are changes requested, am i missing anything here?

brandonroberts:
@nckirik nope, you're good. I just hadn't approved them yet

brandonroberts (approved):

PR #1443, opened 2024-11-15
feat(vitest-angular): introduce application bundle-based Vitest builder

brandonroberts:
## PR Checklist

Closes #

## What is the new behavior?

- This introduces a `build-test` builder that uses the application builder function internally to compile the Angular files before running the tests with Vitest. 
- This builder more closely mimics the behavior of the Angular Karma Builder by bundling the files first.
- This builder does not need a `vite.config.mts` file unless further customization is needed.
- This can be a drop-in replacement for `@analogjs/vitest-angular:test` in the future.

```json
    "test": {
      "executor": "@analogjs/vitest-angular:build-test",
      "options": {
        "tsConfig": "libs/my-lib/tsconfig.spec.json",
        "setupFile": "libs/my-lib/src/test-setup.ts"
      }
    },
```

- The current `test` builder can be used with the `@analogjs/vite-plugin-angular` plugin which has its own internal compilation process using the Angular Compiler directly.

- [ ] Yes
- [x] No


packages/vitest-angular/src/lib/builders/build/vitest.impl.ts
```diff
@@ -0,0 +1,216 @@
...
+  return extraArgs;
+}
+
+function findIncludes(options: {
+  workspaceRoot: string;
+  projectRoot: string;
+  include: string[];
+  exclude: string[];
+}) {
+  const fg = require('fast-glob');
```
Yberion, line 157: What do you think about some "better" alternative https://github.com/es-tooling/module-replacements/blob/main/docs/modules/glob.md ([e18e](https://e18e.dev/)).

brandonroberts: Yea, I don't have strong feelings about switching. Happy to discuss in an issue though

pumano:
would love to test it

PR #1444, opened 2024-11-18
perf(vite-plugin-angular): speed up tests by reusing compiler

yjaaidi:
## PR Checklist

The current plugin creates a new compiler for each test file, thus slowing down tests.

P.S.: I am willing to add some performance tests but let's discuss the strategy first 😊

## What is the new behavior?

This PR speeds up tests considerably (3x to 10x)

- [ ] Yes
- [X] No

### DIFF
--- a/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
+++ b/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
@@ -159,6 +159,9 @@ export function angular(options?: PluginOptions): Plugin[] {
 
     return {
       name: '@analogjs/vite-plugin-angular',
+      async watchChange() {
+        await buildAndAnalyze();
+      },
       async config(config, { command }) {
         watchMode = command === 'serve';
         isProd =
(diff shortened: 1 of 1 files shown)


yjaaidi:
https://github.com/analogjs/analog/pull/1443 is an even better alternative.

packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
```diff
@@ -159,6 +159,9 @@ export function angular(options?: PluginOptions): Plugin[] {
 
     return {
       name: '@analogjs/vite-plugin-angular',
+      async watchChange() {
+        await buildAndAnalyze();
```
brandonroberts, line 163: I think it should be wrapped in a `isTest` block so it still only applies when running tests.

```suggestion
      async watchChange() {
        if (isTest) {
          await buildAndAnalyze();
        }
```

brandonroberts (approved):

brandonroberts:
Published in `1.9.5-beta.1`

PR #1539, opened 2025-01-06
fix(vite-plugin-angular): fix HMR of component styles

mattlewis92:
## PR Checklist

Closes #1534

## What is the new behavior?

HMR of component styles works when using an NgModule bootstrapped application or when animations are enabled

Found by looking at the angular CLI source:
https://github.com/angular/angular-cli/blob/9b57ff0552736e6fef44c0ff7e6535f9ed832c6b/packages/angular/build/src/builders/dev-server/vite-server.ts#L142-L143

https://github.com/angular/angular-cli/blob/9b57ff0552736e6fef44c0ff7e6535f9ed832c6b/packages/angular/build/src/tools/esbuild/angular/compiler-plugin.ts#L707

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #1542, opened 2025-01-06
fix(vite-plugin-angular): correctly implement HMR of component styles

mattlewis92:
## PR Checklist

Follow up to https://github.com/analogjs/analog/pull/1539

## What is the new behavior?

In https://github.com/analogjs/analog/pull/1539 I thought I had fixed HMR, but actually what that change did was disable HMR for component stylesheets and trigger a full reload 🤦‍♂️ 

So this followup implements the missing pieces:
1. Implements the appropriate callback for component stylesheets in the `handleHotUpdate` plugin hook to reimplement HMR of component styles
2. Handles `ViewEncapsulation.Emulated` components by using the `encapsulateStyle` helper

- [ ] Yes
- [x] No

Relevant bits from the CLI where I based this code from:
* https://github.com/clydin/angular-cli/blob/602081d18f3c5e0a878f4bf0986461d8b6041130/packages/angular/build/src/builders/dev-server/vite-server.ts#L479-L514
* https://github.com/angular/angular-cli/blob/ddae37fdf3c1c93bd2ae9c51812a9d469b68bd77/packages/angular/build/src/tools/vite/middlewares/assets-middleware.ts#L84-L135

### DIFF
--- a/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
+++ b/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
@@ -367,40 +367,44 @@ export function angular(options?: PluginOptions): Plugin[] {
           const isDirect = ctx.modules.find(
             (mod) => ctx.file === mod.file && mod.id?.includes('?direct')
           );
-          if (pluginOptions.liveReload && isDirect?.id && isDirect.file) {
-            const isComponentStyle =
-              isDirect.type === 'css' && isComponentStyleSheet(isDirect.id);
-            if (isComponentStyle) {
-              const { encapsulation } = getComponentStyleSheetMeta(isDirect.id);
+          if (isDirect) {
+            if (pluginOptions.liveReload && isDirect?.id && isDirect.file) {
+              const isComponentStyle =
+                isDirect.type === 'css' && isComponentStyleSheet(isDirect.id);
+              if (isComponentStyle) {
+                const { encapsulation } = getComponentStyleSheetMeta(
+                  isDirect.id
+                );
 
-              // Track if the component uses ShadowDOM encapsulation
-              // Shadow DOM components currently require a full reload.
-              // Vite's CSS hot replacement does not support shadow root searching.
-              if (encapsulation !== 'shadow') {
-                ctx.server.ws.send({
-                  type: 'update',
-                  updates: [
-                    {
-                      type: 'css-update',
-                      timestamp: Date.now(),
-                      path: isDirect.id,
-                      acceptedPath: isDirect.file,
-                    },
-                  ],
-                });
+                // Track if the component uses ShadowDOM encapsulation
+                // Shadow DOM components currently require a full reload.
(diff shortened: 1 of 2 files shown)


packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
```diff
@@ -14,6 +14,7 @@ import {
   ResolvedConfig,
   Connect,
 } from 'vite';
+import { encapsulateStyle } from '@angular/compiler';
```
brandonroberts, line 17: Do you know what version this was introduced in? We support Angular v17-19

mattlewis92: Looks like it was added in `17.1.0`: https://github.com/angular/angular/commit/f0377d33a1fc1e5ce304a47adf1bf917094cb2e0

I can just make it lazy as we only need it if liveReload is enabled, which will only work in v19+

brandonroberts: Good idea

brandonroberts:
@mattlewis92 something has broken with the styles in the PR preview 

https://deploy-preview-1542--analog-app.netlify.app/

image


https://analog-app.netlify.app/

image

mattlewis92:
> @mattlewis92 something has broken with the styles in the PR preview
> 
> [deploy-preview-1542--analog-app.netlify.app](https://deploy-preview-1542--analog-app.netlify.app/)
> 
> image
> 
> [analog-app.netlify.app](https://analog-app.netlify.app/)
> 
> image

I think this should fix it! [`9ca000a` (#1542)](https://github.com/analogjs/analog/pull/1542/commits/9ca000ab270e1e6b33b4e0ebbaa5bc68686367ba)

brandonroberts:
@mattlewis92 that does fix the build, but the development server is still the same if you run `nx serve analog-app` and go to `http://localhost:3000`

brandonroberts:
I think it was this PR that introduced the regression https://github.com/analogjs/analog/pull/1539

mattlewis92:
> I think it was this PR that introduced the regression #1539

Yep it's definitely that. The problem seems to be specific to component inline `styles` rather than external `styleUrls`

The vite plugin inlines the style into the href of the stylesheet:
CleanShot 2025-01-07 at 11 00 03@2x

Whereas the angular cli creates a virtual file that contains the compiled css:

CleanShot 2025-01-07 at 11 00 13@2x

mattlewis92:
OK this should do the trick! [`691ee70` (#1542)](https://github.com/analogjs/analog/pull/1542/commits/691ee709f1505c387b8de0385b2ab16183563dbb)

brandonroberts:
@mattlewis92 thanks! Will you add this back in this PR? https://github.com/analogjs/analog/pull/1539

I reverted it on beta so we can keep it all as one change

brandonroberts (approved):

PR #1548, opened 2025-01-07
fix(vite-plugin-angular): fix SSG and HMR of external component stylesheets

mattlewis92:
## PR Checklist

Closes #

## What is the new behavior?

Fixes HMR of component `styleUrls` when using SSG. Without this patch, all component stylesheet urls will be added to the DOM twice, and only the first one will be reloaded dynamically so the changes won't appear on the page.

To test it locally, you can make a change in `apps/analog-app/src/app/pages/shipping/shipping.scss` and observe HMR working on `http://localhost:3000/shipping`

- [ ] Yes
- [x] No

### DIFF
--- a/packages/vite-plugin-angular/src/lib/host.ts
+++ b/packages/vite-plugin-angular/src/lib/host.ts
@@ -12,6 +12,7 @@ import { MarkdownTemplateTransform } from './authoring/markdown-transform.js';
 
 import { createRequire } from 'node:module';
 import { createHash } from 'node:crypto';
+import path from 'path';
 
 const require = createRequire(import.meta.url);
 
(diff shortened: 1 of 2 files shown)


packages/vite-plugin-angular/src/lib/host.ts
```diff
@@ -12,6 +12,7 @@ import { MarkdownTemplateTransform } from './authoring/markdown-transform.js';
 
 import { createRequire } from 'node:module';
 import { createHash } from 'node:crypto';
+import path from 'path';
```
brandonroberts, line 15:
```suggestion
import path from 'node:path';
```

brandonroberts (approved):

PR #1587, opened 2025-02-02
fix(vite-plugin-angular): cache resource reads for agx files

joshuamorony:
## PR Checklist

The Angular compiler will trigger `readResource` for every `agx` file whenever a file changes. This becomes a problem, especially for sites with a larger amount of content files, as the markdown transforms re-run for every single content file on every change. With the current implementation, on a site with about 150 content files, this leads to a wait time of around 10 seconds for every change during development.

## What is the new behavior?

This adds caching to `readResource` for agx files. The modified time of the file is checked, and if it is unchanged since the file was last processed it will use the content from the cache.

Now on a file change, the markdown transforms only need to run for the one file that was changed. This also has the small side benefit that since we process agx files twice to generate the `vFile` data, the cached content of the current file being edited can also be used for this rather than running the markdown processing twice (although this benefit is likely negligible)

- [ ] Yes
- [x] No

### DIFF
--- a/packages/vite-plugin-angular/src/lib/host.ts
+++ b/packages/vite-plugin-angular/src/lib/host.ts
@@ -1,7 +1,7 @@
 import { CompilerHost } from '@angular/compiler-cli';
 import { normalizePath } from 'vite';
 
-import { readFileSync } from 'node:fs';
+import { readFileSync, statSync } from 'node:fs';
 import * as ts from 'typescript';
 import { compileAnalogFile } from './authoring/analog.js';
 import {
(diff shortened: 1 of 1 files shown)


packages/vite-plugin-angular/src/lib/host.ts
```diff
@@ -39,6 +39,8 @@ export function augmentHostWithResources(
 ) {
   const ts = require('typescript');
   const resourceHost = host as CompilerHost;
+  const resourceCache = new Map<string, { content: string; mtime: number }>();
```
brandonroberts, line 42: nit: can we move this inside the `if (options.supportAnalogFormat) {` check?

joshuamorony: I don't think so, the `readResource` function is assigned outside of that scope. Or we could assign separate `readResource` functions (e.g. a readResource with the agx specific behaviour inside of the `if` check, and the normal `readResource` in an `else` block

brandonroberts: Ahh, I see. No, I think what you have is fine

brandonroberts (approved):

PR #1590, opened 2025-02-03
fix(vite-plugin-angular): support double quotes in analog import attributes

BaptisteMahe:
## PR Checklist

While using SFCs I had issues with the import attributes only supporting single quotes. I installed Analog in my NX mono repo and eslint is configured for double quotes.
I spent a couple of hours figuring out it was the quote not parsed correctly in the Analog file processor.
Maybe it's deliberate and I'm the only one in this space using double quotes, but I think double quotes should be supported regarding the spec for import attributes.
Also updated the test case.
I'm not sure that the implementation does cover any edge cases or should cover any other edge cases, any advice on this ?

## What is the new behavior?

Double and single are supported in the same way for import attributes.

- [ ] Yes
- [x] No


brandonroberts (approved):

brandonroberts:
Thanks! @allcontributors add @BaptisteMahe for code

PR #1603, opened 2025-02-07
Refactor: Update to New Control Flow, Remove standalone: true, and Make DI Readonly

JeevanMahesha:
## PR Checklist

This PR updates the project to align with Angular v19 improvements:

Updated to new control flow to leverage the latest Angular features.
Removed standalone: true as it is now the default behavior in Angular v19.
Made dependency injection (DI) readonly for better immutability and code safety.

- [ ] Yes
- [X] No

### DIFF
--- a/apps/analog-app/src/app/pages/shipping/index.page.ts
+++ b/apps/analog-app/src/app/pages/shipping/index.page.ts
@@ -6,20 +6,8 @@ import { CartService } from '../../cart.service';
 @Component({
   selector: 'app-shipping',
   imports: [CurrencyPipe, AsyncPipe],
-  template: `
-    <h3>Shipping Prices</h3>
-    @for (shipping of shippingCosts | async; track $index) {
-    <div class="shipping-item">
-      <span>{{ shipping.type }}</span>
-      <span>{{ shipping.price | currency }}</span>
-    </div>
-    }
-  `,
-  styles: `
-  h3 {
-  border: solid 1px darkblue;
-  }
-  `,
+  templateUrl: './shipping.html',
+  styleUrls: ['./shipping.scss'],
 })
 export default class ShippingComponent {
   private readonly cartService = inject(CartService);
--- a/apps/analog-app/src/app/pages/shipping/index.page.ts
+++ b/apps/analog-app/src/app/pages/shipping/index.page.ts
@@ -1,12 +1,10 @@
 import { AsyncPipe, CurrencyPipe } from '@angular/common';
 import { Component, inject } from '@angular/core';
 
-import { Observable } from 'rxjs';
 import { CartService } from '../../cart.service';
 
 @Component({
   selector: 'app-shipping',
-  standalone: true,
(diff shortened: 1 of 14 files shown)


apps/analog-app/src/app/pages/shipping/index.page.ts
```diff
@@ -1,22 +1,31 @@
...
+      <span>{{ shipping.type }}</span>
+      <span>{{ shipping.price | currency }}</span>
+    </div>
+    }
+  `,
+  styles: `
+  h3 {
+  border: solid 1px darkblue;
+  }
+  `,
```
ajitzero, line 24: Hi, can you please run prettier on all these files?

apps/analog-app/src/app/pages/shipping/index.page.ts
```diff
@@ -1,22 +1,31 @@
...
-  private cartService = inject(CartService);
-
-  shippingCosts!: Observable<{ type: string; price: number }[]>;
+export default class ShippingComponent {
+  private readonly cartService = inject(CartService);
 
-  ngOnInit(): void {
-    this.shippingCosts = this.cartService.getShippingPrices();
-  }
+  shippingCosts: Observable<{ type: string; price: number }[]> =
```
ajitzero, line 29: minor: Type can be inferred now since we are directly assigning it. This was needed previously because the assignment happened separately from declaration.
```suggestion
  shippingCosts =
```
Also, this line will need Prettier to be run again.

apps/analog-app/src/app/pages/shipping/index.page.ts
```diff
@@ -1,22 +1,31 @@
...
-import { Component, inject, OnInit } from '@angular/core';
+import { AsyncPipe, CurrencyPipe } from '@angular/common';
+import { Component, inject } from '@angular/core';
 
 import { Observable } from 'rxjs';
 import { CartService } from '../../cart.service';
 
 @Component({
   selector: 'app-shipping',
   standalone: true,
```
ajitzero, line 9: minor: This one was missed.

```suggestion
```

ajitzero (commented):
Hey, thanks for the changes! I left a few comments around Prettier and a missed standalone flag.

Could you also look at other [migrations](https://angular.dev/reference/migrations) to automate this? It seems that you've made these changes manually.

apps/analog-app/src/app/pages/shipping/index.page.ts
ajitzero: Looks like we are still using pre-3.0 prettier (2.8.3). We should take this up with a separate PR. Please ignore this.

apps/analog-app/src/app/pages/shipping/index.page.ts
ajitzero: I've removed my comments regarding prettier and made those changes in !1604 - No impact on your PR here!

PR #1605, opened 2025-02-10
fix(router): properly parse json response in the FormAction

milosdanilov:
## PR Checklist

Fixes the parsing of a JSON response created using the form server actions, used by the FormAction directive.

The problem arose because the `json` function, on the server, sets the Content-Type header to `application/json; charset=utf-8` (see [this line](https://github.com/analogjs/analog/blob/16e6ba9fda0d7d066c7c75aea69e36faf61b48cb/packages/router/server/actions/src/actions.ts#L24)). The additional `charset=utf-8` parameter caused the content-type check to fail, treating the response as plain text. As a result, the response was incorrectly parsed as text instead of JSON.

## What is the new behavior?

This change ensures the response is correctly handled as JSON, even when the `charset=utf-8` parameter is present in the Content-Type header.

- [ ] Yes
- [x] No

## Other information

Additionally, the cart.page in the analog-app is updated passing a success message in the response back to the component, and the tests that check for the message should pass as well.


brandonroberts:
Thanks @milosdanilov, the fix looks good. We already have a `FormAction` example here that ca be updated.  https://github.com/analogjs/analog/blob/beta/apps/analog-app/src/app/pages/newsletter.page.ts 

Would you revert the changes to the cart and update that one instead?

milosdanilov:
@brandonroberts Sure, will do

brandonroberts (approved):

brandonroberts:
Thanks! @allcontributors add @milosdanilov for code

PR #1658, opened 2025-03-20
feat(content): update marked to latest version and fix custom renderer compatibility

### LINKED ISSUE #1420: marked v14 will break code rendering

win 11
node v20
analogv 1.9.0

content

### Description

Since markedjs changed the internal renderer in v13 to a tokenized renderer and removed the old renderer in v14 it will break the rendering process in analog of some markdown parts. If I discover it correctly analogs markdown renderer overrides some of the renderers functions and since the original renderer changed it now passing tokens to the functions wich the override cant convert correctly.

Here is the release notes of the breaking changes intoduces in v13:
https://github.com/markedjs/marked/releases/tag/v13.0.0

```true
It will render code parts as `[object Object]`
```

### Other information

I'm not sure if we/analog should support this since analog is still installing marked v5 on new projects right now

- [ ] Yes
- [ ] No

### LINKED ISSUE #1535: Old marked depency

platform

### Information

I've been updating a an app to the latest angular version, and it seems analog is using an old version of marked-gfm-heading-ids
image
 I need marked 15 or higher, and anything under v4 of marked-gfm-heading-ids is incompatible with marked <13. Any chance you could update this package?

There is no version that overlaps, i tried installing marked-gfm-heading-ids 4.1 in my package.json but analog requires ^3.0.4. 

- [X] Yes
- [ ] No

kudoas:
## PR Checklist

Closes #1420
Closes #1535

## What is the new behavior?

Custom renderer now properly handles tokenized inputs introduced in marked v13+, ensuring consistent markdown rendering.

- [ ] Yes
- [x] No


brandonroberts:
Thanks @kudoas 

If we automatically update the package.json with the updated versions, this won't be breaking right?

kudoas:
@brandonroberts 

> If we automatically update the package.json with the updated versions, this won't be breaking right?

Thanks for pointing that out — you're right, if package.json is updated automatically, it shouldn't be breaking.
Let me know what changes you think are needed, if any.

brandonroberts:
In order to do that, we need to add migrations. Package upgrades for Nx workspaces can be added here

Example: https://github.com/analogjs/analog/blob/beta/packages/content-plugin/migrations.json#L54

Package upgrades for Angular CLI workspaces need a migration.

https://github.com/analogjs/analog/blob/beta/packages/content-plugin/migrations.json#L10

Example: https://github.com/analogjs/analog/blob/beta/packages/content-plugin/src/migrations/update-markdown-renderer-feature/update-markdown-renderer-feature.ts#L56

kudoas:
@brandonroberts Thanks for the pointers.
I implemented the migrations for both Angular CLI and Nx workspaces. b3cf8e819399b3f38af899003b7e08001e83649e

brandonroberts (approved):

PR #1693, opened 2025-04-21
fix hmr issue

### LINKED ISSUE #1683: hmr issue in storybook

Couldn't create a separate env with this problem

vite-plugin-angular

### Description

We have a strange bug that didn't happen in one of the newly set up projects in our nx monorepo that uses vite-plugin-angular for angular storybook, but when we've migrated other projects to it we're experiencing hmr issue, on every change of the source code storybook shows this error: "Cannot read properties of undefined (reading 'getMeta')" that originates from here https://github.com/analogjs/analog/blob/beta/packages/vite-plugin-angular/src/lib/utils/hmr-candidates.ts#L141, and it seems that conditional chaining there solves the problem:
```
const meta = compiler?.getMeta(updatedNode);
```
Do you have any insights on what might cause this problem? And do you think this would be ok as a fix to this problem?

```
Cannot read properties of undefined (reading 'getMeta')

```

### Other information

_No response_

- [x] Yes
- [ ] No

vsternbach:
## PR Checklist

Closes #1683

- [ ] Yes
- [X] No

### DIFF
--- a/packages/vite-plugin-angular/src/lib/utils/hmr-candidates.ts
+++ b/packages/vite-plugin-angular/src/lib/utils/hmr-candidates.ts
@@ -138,7 +138,7 @@ export function analyzeFileUpdates(
       }
 
       // Check for component class nodes
-      const meta = compiler.getMeta(updatedNode);
+      const meta = compiler?.getMeta(updatedNode);
       if (
         meta?.decorator &&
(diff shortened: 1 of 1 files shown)


brandonroberts (approved):

brandonroberts:
Thanks! @allcontributors add @vsternbach for code

PR #1734, opened 2025-05-16
ci: cleanup Zerops build config

fxck:
also fixes pnpm i asking for a yes prompt by adding CI=true to build envs

### DIFF
--- a/zerops.yaml
+++ b/zerops.yaml
@@ -1,30 +1,15 @@
 zerops:
   - setup: docs0
-    # ==== How to build your application ====
     build:
-      # what technologies should the build
-      # container be based on (can be an array)
       base: nodejs@22
-
-      # what commands to use to build your app
+      envVariables:
+        CI: true
       buildCommands:
         - pnpm i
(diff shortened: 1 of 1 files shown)


zerops.yaml
```diff
@@ -1,30 +1,15 @@
...
-      # of deploy path, this example would deploy
-      # files in the dist folder
       deployFiles:
         - dist/apps/docs-app/~
-
-      # *optional*: which files / folders
-      # to cache for the next build run
       cache:
         - node_modules
-        - pnpm-lock.yaml
```
ajitzero, line 26: IMO, we should be caching this. Is there a specific reason for removing it?

fxck: You are supposed to have the latest version of pnpm-lock.yaml (or package-lock.json etc.) commited along your code, which means Zerops fetches it from the repo into the build container anyway (unlike node_modules), so there's no reason to cache it.

brandonroberts (approved):

PR #1754, opened 2025-05-29
fix(vitest-angular): update `@angular-devkit/architect` dep range to include v20

leosvelperez:
## PR Checklist

The `@analogjs/vitest-angular` has a peer dependency range for `@angular-devkit/architect` that doesn't include its v20 version.

Closes #

## What is the new behavior?

The `@analogjs/vitest-angular` supports `@angular-devkit/architect@20`.

- [ ] Yes
- [x] No


brandonroberts:
Thanks! @allcontributors add @leosvelperez for infra

Issue #1698, opened 2025-04-22
Duplicate tRPC request during SSR


DevWedeloper:
### Please provide the environment you discovered this bug in.

```  
"dependencies": {
  "@analogjs/router": "^1.15.1",
  "@analogjs/trpc": "^0.3.0",
  "@angular/animations": "^19.0.0",
  "@angular/common": "^19.0.0",
  "@angular/compiler": "^19.0.0",
  "@angular/core": "^19.0.0",
  "@angular/forms": "^19.0.0",
  "@angular/platform-browser": "^19.0.0",
  "@angular/platform-browser-dynamic": "^19.0.0",
  "@angular/platform-server": "^19.0.0",
  "@angular/router": "^19.0.0",
  "@trpc/client": "^10.45.2",
  "@trpc/server": "^10.45.2",
  "rxjs": "~7.8.0",
  "tslib": "^2.3.0",
  "zone.js": "~0.15.0"
},
"devDependencies": {
  "@analogjs/platform": "^1.15.1",
  "@analogjs/vite-plugin-angular": "^1.15.1",
  "@analogjs/vitest-angular": "^1.15.1",
  "@angular-devkit/build-angular": "^19.0.0",
  "@angular/build": "^19.0.0",
  "@angular/cli": "^19.0.0",
  "@angular/compiler-cli": "^19.0.0",
  "jsdom": "^22.0.0",
  "typescript": "~5.8.0",
  "vite": "^6.0.0",
  "vite-tsconfig-paths": "^4.2.0",
  "vitest": "^3.0.0"
}
```

### Which area/package is the issue in?

trpc

### Description

When using `analogjs/trpc`, a tRPC get (query) request made during SSR results in two calls: one on the server and one again on the client. This doesn't happen when using HttpClient.

### Please provide the exception or error you saw

```true
Two requests are made, one on the server, and one again on the client.
```

### Other information

**Work Around**

Using `pendingUntilEvent()` can prevent the duplicate request by blocking application stability until the data is loaded. However, this feels unintuitive and might not be desirable in all scenarios—especially when some data shouldn't delay the app becoming stable.

**Minimal reproduction**

https://github.com/DevWedeloper/analog-trpc-ssr-double-fetch

### I would be willing to submit a PR to fix this issue

- [ ] Yes
- [x] No

brandonroberts:
tRPC support is in maintenance mode. See #1894

PR #1822, opened 2025-08-03
fix: duplicate tRPC request during SSR

benpsnyder:
Fixes [#1698 - Duplicate tRPC request during SSR](https://github.com/analogjs/analog/issues/1698)

### Key Changes Made:

1. **Early Query Check**: Added an early check to only process query operations (mutations don't need caching)

2. **Client-Side Cache Priority**: On the browser, we now check for cached data FIRST before making any network requests. If data exists in TransferState, we use it immediately and skip the network call.

3. **Simplified Cache Logic**: 
   - Server-side: Always cache queries
   - Client-side: Only cache during initial hydration (when `isCacheActive` is true)

4. **Fixed Import Types**: Updated imports to use `type` imports where appropriate to fix linting warnings

### How It Works Now:

1. **During SSR**: The server executes the tRPC query and stores the result in Angular's TransferState
2. **During Client Hydration**: Before making a network request, the client checks TransferState first. If data exists, it uses that data and removes it from TransferState (to prevent memory leaks)
3. **After Hydration**: Once the app is stable and the cache timeout expires, subsequent queries work normally without TransferState

The solution maintains compatibility with the existing API while fixing the core issue of duplicate network requests during SSR hydration.


packages/trpc/src/lib/client/cache-state.ts
```diff
@@ -22,10 +22,16 @@ export const provideTrpcCacheStateStatusManager = () => ({
...
-    return () =>
-      appRef.isStable
-        .pipe(first((isStable) => isStable))
-        .subscribe(() => cacheState.isCacheActive.next(false));
+    return () => {
+      // Wait for app to be stable, then add a small delay to ensure
+      // all initial tRPC queries have completed before deactivating cache
+      appRef.isStable.pipe(first((isStable) => isStable)).subscribe(() => {
+        // Add a small delay to ensure all initial queries complete
+        timer(100).subscribe(() => {
```
brandonroberts: Nested subscriptions should be avoided.

benpsnyder: Replaced nested subscriptions with switchMap: Instead of subscribing to appRef.isStable and then subscribing to timer(100) inside that subscription, I used switchMap to chain them together.

benpsnyder:
> How can we verify this works as intended?

May I suggest moving this into `alpha` and then I will re-test it again against the sample repo supplied in [#1698 - Duplicate tRPC request during SSR](https://github.com/analogjs/analog/issues/1698)

brandonroberts:
That's fine. We can cherry-pick it into beta after being validated.

> > How can we verify this works as intended?
> 
> May I suggest moving this into `alpha` and then I will re-test it again against the sample repo supplied in [#1698 - Duplicate tRPC request during SSR](https://github.com/analogjs/analog/issues/1698)

benpsnyder:
This PR is being closed and accelerated into this fork: [ESM Transition Plan for AnalogJS v2.0.0-alpha #1844 | 📢⚠️ Fork Alert (for battle testing)](https://github.com/analogjs/analog/issues/1844#issuecomment-3198701307)

Issue #1475, opened 2024-12-05
Strip ng-reflect-* and ng-version in snapshots


k3nsei:
### Which scope/s are relevant/related to the feature request?

vitest-angular

### Information

When switched to Istanbul from V8 as coverage provider our tests stopped passing.
Vitest started generating different snapshots when one of component gets function as its input.

### Describe any alternatives/workarounds you're currently using

Don't try serialize functions when creating snapshots

### I would be willing to submit a PR to fix this issue

- [ ] Yes
- [X] No

brandonroberts:
Thanks @k3nsei. You can try this out manually if you still have `jest-preset-angular` installed.

Add this to the `vite.config.ts`:

```ts
    test: {
      // .. other config
      snapshotSerializers: [
        'jest-preset-angular/build/serializers/no-ng-attributes'
      ]
    },
```

We'll look at supporting this in the package itself.

k3nsei:
I was trying things described here https://marmicode.io/blog/angular-template-code-coverage-and-future-proof-testing and just find out inconsistent behaviour leading to different snapshot each test run. Thats why I'm reporting and asking for it as new feature. I just moved back to V8 as provider because I cannot make our project with not working test suite.

MattiJarvinen:
Starting from Angular 20, ng-reflect-* aren't rendered without provideNgReflectAttributes provider. Atleast we had to refactor tests relying on ng-reflect-* for attributes on stub components in another project when updating to Angular 20.

See https://angular.dev/api/core/provideNgReflectAttributes

brandonroberts:
Ok cool. I'll consider this closed

PR #1824, opened 2025-08-03
fix: strip ng-reflect-* and ng-version in snapshots

benpsnyder:
## Summary

I have successfully fixed the issue described in [GitHub issue #1475](https://github.com/analogjs/analog/issues/1475). Here's what was accomplished:

### Problem
When switching from V8 to Istanbul as the coverage provider, snapshot tests were failing because Angular's `ng-reflect-*` attributes were being included in snapshots. These attributes can vary between different coverage providers, causing snapshot tests to fail when switching between them.

### Solution
I modified the snapshot serialization logic in two key files:

1. **`packages/vitest-angular/setup-snapshots.ts`**
2. **`packages/vite-plugin-angular/setup-vitest.ts`**

### Changes Made
- Added a `stripAngularAttributes` function that recursively removes:
  - All `ng-reflect-*` attributes (the main issue)
  - The `ng-version` attribute (for consistency)
- The function processes the DOM tree recursively to ensure all elements are cleaned
- Applied the stripping logic to the serialized output before returning it

### Verification
✅ **All `ng-reflect-*` attributes have been successfully stripped from snapshots**  
✅ **Snapshot tests are now passing**  
✅ **The fix works regardless of coverage provider**  
✅ **Tests confirmed the fix works with both V8 and Istanbul coverage providers**

The fix ensures that snapshot tests will no longer fail when switching between different coverage providers, as the problematic attributes that vary between providers are now consistently stripped from the serialized output.

## Breaking Changes

This change modifies the snapshot serialization behavior in both `@analogjs/vitest-angular` and `@analogjs/vite-plugin-angular` packages. Snapshot tests will now have Angular-specific attributes stripped from the serialized output.

### BEFORE:

Snapshots included Angular-specific attributes that could vary between coverage providers:

```html
<div ng-reflect-ng-if="true" ng-reflect-ng-class="[object Object]" ng-version="17.0.0">
  <span>Hello World</span>
</div>
```

### AFTER:

Snapshots now have Angular-specific attributes stripped for consistency:

```html
<div>
  <span>Hello World</span>
</div>
```

**Impact**: Existing snapshot tests may need to be updated to reflect the new serialization format. The stripped attributes (`ng-reflect-*` and `ng-version`) are no longer included in snapshot comparisons, which should make tests more reliable when switching between different coverage providers.


libs/card/src/lib/autocomplete/__snapshots__/autocomplete.component.spec.ts.snap
```diff
@@ -3,47 +3,41 @@
 exports[`GIVEN AutocompleteComponent > WHEN input value is empty > AND input is focused > THEN component should match snapshot 1`] = `
 <div
   id="root2"
-  ng-version="19.0.4"
+  ng-version="20.1.4"
```
brandonroberts: Shouldn't the `ng-version` be removed?

brandonroberts (commented):
Breaking changes should target the `alpha` branch and use the correct format in the PR

https://analogjs.org/docs/contributing#breaking-changes

libs/card/src/lib/autocomplete/__snapshots__/autocomplete.component.spec.ts.snap
benpsnyder: @brandonroberts resolved

packages/nx-plugin/src/generators/app/generator.ts
```diff
@@ -26,6 +26,7 @@ import { addTrpc } from './lib/add-trpc';
 import { cleanupFiles } from './lib/cleanup-files';
 import { addAnalogProjectConfig } from './lib/add-analog-project-config';
 import { updateIndex } from './lib/update-index-html';
+import { addAnalogDependencies } from './lib/add-analog-dependencies';
```
brandonroberts: Revert these changes as they aren't related to this PR

packages/vitest-angular/setup-snapshots.ts
```diff
@@ -132,7 +132,41 @@ function fixtureVitestSerializer(fixture: any) {
...
+  stripAngularAttributes(result);
+
+  return result;
+}
+
+/**
+ * Recursively strip ng-reflect-* and ng-version attributes from DOM elements
+ * @param element The DOM element to process
+ */
+function stripAngularAttributes(element: any) {
```
brandonroberts: Let's return the `element` instead of mutating the result

packages/vitest-angular/setup-snapshots.ts
```diff
@@ -132,7 +132,41 @@ function fixtureVitestSerializer(fixture: any) {
     'text/html',
   );
 
-  return doc.body.childNodes[0];
+  const result = doc.body.childNodes[0];
+
+  // Strip ng-reflect-* and ng-version attributes from the serialized output
+  stripAngularAttributes(result);
```
brandonroberts:
```suggestion
  const strippedResult = stripAngularAttributes(result);
```

packages/vitest-angular/setup-snapshots.ts
```diff
@@ -132,7 +132,41 @@ function fixtureVitestSerializer(fixture: any) {
...
+
+  return result;
```
brandonroberts:
```suggestion
  return strippedResult;
```

packages/vite-plugin-angular/setup-vitest.ts
```diff
@@ -201,7 +201,41 @@ function fixtureVitestSerializer(fixture: any) {
     'text/html',
   );
 
-  return doc.body.childNodes[0];
+  const result = doc.body.childNodes[0];
```
brandonroberts: We can revert these changes. This will be removed in the next major

packages/nx-plugin/src/generators/app/generator.ts
benpsnyder: It breaks the build. Need to then bump package versions. Doing it now so you can see.

packages/vite-plugin-angular/setup-vitest.ts
benpsnyder: Revert all changes in this whole file?

packages/nx-plugin/src/generators/app/generator.ts
benpsnyder: Please recommend the fix to this: https://cloud.nx.app/runs/VzcQD62sRo/task/nx-plugin%3Atest

packages/vite-plugin-angular/setup-vitest.ts
brandonroberts: Yes

benpsnyder: done

PR #1862, opened 2025-08-31
fix(vite-plugin-angular): add back defaults for `resolve.conditions`

### LINKED ISSUE #1847: vite-plugin-angular potentially breaks vite's export resolution

vite 6+

vite-plugin-angular

### Description

In vite 6 a [breaking change](https://v6.vite.dev/guide/migration.html#default-value-for-resolve-conditions) was introduced. `vite-plugin-angular` [did not migrate this](https://github.com/analogjs/analog/blob/45f0f1a9e33a3d8619a9093b4fe402bd789cb492/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts#L213), hence potentially causing issues as shown below.

As a workaround one can add
```
import { defaultClientConditions } from 'vite';
...
resolve: {
  conditions: [...defaultClientConditions]
}
```
to the vite config, though this does mix up the order of the lookup.

```true
Uncaught TypeError: Cannot destructure property '__extends' of 'import_tslib.default' as it is undefined.

and

Uncaught TypeError: import_tslib.default is undefined
```

### Other information

_No response_

- [x] Yes
- [ ] No

st3h3n:
## PR Checklist

Closes #1847

## What is the new behavior?

Defaults are added back. Note that I didn't add tests, because I have no idea how that could be tested or whether it even should be. Also I didn't investigate whether adding `'style'` is even necessary and just assumed it to be correct. If it isn't, the entire block could be removed instead, as there is a fallback to the defaults in that case.

- [x] Yes, in theory
- [x] No, in practice

The fact that this was not done when upgrading to Vite 6 was a breaking change. So if at all this could only be a breaking change for people who somehow started relying on the broken behavior since then. Because people might have used now obsolete workarounds, it's definitely worth adding a note though.

### DIFF
--- a/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
+++ b/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
@@ -13,6 +13,7 @@ import {
   ViteDevServer,
   preprocessCSS,
   ResolvedConfig,
+  defaultClientConditions,
 } from 'vite';
 import * as ngCompiler from '@angular/compiler';
 
@@ -210,7 +211,7 @@ export function angular(options?: PluginOptions): Plugin[] {
             },
           },
           resolve: {
-            conditions: ['style'],
+            conditions: ['style', ...defaultClientConditions],
           },
         };
       },


brandonroberts:
Thanks @st3h3n. This actually fixes an issue we were having in another repo with the Medusa JS SDK where it wasn't importing ESM by default.

st3h3n:
Oh yea, I could have added that to the description, because that turned out to be the same problem for me. The tslib exceptions mentioned in the issue, were caused because `default: "./tslib.js"` was imported rather than `module.default: "./tslib.es6.mjs"`. Well I guess now you can remove an obsolete workaround just like I mentioned above 😁.

PR #1892, opened 2025-10-10
feat(vite-plugin-angular): add support for tsconfig as getter function

shairez:
## PR Checklist

Recently, I've made buildable libs available for vitest in Nx (to improve testing speed in large nx monorepos)

In order to be able to have the angular compiler work with the tsconfig generated by Nx (with the paths pointing to the `dist/` folder), a more dynamic way to get the tsconfig path is needed in the angular-vite plugin.

## What is the new behavior?

My change is simple, it allows the user to either provide a string (like the current implementation) or to provide a getter function. 

That way, during the angular compilation, once Nx already generated the "dist" tsconfig, we can make sure that generated tsconfig path is being passed to the `performCompilation` function, and by that the angular compiler could use the already built libs paths (because the getter runs dynamically and not only when configuring the plugin)

- [ ] Yes
- [x] No

### DIFF
--- a/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
+++ b/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
@@ -1,4 +1,5 @@
 import { CompilerHost, NgtscProgram } from '@angular/compiler-cli';
+import { mkdirSync, writeFileSync } from 'node:fs';
 import {
   basename,
   dirname,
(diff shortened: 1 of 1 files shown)


shairez:
working on fixing the tests now

shairez:
@brandonroberts it's not as clean as I hoped for, but there's an issue I tried to avoid. 

The Nx plugin is generating the tsconfig during the `configResolved` hook. 

So I needed to find a way to resolve the tsconfig twice, both in `config` and in `configResolved`. 

It's not pretty looking, but maybe you have a different idea on how to solve it in a cleaner way. 
If so, I'd be happy to make the changes.

PR #1929, opened 2025-11-03
fix(docs): add `netlify init` configuration steps for Netlify

serhalp:
## PR Checklist

The Netlify deployment docs currently only document a manual configuration approach using the Netlify app's UI.

## What is the new behavior?

Using the Netlify CLI instead automatically configures the build settings, so this PR documents this as the preferred configuration approach, documents the `netlify.toml` that this generates, and moves the manual UI-based configuration to a "Manual configuration" subsection.

- [ ] Yes
- [x] No

## Other information

This is out of scope for this PR but... FYI we'd love to be able to not need that manual functions dir step. Our current framework auto-configuration system doesn't support populating this field. We've never needed this for any other framework, because our assumption is that frameworks will respect the [Netlify Frameworks API](https://docs.netlify.com/build/frameworks/frameworks-api/) when targeting Netlify. Since you're using Nitro (which does this in its Netlify presets), I'm curious why this is ending up in `dist/analog`? 🤔

### DIFF
--- a/apps/docs-app/docs/features/deployment/providers.md
+++ b/apps/docs-app/docs/features/deployment/providers.md
@@ -140,12 +140,41 @@ You can also check out [Github Integration](https://docs.zerops.io/references/gi
 
 Analog supports deploying on [Netlify](https://netlify.com/) with minimal configuration.
 
-### Deploying the Project
+### Deploying the project
 
 <Tabs groupId="porject-type">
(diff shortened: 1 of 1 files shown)


brandonroberts:
Thanks @serhalp!  We would definitely like to make it zero config if possible. I thought there was some auto-detection set up for Analog at some point. We'll likely revisit the output directory when nitro v3 goes stable.

PR #1964, opened 2025-11-20
feat(vite-plugin-angular): add support for rolldown-vite plugins

brandonroberts:
## PR Checklist

Closes #

## What is the new behavior?

When using `rolldown-vite`, the plugins use `rolldownOptions` and `oxc` options for dev/build/test/transpilation instead of `rollup` and `esbuild` equivalents.

- [ ] Yes
- [x] No

### DIFF
--- a/package.json
+++ b/package.json
@@ -196,7 +196,7 @@
     "ts-morph": "^26.0.0",
     "ts-node": "10.9.2",
     "typescript": "5.9.3",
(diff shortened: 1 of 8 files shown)


osnoser1:
I tried to use this plugin with tsdown and I had a couple of issues, I wonder if with this update would make easier the integration, I'll check it in a bit 😄

PR #1995, opened 2025-12-07
fix(vitest-angular): reset `TestBed` between tests in Vitest Browser Mode

yjaaidi:
Vitest Browser Mode caches modules and does not re-execute side-effects.
Therefore, when hooks are cleared, they are not set up again so TestBed
is not reset after each test.

Vitest was fixed to allow side-effects on setup files by
not caching them (see https://github.com/vitest-dev/vitest/pull/8884),
but the behavior is not recursive so hooks that are
registered as side effects in other files such as
`@analogjs/vitest-angular/setup-testbed.ts` will not work.

The solution is to register these hooks inside the `setUsetupTestBed` function.


libs/my-package/project.json
```diff
@@ -25,8 +25,10 @@
       }
     },
     "test": {
-      "executor": "@nx/vite:test",
```
yjaaidi, line 28: `@nx/vite:test` does not seem to support browser mode.

It would be nice to use the Nx vitest plugin instead of the manual command I've added/

packages/vitest-angular/setup-testbed.ts
```diff
@@ -22,6 +19,9 @@ type TestBedSetupOptions = {
 export function setupTestBed(
   options: TestBedSetupOptions = { zoneless: true, providers: [] },
 ) {
+  beforeEach(getCleanupHook(false));
```
yjaaidi, line 22: Here, we're sure the hooks will be registered.

package.json
```diff
@@ -192,17 +192,18 @@
...
     "ts-node": "10.9.2",
     "typescript": "5.9.3",
-    "vite": "^8.0.0-beta.0",
+    "vite": "npm:rolldown-vite@latest",
     "vite-plugin-eslint": "^1.8.1",
     "vite-plugin-inspect": "11.3.2",
     "vite-tsconfig-paths": "4.2.0",
     "vitefu": "^1.1.1",
-    "vitest": "^4.0.0",
+    "vitest": "^4.0.15",
```
yjaaidi, line 206: Bumped vitest to sync with `@vitest/browser-playwright`.

libs/my-package/src/lib/my-package/reset-test-bed-between-tests.spec.ts
```diff
@@ -0,0 +1,20 @@
+import { InjectionToken } from '@angular/core';
```
yjaaidi: I've added my test in this package again.
If you're ok with it, I'd like to move or copy this package to something like `tests/vitest-angular-tests` to make it more things more explicit.

libs/my-package/project.json
brandonroberts: Switching to the `@nx/vitest` package is fine after Nx 22.2 is released.

libs/my-package/src/lib/my-package/reset-test-bed-between-tests.spec.ts
brandonroberts: Moving the tests sounds reasonable to me

packages/vitest-angular/setup-testbed.ts
brandonroberts: Did you see any tests slow down by putting them here? Or does it have no impact with the Vitest fixes

brandonroberts:
@yjaaidi just needs merge conflicts resolved.

packages/vitest-angular/setup-testbed.ts
yjaaidi: No slowdown observed, and it shouldn't slow down anything.

libs/my-package/src/lib/my-package/reset-test-bed-between-tests.spec.ts
yjaaidi: I'll do that in a distinct PR then.

yjaaidi:
> @yjaaidi just needs merge conflicts resolved.

Really weird, I can't reproduce the conflict locally 😅.
I've reset it to `main` and updated it, but github is still complaining.
Seems like some caching issue or I'm getting something wrong.

Any clues? Should I open another PR?

Update: 🤦🏻‍♂️ I was targeting `beta` instead of `main`

yjaaidi:
P.S.: I've updated the tests to avoid false negatives where the tests could pass when using multiple workers, or disabling/removing other tests such as the `aot` one. We need two test files to reproduce the bug.

PR #1998, opened 2025-12-08
fix(storybook-angular): creates config.plugin array when undefined

### LINKED ISSUE #1992: [storybook-angular] Cannot read properties of undefined (reading 'flat')

(My apologies for not being able to provide a reproduction. The StackBlitz example is having a completely separate issue at the moment: `Error: Cannot load native addon because loading addons is disabled`)

storybook-angular

### Description

When running `vitest` in browser mode to test stories with `@analogjs/storybook-angular` an error is thrown.

```true
TypeError: Cannot read properties of undefined (reading 'flat')
```

### Other information

The error comes from this line:

https://github.com/analogjs/analog/blob/0b255202253a418f88717c6e178f0bb179725e3c/packages/storybook-angular/src/lib/preset.ts#L39-L43

I haven't been able to determine exactly why the value is undefined in my project vs the example project, but it works as expected by making this change:

```diff
  export const viteFinal = async (config, options) => {
+   config.plugins ??= [];
    // Remove any loaded analogjs plugins from a vite.config.(m)ts file
    config.plugins = config.plugins
      .flat()
      .filter((plugin) => !plugin.name.includes('analogjs'));
```

- [x] Yes
- [ ] No

shannonmoeller:
## PR Checklist

Ensures the `config.plugins` array exists if it's missing to prevent the following error:

```
TypeError: Cannot read properties of undefined (reading 'flat')
```

Closes #1992

## What is the new behavior?

No new behavior. Just fixes an edge case.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/storybook-angular/src/lib/preset.ts
+++ b/packages/storybook-angular/src/lib/preset.ts
@@ -38,7 +38,7 @@ export const core = async (config, options) => {
 };
 export const viteFinal = async (config, options) => {
   // Remove any loaded analogjs plugins from a vite.config.(m)ts file
-  config.plugins = config.plugins
+  config.plugins = (config.plugins ?? [])
     .flat()
     .filter((plugin) => !plugin.name.includes('analogjs'));
 


shannonmoeller:
Thank you!

PR #2025, opened 2025-12-19
fix(storybook-angular): resolve relative styles from project root

### LINKED ISSUE #2023: [Bug] Storybook global styles are not resolved correctly from project root
> And another problem I found is that global styles (from `styles` option in `angular.json`) are not resolved correctly if `.storybook` directory is not located in the root of the project, but somewhere else.
> 
> In this case `npm run storybook` reports
> ```
> Vite Internal server error: Failed to resolve import
> │  "src/stories/styles/global.scss" from "src/.storybook/preview.ts". Does the file
> │  exist?
> ```
> and `npm run build-storybook` reports
> ```
> [vite]: Rollup failed to resolve import "src/stories/styles/global.scss" from
> │  "./src/.storybook/preview.ts".
> ```
> Using `/src/...` instead of `src/...` doesn't fix the issue either.
> 
> I think we should resolve proper import path for each global style like `path.resolve(<location_of_angular_json>, <project_root_from_angular_json>, <global_style_path>)` instead of just taking paths from the config as-is.
> 
> I've reproduced the problem in the same repo in the `global-styles-import-problem` branch: https://github.com/th0r/ng-analogjs-storybook-angular-styles-bug/tree/global-styles-import-problem 

 _Originally posted by @th0r in [#2018](https://github.com/analogjs/analog/issues/2018#issuecomment-3675614332)_

brandonroberts:
## PR Checklist

Closes #2023 

## What is the new behavior?

- Relative styles (`src`) are resolved from the workspace root + project root + styles path.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/storybook-angular/src/lib/preset.ts
+++ b/packages/storybook-angular/src/lib/preset.ts
@@ -120,7 +120,7 @@ function angularOptionsPlugin(options, { normalizePath }) {
 
       return;
     },
-    transform(code, id) {
+    async transform(code, id) {
       if (
         normalizePath(id).endsWith(
           normalizePath(`${options.configDir}/preview.ts`),
@@ -141,9 +141,27 @@ function angularOptionsPlugin(options, { normalizePath }) {
           imports.push('zone.js');
         }
 
+        const projectConfig =
+          await options.angularBuilderContext.getProjectMetadata(
+            options.angularBuilderContext.target.project,
+          );
+
         return {
           code: `
-            ${imports.map((extraImport) => `import '${extraImport}';`).join('\n')}
+            ${imports
+              .map((extraImport) => {
+                if (
+                  extraImport.startsWith('.') ||
+                  extraImport.startsWith('src')
+                ) {
+                  // relative to root
+                  return `import '${resolve(options.angularBuilderContext.workspaceRoot, projectConfig.root, extraImport)}';`;
+                }
+
+                // absolute import
+                return `import '${extraImport}';`;
+              })
+              .join('\n')}
             ${code}
           `,
         };


packages/storybook-angular/src/lib/preset.ts
```diff
@@ -141,9 +141,27 @@ function angularOptionsPlugin(options, { normalizePath }) {
...
+                  extraImport.startsWith('.') ||
+                  extraImport.startsWith('src')
+                ) {
+                  // relative to root
+                  return `import '${resolve(options.angularBuilderContext.workspaceRoot, projectConfig.root, extraImport)}';`;
+                }
+
+                // absolute import
+                return `import '${extraImport}';`;
+              })
```
th0r, line 163: Hey @brandonroberts, actually there is no need in extra conditions here as `path.resolve('/abosulte_path_to_project', '/absolute_path_to_style')` will resolve to `/absolute_path_to_style` so all this code code can be replace with a simple return `import '${resolve(options.angularBuilderContext.workspaceRoot, projectConfig.root, extraImport)}';` and it will cover all the cases.

th0r: For example in our project all the code is located not in the `src` directory, so specifying `styles: ['path/to/our/src/global.scss']` still won't work. I understand that we could start the path with a dot, but why introduce these confusing rules if they can we avoided. 🤷‍♂️

brandonroberts: Always use the resolved path won't cover importing styles from `node_modules` for example. You should be able to use a relative path to import from the correct path if its outside the project root.

th0r: Hm, why? If project root is `/home/user/project` then `path.resolve('/home/user/project', '../node_modules')` will be resolved to `/home/user/node_modules`. Isn't it what we want?

brandonroberts: In Nx workspaces, paths are resolved from the root, so we don't want to always resolve from the project root

PR #2033, opened 2026-01-08
fix(storybook): use vite config root when angularBuilderContext unavailable

### LINKED ISSUE #2032: angularBuilderContext is undefined when running Storybook tests via @storybook/addon-vitest

- @analogjs/storybook-angular: 2.2.1
- - @storybook/addon-vitest: 10.1.11
- - - storybook: 10.1.11
- - - - vitest: 4.0.16
- - - - - @angular/core: 21.0.6
- - - - - - Node: 24.x
- - - - - - - Package manager: pnpm 10.x
- - - - - - - - Monorepo: Nx 22.3.3

Unfortunately I was unable to reproduce this in StackBlitz - it seems to require a more complex project setup with the vitest addon configured.

storybook-angular

## Summary

When running Storybook component tests via `@storybook/addon-vitest`, the `angularOptionsPlugin` in `preset.js` crashes because `angularBuilderContext` is undefined.

## Root Cause

When running via Vitest (not Angular builders), there's no `angularBuilderContext` available since Vitest doesn't use Angular's build system. The `transform` function in `angularOptionsPlugin` attempts to access `options.angularBuilderContext.getProjectMetadata()` without checking if `angularBuilderContext` exists first.

## Location

The issue is in `@analogjs/storybook-angular/src/lib/preset.js` around line 116, in the `angularOptionsPlugin` function's `transform` method.

## Suggested Fix

Add a guard at the beginning of the `transform` function to skip transformation when `angularBuilderContext` is not available:

```javascript
async transform(code, id) {
  // Skip transformation when angularBuilderContext is not available
  // (e.g., when running via Vitest instead of Angular builders)
  if (!options.angularBuilderContext) {
    return;
  }
  // ... rest of existing transform logic
}
```

## Workaround

Currently using a pnpm patch to add this guard. Happy to submit a PR with this fix if you'd like!

```true
TypeError: Cannot read properties of undefined (reading 'getProjectMetadata')
    at Object.transform (node_modules/@analogjs/storybook-angular/src/lib/preset.js:116:65)
```

### Other information

If you wish I can provide a PR if the suggested fix sounds okay. If you don't approve I can attempt to find and alternative solution with your assistance and then provide the PR, if you wish.

This issue only occurs when using `@storybook/addon-vitest` to run Storybook stories as Vitest tests. Running Storybook normally (via `ng run app:storybook` or Angular builders) works fine because the builder provides the `angularBuilderContext`.

The issue is that when running via Vitest, there's no Angular builder in the execution chain, so `angularBuilderContext` is never populated.

Note: This is a different issue from #1832, which addressed `process is not defined` errors.

---
*This issue was drafted with assistance from [Claude Code](https://claude.ai/code).*

- [x] Yes
- [ ] No

---  
Since this was mainly written with Claude I personally just wanted to take a moment to tell you, thank you for the great job you are doing. You are truly an inspiration, it's been great following your journey :) Keep it up!

palexcast:
## Summary

When running Storybook tests via `@storybook/addon-vitest`, the `angularBuilderContext` is undefined since Vitest doesn't use Angular's build system. This caused a crash in the `angularOptionsPlugin`:

```
TypeError: Cannot read properties of undefined (reading 'getProjectMetadata')
```

## Solution

Following @brandonroberts' suggestion in #2032:

1. Store the Vite `userConfig` from the `config()` hook in a `resolvedConfig` variable
2. Use `config.root` as the project root when `angularBuilderContext` is unavailable
3. Maintain backwards compatibility when `angularBuilderContext` IS available

**Fallback chain:** `angularBuilderContext.workspaceRoot` → `userConfig.root` → `process.cwd()`

## Testing

Tested in a real Nx monorepo with:
- `@analogjs/storybook-angular: 2.2.1`
- `@storybook/addon-vitest: 10.1.11`
- `vitest: 4.0.16`
- `@angular/core: 21.0.6`

All Storybook tests pass via Vitest after this fix.

Fixes #2032

### DIFF
--- a/packages/storybook-angular/src/lib/preset.spec.ts
+++ b/packages/storybook-angular/src/lib/preset.spec.ts
@@ -1,0 +1,93 @@
+import { describe, it, expect, vi } from 'vitest';
+import { viteFinal } from './preset';
+
+describe('viteFinal', () => {
+  const createMockOptions = (overrides = {}) => ({
+    configDir: '.storybook',
+    presets: {
+      apply: vi.fn().mockResolvedValue({ options: {} }),
+    },
+    angularBuilderOptions: {},
+    ...overrides,
+  });
+
+  describe('when angularBuilderContext is undefined', () => {
+    it('should not crash during plugin initialization', async () => {
+      const config = { plugins: [] };
+      const options = createMockOptions({
+        // angularBuilderContext is intentionally undefined
+      });
+
+      await expect(viteFinal(config, options)).resolves.not.toThrow();
+    });
+
+    it('should return a valid vite config', async () => {
+      const config = { plugins: [] };
+      const options = createMockOptions();
+
+      const result = await viteFinal(config, options);
+
+      expect(result).toBeDefined();
+      expect(result.plugins).toBeDefined();
+      expect(Array.isArray(result.plugins)).toBe(true);
+    });
+
+    it('should include the angular options plugin', async () => {
+      const config = { plugins: [] };
+      const options = createMockOptions();
+
+      const result = await viteFinal(config, options);
+
+      const angularOptionsPlugin = result.plugins
+        .flat()
+        .find((p) => p?.name === 'analogjs-storybook-options-plugin');
+      expect(angularOptionsPlugin).toBeDefined();
+    });
+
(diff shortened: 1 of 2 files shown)


brandonroberts (approved):
LGTM

brandonroberts:
Thanks @palexcast!

PR #2053, opened 2026-02-03
fix(vite-plugin-angular): fall back to oxc/esbuild when Angular compiler has no output for a file

Kenzo-Wada:
## PR Checklist

The transform handler returned `{ code: '', map: null }` for `.ts` files not emitted by the Angular compiler. Vite interpreted this as "the plugin handled the file and the result is empty", so no other transformer processed the file.
This caused Vitest to report "No test suite found in file" for any `.spec.ts` file — including plain TypeScript tests with no Angular imports.

## What is the new behavior?

When `fileEmitter(id)` returns `undefined`, the transform handler now falls back to `transformWithOxc` (Vite 7+) or `transformWithEsbuild` to compile the TypeScript file, instead of returning empty code.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
+++ b/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
@@ -649,8 +649,22 @@ export function angular(options?: PluginOptions): Plugin[] {
             this.error(`${typescriptResult.errors.join('\n')}`);
           }
 
-          // return fileEmitter
-          let data = typescriptResult?.content ?? '';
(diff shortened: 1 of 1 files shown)


brandonroberts:
Thanks @Kenzo-Wada . I'd like to understand this better.

I've tested this before also and it causes false positives in the case where the files should actually be handled by the Angular Compiler but they are missing from the tsconfig or the tsconfig isn't loaded correctly, causing the files to be missing from the compilation.

What issues are you running into?

PR #2054, opened 2026-02-04
feat(vitest-angular): add `teardown.destroyAfterEach` option and deprecate `browserMode` option

yjaaidi:
## PR Checklist

The `browserMode` option can be misleading.
Toggling off `destroyAfterEach` has a side effect of making destruction errors ambiguous. For instance, if a component throws an error when destroyed, then it is the next test that will fail which can make it harder to find the root cause.

Therefore, users should only use this option when debugging.

## What is the new behavior?

- Deprecate `browserMode` option
- Add `teardown.destroyAfterEach` (which aligns partially with TestBed options)

- [ ] Yes
- [X] No

### 2. Toggle off `destroyAfterEach` when running in headed mode

Such behavior is only needed when running tests in a headed browser.
The package could detected headed mode and toggle off this option.
IMO, this would be too surprising and can cause discrepancies between headed and headless mode.

### DIFF
--- a/packages/vitest-angular/setup-testbed.ts
+++ b/packages/vitest-angular/setup-testbed.ts
@@ -22,7 +22,7 @@ type TestBedSetupOptions = {
   /**
    * @deprecated Use `teardown.destroyAfterEach` instead.
    * @sunset 3.0.0
-   */  
+   */
   browserMode?: boolean;
   teardown?: {
     destroyAfterEach: boolean;


yjaaidi:
This is my bad.
@shairez pinged me about the PR #2012 introducing `browserMode` option but I couldn't react fast enough.

shairez:
@yjaaidi thanks
so how would you set this up automatically for headed mode?

yjaaidi:
Thank you for the original PR!

I guess it could be achieved like this:

```ts
import { server } from 'vitest/browser';

const isHeadless = server.config.browser.headless;
```

The problem is that such heuristics can be brittle. I can already imagine issues such as "my tests break with headed browser mode".

shairez:
> Thank you for the original PR!
> 
> I guess it could be achieved like this:
> 
> ```ts
> import { server } from 'vitest/browser';
> 
> const isHeadless = server.config.browser.headless;
> ```
> 
> The problem is that such heuristics can be brittle. I can already imagine issues such as "my tests break with headed browser mode".

Thanks!

I wonder if we could offer a documented easy way [in the README](https://github.com/analogjs/analog/pull/2054/changes#diff-2a7212c2a5706048836851019e2e96b3e2b5026cf9b7920fde41d79100ced0bfR255) to make it more automatic for the end users

brandonroberts:
What's the verdict on this PR?

yjaaidi:
What do you think @shairez? I feel that the `teardown.destroyAfterEach` option is the least surprising behavior. Users can also easily find more about it in the Angular docs and community resources without having to dive into the code.

shairez:
@brandonroberts @yjaaidi sorry for the delay... things have been wild around here :sweat_smile: ...

OK cool let's update the name, as long as we keep recommending to use it for headed mode, and add this `isHeadless` to the README to let users configure it themselves, LGTM

PR #2070, opened 2026-02-17
Add imports and update providers in setupTestBed

prclm:
## PR Checklist

setupTestBed() now works with (service) providers and imports (NgModules).

Closes #

## What is the new behavior?

the "providers" in setupTestBed() was wrongly passed to initTestEnvironment() which only expects ngModules.

Now a CustomTestModule is created to support "real" providers, as well as imports in the setupTestBed().

- [x ] Yes
- [ ] No

"providers" must be renamed to "imports" in the setupTestBed() call.

### DIFF
--- a/packages/vitest-angular/setup-testbed.ts
+++ b/packages/vitest-angular/setup-testbed.ts
@@ -1,4 +1,11 @@
-import { EnvironmentProviders, inject, NgModule, Provider, provideZonelessChangeDetection, Type } from '@angular/core';
+import {
+  EnvironmentProviders,
+  inject,
+  NgModule,
+  Provider,
+  provideZonelessChangeDetection,
+  Type,
+} from '@angular/core';
 import {
   ɵgetCleanupHook as getCleanupHook,
   getTestBed,
@@ -13,7 +20,7 @@ const ANGULAR_TESTBED_SETUP = Symbol.for('testbed-setup');
 
 type TestBedSetupOptions = {
   zoneless?: boolean;
-  imports?: Type<any> | Type<any>[];
+  imports?: Type<any>[];
   providers?: (Provider | EnvironmentProviders)[];
   browserMode?: boolean;
 };


brandonroberts:
Thanks @prclm. Let's keep the NgModule setup as-is and just add the `imports` to the `TestModule`.

yjaaidi:
Hey @prclm, I stumbled upon the `providers` issue today then I came up with #2072. I tried to find an issue to close but didn't think of looking up PRs 😅. Sorry for the race condition.

yjaaidi:
By the way, about `imports` option, do you have a use case where https://angular.dev/api/core/importProvidersFrom is not enough?

prclm:
@yjaaidi I don't have a usecase. But i didn't wanted to "break" someones code completely. if they actually had ngmodules in the "providers". then they could just rename "providers" to "imports" and it should work as before.

I merged your changes, and modified them just a little. 

Feel free to just close this PR if the "imports" option is not desired.
Otherwise i would be happy to contribute :)

packages/vitest-angular/setup-testbed.ts
```diff
@@ -17,12 +13,14 @@ const ANGULAR_TESTBED_SETUP = Symbol.for('testbed-setup');
 
 type TestBedSetupOptions = {
   zoneless?: boolean;
-  providers?: Provider[];
+  imports?: NgModule['imports'];
+  providers?: NgModule['providers'];
```
brandonroberts: We don't want this tied to an NgModule

```suggestion
  providers?: Provider[]
```

packages/vitest-angular/setup-testbed.ts
```diff
@@ -35,13 +33,14 @@ export function setupTestBed({
     @NgModule({
       providers: [
         ...(zoneless ? [provideZonelessChangeDetection()] : []),
-        providers,
+        ...providers,
```
brandonroberts: We don't need to unwrap this
```suggestion
        providers,
```

packages/vitest-angular/setup-testbed.ts
```diff
@@ -35,13 +33,14 @@ export function setupTestBed({
...
       ],
+      imports: imports || [],
     })
-    class TestModule {}
+    class TestModule extends BrowserTestingModule {}
```
brandonroberts: Let's not extend this
```suggestion
    class TestModule {}
```

packages/vitest-angular/setup-testbed.ts
```diff
@@ -35,13 +33,14 @@ export function setupTestBed({
...
 
     getTestBed().initTestEnvironment(
-      [BrowserTestingModule, TestModule],
+      [TestModule],
```
brandonroberts: Keep as-is

```suggestion
      [BrowserTestingModule, TestModule],
```

brandonroberts (changes requested):

prclm:
@brandonroberts  i updated my changes

packages/vitest-angular/setup-testbed.ts
```diff
@@ -1,7 +1,10 @@
 import {
+  EnvironmentProviders,
+  inject,
```
brandonroberts, line 3:
```suggestion
```

packages/vitest-angular/setup-testbed.ts
```diff
@@ -17,12 +20,14 @@ const ANGULAR_TESTBED_SETUP = Symbol.for('testbed-setup');
 
 type TestBedSetupOptions = {
   zoneless?: boolean;
-  providers?: Provider[];
+  imports?: Type<any>[];
```
brandonroberts, line 23:
```suggestion
  imports?: Type<unknown>[];
```

brandonroberts:
Thanks @prclm. On second conversation with @yjaaidi around testing patterns, we don't think encouraging further usage of NgModules even for convenience is the path we want to follow for this.

Issue #2063, opened 2026-02-11
`@analogjs/vite-plugin-angular` missing dependencies


gtbuchanan:
### Please provide the environment you discovered this bug in.

https://stackblitz.com/edit/github-387ipnd6

### Which area/package is the issue in?

vite-plugin-angular

### Description

`tinyglobby` and `typescript` are [referenced](https://github.com/analogjs/analog/blob/daaba0d5971eba1d3922703e34848d3f71a2e31d/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts#L1-L19) by `@analogjs/vite-plugin-angular`, but are not defined as dependencies in `package.json`. Using `pnpm` with `hoist=false` in `.npmrc` exposes this problem. I would expect `tinyglobby` to be listed in `dependencies` and `typescript` to be listed in `peerDependencies`.

### Please provide the exception or error you saw

```true
Error: Cannot find package 'tinyglobby' imported from my-project\node_modules\.pnpm\@analogjs+vite-plugin-angul_71c0728489db9680a58f98aee9b4eb1a\node_modules\@analogjs\vite-plugin-angular\src\lib\angular-vite-plugin.js
        at Object.getPackageJSONURL (node:internal/modules/package_json_reader:316:9)
        at packageResolve (node:internal/modules/esm/resolve:768:81)
        at moduleResolve (node:internal/modules/esm/resolve:858:18)
        at defaultResolve (node:internal/modules/esm/resolve:990:11)
        at ModuleLoader.#cachedDefaultResolve (node:internal/modules/esm/loader:718:20)
        at ModuleLoader.#resolveAndMaybeBlockOnLoaderThread (node:internal/modules/esm/loader:735:38)
        at ModuleLoader.resolveSync (node:internal/modules/esm/loader:764:52)
        at ModuleLoader.getOrCreateModuleJob (node:internal/modules/esm/loader:617:27)
        at ModuleJobSync.syncLink (node:internal/modules/esm/module_job:143:33)
        at ModuleJobSync.link (node:internal/modules/esm/module_job:463:17)
```

### I would be willing to submit a PR to fix this issue

- [x] Yes
- [ ] No

brandonroberts:
Thanks @gtbuchanan. `tinyglobby` should be added to the `dependencies`, in the `package.json`, but `typescript` should already be included in the projects' package.json either way

gtbuchanan:
[...]

PR #2071, opened 2026-02-17
fix(vitest-angular): add missing zone.js optional peer dependency

gtbuchanan:
## PR Checklist

Related to #2063 and #2069. pnpm's `hoist=false` prevents packages from having implicit dependencies.

## What is the new behavior?

This defines `zone.js` as an optional peer dependency so `@analogjs/vitest-angular` can actually find it. I pulled the version from the root `package.json`.

- [ ] Yes
- [x] No

The peer dependency is optional, so it should not be a breaking change given a non-ancient package manager version is used.

## Other information

I noticed `pnpm-lock.yaml` wasn't updated when I ran `pnpm i` after this change. That was unexpected for me. Let me know if there is something special I need to do, or if this is a symptom of using Nx.

### DIFF
--- a/packages/vitest-angular/package.json
+++ b/packages/vitest-angular/package.json
@@ -36,7 +36,13 @@
     "@analogjs/vite-plugin-angular": "*",
     "@angular-devkit/architect": ">=0.1500.0 < 0.2200.0",
     "@angular-devkit/schematics": ">=17.0.0",
-    "vitest": "^1.3.1 || ^2.0.0 || ^3.0.0 || ^4.0.0"
+    "vitest": "^1.3.1 || ^2.0.0 || ^3.0.0 || ^4.0.0",
+    "zone.js": ">=0.14.0"
+  },
+  "peerDependenciesMeta": {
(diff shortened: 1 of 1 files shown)


packages/vitest-angular/package.json
```diff
@@ -36,7 +36,13 @@
     "@analogjs/vite-plugin-angular": "*",
     "@angular-devkit/architect": ">=0.1500.0 < 0.2200.0",
     "@angular-devkit/schematics": ">=17.0.0",
-    "vitest": "^1.3.1 || ^2.0.0 || ^3.0.0 || ^4.0.0"
+    "vitest": "^1.3.1 || ^2.0.0 || ^3.0.0 || ^4.0.0",
+    "zone.js": "^0.15.0"
```
brandonroberts: Let's expand the zone.js version to include older versions of Angular.

```suggestion
    "zone.js": ">=0.14.0"
```

gtbuchanan: Done

PR #2086, opened 2026-02-24
fix(storybook-angular): add missing applyDecorators to render annotaions

### LINKED ISSUE #2074: [storybook] componentWrapperDecorator broken in vitest

https://github.com/brandonroberts/angular-vite-storybook/compare/main...prclm:angular-vite-storybook:decorators-bug

I reproduced the bug in the fork of brandonroberts/angular-vite-storybook.

vitest-angular

### Description

I don't know where the bug is actually happening. Could also be in the "@storybook/addon-vitest" package.

When rendering a story in vitest, that is wrapped with a `componentWrapperDecorator`, the decorator function does not receive the story, but "undefined". And then it renders the decorator with the text "undefined" inside, not the story.

```true

```

### Other information

_No response_

- [ ] Yes
- [ ] No

Bayathy:
## PR Checklist

Closes #2074 

## What is the new behavior?

  `@storybook/angular`'s [`componentWrapperDecorator()` depends on `storyFn().template` internally](https://github.com/storybookjs/storybook/blob/10d1e88911be83d5467d32ccd446d5499d00b8e5/code/frameworks/angular/src/client/decorators.ts#L60).

  This `storyFn().template` is resolved by `prepareMain()`, which is called inside Storybook's  [`decorateStory()`](https://github.com/storybookjs/storybook/blob/10d1e88911be83d5467d32ccd446d5499d00b8e5/code/frameworks/angular/src/client/decorateStory.ts#L7). This function is [exported as `applyDecorators` within Storybook](https://github.com/storybookjs/storybook/blob/10d1e88911be83d5467d32ccd446d5499d00b8e5/code/frameworks/angular/src/client/config.ts#L4).

  However, before this fix, `decorateStory()` was not being called at runtime. Instead, Storybook's generic renderer
  [`defaultDecorateStory`](https://github.com/storybookjs/storybook/blob/10d1e88911be83d5467d32ccd446d5499d00b8e5/code/core/src/preview-api/modules/store/decorators.ts#L49) was being used. Since  `defaultDecorateStory` does not call `prepareMain()`, `storyFn().template` remained unresolved.

  This happens because [`prepareStory()`](https://github.com/storybookjs/storybook/blob/10d1e88911be83d5467d32ccd446d5499d00b8e5/code/core/src/preview-api/modules/store/csf/prepareStory.ts#L37C17-L37C29), which
   is called during portable story rendering in the Vitest environment, contains the following:

```ts
  // Currently it is only possible to set these globally
  const { applyDecorators = defaultDecorateStory, runStep } = projectAnnotations;
```

  When applyDecorators is not defined in `projectAnnotations`, it falls back to `defaultDecorateStory`.

  With this fix, `applyDecorators `is now included in `renderAnnotations` and passed through to `prepareStory()`. This causes Angular's `decorateStory()` to be used, which calls `prepareMain()` and properly resolves `storyFn().template`.

  As a result, `storyFn().template` inside `componentWrapperDecorator() `is no longer undefined.

- [ ] Yes
- [x] No

## Other information

This is my first PR to an OSS project. English is not my first language, so I wrote this description in Japanese first and then had it translated. I apologize if anything is unclear or hard to read.

### DIFF
--- a/packages/storybook-angular/src/lib/testing.ts
+++ b/packages/storybook-angular/src/lib/testing.ts
@@ -21,6 +21,7 @@ export async function renderToCanvas(
 const renderAnnotations = {
   render,
   renderToCanvas,
+  applyDecorators: configAnnotations.applyDecorators,
 };
 
 export function setProjectAnnotations(


Bayathy:
The reproduction environment is here.

https://github.com/Bayathy/analog-storybook

When running `npm run test-storybook` in this repository, the following result is obtained:

```shell
 FAIL   storybook (chromium)  src/stories/page.stories.ts > Logged In
TestingLibraryElementError:
Click to debug the error directly in Storybook: http://localhost:6006/?path=/story/example-page--logged-in&addonPanel=storybook/interactions/panel

Unable to find an accessible element with the role "button" and name `/Log in/i`

There are no accessible roles. But there might be some inaccessible roles. If you wish to access them, then set the `hidden` option to `true`. Learn more about this here: https://testing-library.com/docs/dom-testing-library/api-queries#byrole

Ignored nodes: comments, script, style
<div
  data-sb-story-uid="example-page--logged-in-0"
  id="example-page--logged-in"
>
  <example-page--logged-in
    example-page--logged-in-0=""
    ng-version="21.1.5"
  >
    <div
      style="padding: 1rem;"
    >
      undefined
    </div>
  </example-page--logged-in>
</div>
 ❯ getByRole src/stories/page.stories.ts:24:31
     22|   play: async ({ canvasElement }) => {
     23|     const canvas = within(canvasElement);
     24|     const loginButton = canvas.getByRole('button', { name: /Log in/i });
       |                               ^
     25|     await expect(loginButton).toBeInTheDocument();
     26|     await userEvent.click(loginButton);
 ❯ node_modules/.cache/storybook/10.2.11/95209e453a51f1cc97c9c8cfcc07cc086d2a83ebdbfe04ef274ba9013d3e81d4/sb-vitest/deps/@storybook_addon-vitest_internal_test-utils.js?v=8217aa13:126:184
 ```
 
 
The story rendered inside componentWrapperDecorator is becoming undefined.

**Review by brandonroberts — APPROVED** [review]

**Comment by brandonroberts** [comment]
Thanks @Bayathy!

**Comment by prclm** [comment]
@Bayathy @brandonroberts I installed 2.3.0-beta.17 and it breaks my storybook interaction test in vitest where i use output-binding. (The test works inside storybook ui and with  2.3.0-beta.16). 

It looks to me that the render in vitest does not add the component bindings correctly.


```typescript
import { Meta, StoryObj } from '@storybook/angular';
import { expect, fn } from 'storybook/test';

import { DemoComponent } from './demo.component';

const meta = {
  component: DemoComponent,
  args: { btnClicked: fn() },
} satisfies Meta<DemoComponent>;

export default meta;

type Story = StoryObj<DemoComponent>;

export const Default: Story = {
  play: async ({ canvas, userEvent, args }) => {
    await userEvent.click(canvas.getByTestId('demo-btn'));
    // the mock fn() does not get called, it is not bound correctly
    void expect(args.btnClicked).toBeCalled();
  },
};

```

prclm:
here is what i get when i log the story in a decorator:

```typescript
componentWrapperDecorator((story) => {
  console.log({ story });
  return `
    <div id="wrapper">
      ${story}
    </div>`;
});
```

2.3.0-beta.17
```
stdout | src/lib/demo.stories.ts > Default
{
  "story": "<app-demo    />",
}
```

2.3.0-beta.16
```
stdout | src/lib/demo.stories.ts > Default
{
  "story": "<app-demo (btnClicked)="btnClicked($event)"  />",
}
```

prclm:
at the moment i work around the decorator problem with this helper to test, if the story runs inside vitest. Then i don't apply my decorators at all...

```ts
/**
 * Helper method to test if in vitest environment.
 * @returns If is vitest environment.
 */
export function isVitestBrowser(): boolean {
  return window.__vitest_browser__ ?? false;
}
```

But when this patch will land in the final release, i would not know how to get my tests fixed.
For me it is more imporant to have component bindings of my components work in my tests, then having the decorators rendered.

@brandonroberts  I would prefer to revert this change, if it can not be fixed...

brandonroberts:
Yes, we will revert this and revisit other options. We at least have a reproduction now

Bayathy:
@prclm 
Please add the following code to the story file that was previously shared as a reference.

This issue might be related:
https://github.com/storybookjs/storybook/issues/20309

```diff
import type { Meta, StoryObj } from '@storybook/angular';
import { expect, fn } from 'storybook/test';

import { DemoComponent } from './demo.component';

const meta: Meta<DemoComponent> = {
  title: 'Example/Demo',
  component: DemoComponent,
+ argTypes: { btnClicked: { action: 'btnClicked' }, },
  args: { btnClicked: fn() },
};

export default meta;
type Story = StoryObj<DemoComponent>;

export const Default: Story = {
  play: async ({ canvas, userEvent, args }) => {
    await userEvent.click(canvas.getByTestId('demo-btn'));
    void expect(args.btnClicked).toBeCalled();
  },
};
```

With this modification, the story output in my environment looked like this:
```ts
<app-demo (btnClicked)="btnClicked($event)"></app-demo>
```

prclm:
> @prclm Please add the following code to the story file that was previously shared as a reference.
> 
> This issue might be related: [storybookjs/storybook#20309](https://github.com/storybookjs/storybook/issues/20309)
> 
> ```diff
> import type { Meta, StoryObj } from '@storybook/angular';
> import { expect, fn } from 'storybook/test';
> 
> import { DemoComponent } from './demo.component';
> 
> const meta: Meta<DemoComponent> = {
>   title: 'Example/Demo',
>   component: DemoComponent,
> + argTypes: { btnClicked: { action: 'btnClicked' }, },
>   args: { btnClicked: fn() },
> };
> 
> export default meta;
> type Story = StoryObj<DemoComponent>;
> 
> export const Default: Story = {
>   play: async ({ canvas, userEvent, args }) => {
>     await userEvent.click(canvas.getByTestId('demo-btn'));
>     void expect(args.btnClicked).toBeCalled();
>   },
> };
> ```
> 
> With this modification, the story output in my environment looked like this:
> 
> ```ts
> <app-demo (btnClicked)="btnClicked($event)"></app-demo>
> ```

Thanks @Bayathy, this indeed fixes the tests. Do you think it is possible to inject a console.log somehow to give the developer a hint what is going on?

@brandonroberts is it possible to bring this MR changes back in the next version and maybe make a "troubleshooting" note somewhere if people run in this bug, that they need to add the actions as "argTypes"?

brandonroberts:
@prclm Sure, we can add it to the bottom of the guide

PR #2090, opened 2026-02-26
fix(vite-plugin-angular): hash styleId to prevent filename exceeding max length

### LINKED ISSUE #2077: [@analogjs/vite-plugin-angular] `ENAMETOOLONG` when preprocessing inline CSS

Hello. I have a project which uses `vite-plugin-angular` to run vitest. It also uses Tailwindcss which uses Postcss. Altogether, when these plugins combine with the Angular compiler, they occasionally create this bug where the content of the inline component css is encoded to base64 (by `@angular/build`) and then used as a temporary/virtual file name (by `@analogjs/vite-plugin-angular`) for vitest to reference (in postcss). This triggers an exception where `fs.statSync` from Postcss throws `ENAMETOOLONG` because the encoded content exceeds the linux/macos file name limit of 255 bytes. 

vite-plugin-angular

### Description

I don't think this particular bug is specific to any one of these projects but rather a confluence of all of them together. It is, however, solvable from this plugin's source, which is how I've fixed it with a patch in my project's repo. 

In `angular-jit-plugin.ts` of `vite-plugin-angular` you use `styleId` directly from `angular-build` which can occasionally exceed 255 bytes and is used directly by postcss for the temporary filename. My proposal is to instead hash this string with `sha256` and use the 16-byte hash as the `styleId` instead. It's a 3-line fix:

```
import { createHash } from 'node:crypto';
line 25: const hash = createHash('sha256').update(styleId).digest('hex').slice(0, 16);
line 36: replace ${styleId} with ${hash}
```

If you're worried about it, you could even put the fix behind a configuration flag to allow the previous behavior to send the raw base64 content as the styleId instead of the sha256 content.

Thanks

```true
Error: ENAMETOOLONG: name too long, stat 'CiAgI[...truncated...]A==.css'
    at Object.statSync (node:fs:1739:25)
    at Object.Once (node_modules/@tailwindcss/postcss/dist/index.js:10:4612)
    at LazyResult.runAsync (node_modules/postcss/lib/lazy-result.js:293:11)
    at runPostCSS (node_modules/vite/dist/node/chunks/config.js:30145:19)
    at compilePostCSS (node_modules/vite/dist/node/chunks/config.js:30129:6)
    at compileCSS (node_modules/vite/dist/node/chunks/config.js:30059:26)
    at preprocessCSS (node_modules/vite/dist/node/chunks/config.js:30221:9)
    at LoadPluginContext.load (node_modules/@analogjs/vite-plugin-angular/src/lib/angular-jit-plugin.js:21:38)
    at EnvironmentPluginContainer.load (node_modules/vite/dist/node/chunks/config.js:28759:19)
    at loadAndTransform (node_modules/vite/dist/node/chunks/config.js:22628:21) {
  errno: -63,
  code: 'ENAMETOOLONG',
  syscall: 'stat',
  path: 'CiAgICAgI[...truncated...]A==.css'
}
CssSyntaxError: [postcss] tailwindcss: /workspace/project/CiAgI[...truncated...]A==.css:1:1: ENAMETOOLONG: name too long, stat 'CiAgI[...truncated...]A==.css'

> 1 | 
    | ^
  2 |       @reference 'tailwindcss';
  3 |       :host {
```

### Other information

_No response_

- [x] Yes
- [ ] No

castral:
`styleId` from `@angular/build` was being used directly as a temp filename but it could potentially exceed 255 bytes of base64-encoded data causing third party tooling to throw `ENAMETOOLONG` on `fs.statSync` calls.

Closes #2077

## What is the new behavior?

`styleId` is hashed in `SHA256` to limit the temporary filename length to 16 bytes.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/vite-plugin-angular/src/lib/angular-jit-plugin.ts
+++ b/packages/vite-plugin-angular/src/lib/angular-jit-plugin.ts
@@ -1,3 +1,4 @@
+import { createHash } from 'node:crypto';
 import { Plugin, ResolvedConfig, preprocessCSS } from 'vite';
 
 export function jitPlugin({
@@ -22,6 +23,11 @@ export function jitPlugin({
     async load(id: string) {
       if (id.includes('virtual:angular:jit:style:inline;')) {
         const styleId = id.split('style:inline;')[1];
+        // styleId may exceed 255 bytes of base64-encoded content, limit to 16
+        const styleIdHash = createHash('sha256')
+          .update(styleId)
+          .digest('hex')
+          .slice(0, 16);
 
         const decodedStyles = Buffer.from(
           decodeURIComponent(styleId),
@@ -33,7 +39,7 @@ export function jitPlugin({
         try {
           const compiled = await preprocessCSS(
             decodedStyles,
-            `${styleId}.${inlineStylesExtension}?direct`,
+            `${styleIdHash}.${inlineStylesExtension}?direct`,
             config,
           );
           styles = compiled?.code;


brandonroberts:
Thanks @castral!

PR #2091, opened 2026-02-27
feat(content): extract TOC to be property on contentFile

mhartington:
## PR Checklist

Closes #

## What is the new behavior?
This extracts the content headers and appends them to the object returned by content-file. 

- [ ] Yes
- [ ] No
- [X] Maybe?

This does change what gets returned by `inejctContent` and `contentFileResource`, but it doesn't remove the ability to manually call `getContentHeadings`


brandonroberts (approved):
Clean! 🧹

PR #2095, opened 2026-03-01
fix(vite-plugin-angular): use single ? for JIT style resolve suffix

### LINKED ISSUE #2094: fix(vite-plugin-angular): JIT style resolveId uses ?? (double question mark) instead of ? — SCSS not preprocessed
## Bug Description

The `resolveId` hook in `angular-vite-plugin.ts` (line 507) uses `??` (double question mark) for the style/template URL suffix:

```ts
return `${normalizePath(
  resolve(dirname(importer as string), path),
)}??${id.includes(':style') ? 'inline' : 'raw'}`;
```

This produces URLs like `path??inline` instead of `path?inline`.

## Why This Is Wrong

- `?inline` → URL search params: `{ inline: '' }` ✅ (Vite recognizes this as an inline CSS import)
- `??inline` → URL search params: `{ '?inline': '' }` ❌ (Vite does NOT recognize this)

```js
new URL("file:///path/to/file?inline").searchParams  // { inline: "" } ✅
new URL("file:///path/to/file??inline").searchParams  // { "?inline": "" } ❌
```

## Impact

With `??inline`, Vite's CSS pipeline does not process the file, so SCSS/LESS files are loaded as raw text without preprocessing. This causes **production builds** (e.g., Storybook) to contain raw SCSS (e.g., `&.class-name` nesting) instead of compiled CSS.

## Reproduction

1. Create an Angular component with an external `.scss` styleUrl
2. Build with `@analogjs/vite-plugin-angular` in JIT mode (`jit: true`)
3. Inspect the production bundle — SCSS is not compiled to CSS

## Expected Behavior

Style files should be resolved with `?inline` (single `?`) so Vite runs them through the CSS preprocessor pipeline (SCSS → CSS).

## Fix

Change `??` to `?` in the `resolveId` hook:

```diff
- )}??${id.includes(':style') ? 'inline' : 'raw'}`;
+ )}?${id.includes(':style') ? 'inline' : 'raw'}`;
```

This is a regression from commit bcb6da9f (PR #2024), which correctly added the style/template differentiation but used `??` instead of `?`.

## Workaround

Add an `enforce: 'pre'` Vite plugin that intercepts `angular:jit:style:file;` IDs and resolves them with `?inline`:

```ts
{
  name: 'angular-jit-scss-inline-resolve',
  enforce: 'pre',
  resolveId(id, importer) {
    if (id.startsWith('angular:jit:style:file;') && importer) {
      const filePath = id.split(';')[1];
      return normalizePath(resolve(dirname(importer), filePath)) + '?inline';
    }
  }
}
```

## Versions

- `@analogjs/vite-plugin-angular`: 2.2.1 (and all versions since PR #2024)
- Vite: 7.x
- Angular: 20.x

AlonMiz:
## Summary

Fixes #2094

The `resolveId` hook uses `??` (double question mark) for the style/template URL suffix, producing URLs like `path??inline`. Vite does not recognize `??inline` as an inline CSS import — it only recognizes `?inline` (single `?`).

This causes SCSS/LESS files to be loaded as raw text without preprocessing in production builds.

## The Fix

```diff
- )}??${id.includes(':style') ? 'inline' : 'raw'}`;
+ )}?${id.includes(':style') ? 'inline' : 'raw'}`;
```

One character change: `??` → `?`.

## Verification

```js
new URL("file:///path/to/file?inline").searchParams   // { inline: "" } ✅
new URL("file:///path/to/file??inline").searchParams   // { "?inline": "" } ❌
```

## Context

This is a regression from bcb6da9f (PR #2024), which correctly added the style vs template differentiation but accidentally used `??` instead of `?`.

### DIFF
--- a/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
+++ b/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
@@ -504,7 +504,7 @@ export function angular(options?: PluginOptions): Plugin[] {
           const path = id.split(';')[1];
           return `${normalizePath(
             resolve(dirname(importer as string), path),
-          )}??${id.includes(':style') ? 'inline' : 'raw'}`;
+          )}?${id.includes(':style') ? 'inline' : 'raw'}`;
         }
 
         // Map angular external styleUrls to the source file


AlonMiz:
Closing to recreate with a clean branch based on upstream main.

Issue #2040, opened 2026-01-14
Update to nitropack 2.13


Martinspire:
### Please provide the environment you discovered this bug in.

There's a few packages with npm audit warnings related to analogjs and upgrading nitropack to 2.13 would get rid of the glob audit issue.

Unfortunately listhen still hasn't been updated yet ([the PR has been open since november](https://github.com/unjs/listhen/pull/212)), so its not possible yet to fix it. Though you could alwasy have overrides present if you really wanted to fix them.

### Which area/package is the issue in?

platform

### Description

Packages to update:
@analogjs/vite-plugin-nitro
@analogjs/platform

### Please provide the exception or error you saw

```true
┌─────────────────────┬────────────────────────────────────────────────────────┐
│ high                │ glob CLI: Command injection via -c/--cmd executes      │
│                     │ matches with shell:true                                │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ Package             │ glob                                                   │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ Vulnerable versions │ >=10.2.0 <10.5.0                                       │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ Patched versions    │ >=10.5.0                                               │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ Paths               │ .>@analogjs/platform>nitropack>@vercel/nft>glob        │
├─────────────────────┼────────────────────────────────────────────────────────┤
│ More info           │ https://github.com/advisories/GHSA-5j98-mcp5-4vw2      │
└─────────────────────┴────────────────────────────────────────────────────────┘
```

### Other information

@vercel/nft went from 0.30 to 1.2.0

### I would be willing to submit a PR to fix this issue

- [ ] Yes
- [ ] No

PR #2106, opened 2026-03-10
fix: update dependencies and package manager versions across projects

benpsnyder:
## PR Checklist

Updates the repo's package manager and dependency versions across the workspace, while making the small config changes needed to keep the example apps and docs app compatible with those upgrades.

## What is the new behavior?

- Bumps the workspace to `pnpm@10.32.0` and refreshes a broad set of root dependencies and lockfile entries.
- Upgrades the docs app to newer Docusaurus packages and adds explicit sidebar `key` values for repeated `Overview` items so the sidebar remains stable after the upgrade.
- Updates the Vite configs in the example apps to use `PluginOption` for `rollup-plugin-visualizer`, which matches the newer Vite typings.
- Aligns `libs/my-package` with the current Angular/`tslib` versions used in the repo.
- Switches the `SatoriOptions` import to `satori`, matching the newer package export.

- [ ] Yes
- [x] No

## Other information

This is primarily a maintenance/update PR. There should be no intended user-facing behavior changes beyond keeping the workspace compatible with the newer dependency set.

"ship the dependency upgrades" energy

### DIFF
--- a/packages/platform/package.json
+++ b/packages/platform/package.json
@@ -28,21 +28,21 @@
     "url": "https://github.com/sponsors/brandonroberts"
   },
   "dependencies": {
-    "nitropack": "^2.11.0",
+    "nitropack": "^2.13.1",
     "@analogjs/vite-plugin-angular": "^2.4.0-beta.4",
     "@analogjs/vite-plugin-nitro": "^2.4.0-beta.4",
-    "vitefu": "^1.0.0"
+    "vitefu": "^1.1.2"
   },
   "peerDependencies": {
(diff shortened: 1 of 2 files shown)


benpsnyder:
Would resolve https://github.com/analogjs/analog/issues/2040

brandonroberts:
To resolve #2040 the package.json for the vite-plugin-nitro and platform packages needs to be updated also to `^2.13.0`

benpsnyder:
> To resolve #2040 the package.json for the vite-plugin-nitro and platform packages needs to be updated also to `^2.13.0`

done

PR #2109, opened 2026-03-11
fix: concurrent compilation of ssr and client caused corrupted bundles

### LINKED ISSUE #2103: Bug: Race condition in `performCompilation`

I created a file `repro.js` that can be run with `node repro.js --delay-both` it:
- patches `node_modules/@analogjs/vite-plugin-angular/src/lib/angular-vite-plugin.js` to add an artificial delay in some parts of the compilation to trigger the race condition in a deterministic way.
- runs `npm run build`
- restores the patched plugin

https://stackblitz.com/edit/github-aqa3usxg?file=repro.js&preset=node

```bash
# Mode 1: Delay first call after analyzeAsync → client corrupted
node repro.js --with-delay

# Mode 2: Delay second call after analyzeAsync → client corrupted
node repro.js --delay-second

# Mode 3: Delay both calls → BOTH bundles corrupted
node repro.js --delay-both
```

vite-plugin-angular

### Description

I have an AnalogJS app that sometimes fails at runtime in a totally non-deterministic way: without changing anything in the codebase, It returns an error `JIT compiler unavailable` on all calls. It happens approximately 1 out of 5 times. When it fails, the AOT transformation fails so there are  still `templateUrl` and `__decorate` calls in the output files.

It was really difficult to narrow the issue in a minimal app so I chose to patch node modules to add artificial delays. My real app uses @defer, server only dependency and dynamically loaded client dependencies.

## In `vite-plugin-nitro.js`

When `@analogjs/vite-plugin-nitro` builds client and SSR environments concurrently via `Promise.all` with `sharedPlugins: true`, the Angular plugin's `performCompilation()` function corrupts the build output. The AOT transformers silently produce no output, leaving raw `templateUrl`/`styleUrl` strings and `__decorate()` calls in the bundle instead of AOT-compiled component definitions.

This is a **data race on shared mutable closure state** — not a timing-dependent flake. The race window is deterministic once two `buildStart` hooks overlap.

## In `angular-vite-plugin.js`

The following variables are **closure-scoped** (shared across all calls) and mutated without synchronization:

| Variable | Line | Mutated at |
|---|---|---|
| `builder` | ~95 | L748, L752, L758 |
| `nextProgram` | ~95 | L749 |
| `outputFile` | L95 | L842 |
| `outputFiles` | L96 | L780 |

### The race sequence

1. **Call 1** (client) enters `performCompilation`, creates `angularProgram_1`, sets shared `builder = B1` and `nextProgram = P1` (L744-749), then yields at `await angularCompiler.analyzeAsync()` (L761).

2. **Call 2** (SSR) enters `performCompilation`, creates `angularProgram_2`, **overwrites** `builder = B2` and `nextProgram = P2` (same lines), then yields at `analyzeAsync()`.

3. Whichever call finishes `analyzeAsync()` first proceeds to:
   - Get transformers from **its own** `angularCompiler.prepareEmit().transformers` (L769) — these transformers hold references to its own `NgtscProgram`'s `TraitCompiler` internal `Map<ts.Declaration, ...>`
   - Create `writeOutputFile` (L789) — this closure captures the **shared** `builder` variable (now `B2`, set by Call 2)
   - Set `outputFile = writeOutputFile` (L842)

4. The other call finishes and **overwrites** `outputFile` with its own `writeOutputFile`.

5. When Vite later calls `fileEmitter(file)` (L97-99), it always reads the **current** `outputFile` — whichever was set last.

```true
`Error: JIT compiler unavailable` when AOT transformation fails because of a race condition between
```

### Other information

If you care about what Claude think of this issue, here are some proposals:

<details>

## Proposed fix

The root cause is shared mutable state across concurrent `performCompilation()` calls. There are several approaches, roughly ordered by invasiveness:

### Option A: Serialize the builds (minimal change)

In `vite-plugin-nitro.js`, replace `Promise.all` with sequential builds:

```js
buildApp: async (builder) => {
    await builder.build(builder.environments['client']);
    if (options?.ssr || nitroConfig.prerender?.routes?.length) {
        await builder.build(builder.environments['ssr']);
    }
},
```

**Pros:** One-line fix, eliminates the race entirely.
**Cons:** Loses parallelism — build time increases (though for most projects the difference is small since both builds are CPU-bound on the same thread).

### Option B: Use `sharedPlugins: false`

```js
builder: {
    sharedPlugins: false,
    // ...
}
```

Each environment gets its own plugin instance with independent closure state.

**Pros:** Simple configuration change. Each build gets isolated state.
**Cons:** May have other implications (duplicated plugin initialization, different plugin ordering assumptions).

### Option C: Scope mutable state per environment (robust fix)

Make `builder`, `nextProgram`, `outputFile`, and `outputFiles` keyed per environment (e.g., using a `Map<string, ...>` keyed by environment name). The `performCompilation` function already receives `config` which contains the environment info.

```js
// Instead of:
let builder;
let outputFile;

// Use:
const builderPerEnv = new Map();
const outputFilePerEnv = new Map();
```

**Pros:** Preserves parallelism, correct by construction, no external changes needed.
**Cons:** More invasive refactor of the closure state.

### Option D: Add a mutex/semaphore around `performCompilation`

Use an async lock so only one `performCompilation` runs at a time:

```js
let compilationLock = Promise.resolve();

async function performCompilation(config, ids) {
    const release = await acquireLock();
    try {
        // ... existing code ...
    } finally {
        release();
    }
}
```

**Pros:** Minimal changes to existing logic.
**Cons:** Serializes compilation (same perf impact as Option A), but at a finer granularity.

---

**Recommendation:** Option C (per-environment state) is the most correct fix. Option A (sequential builds) is the simplest and safest short-term mitigation.
</details>

In order to "fix" the issue, I use [patch-package](https://npmx.dev/package/patch-package) with the following patch:

```diff
diff --git a/node_modules/@analogjs/vite-plugin-angular/src/lib/angular-vite-plugin.js b/node_modules/@analogjs/vite-plugin-angular/src/lib/angular-vite-plugin.js
index 11cdb58..f03e45e 100644
--- a/node_modules/@analogjs/vite-plugin-angular/src/lib/angular-vite-plugin.js
+++ b/node_modules/@analogjs/vite-plugin-angular/src/lib/angular-vite-plugin.js
@@ -103,6 +103,11 @@ export function angular(options) {
     const fileTransformMap = new Map();
     let styleTransform;
     let pendingCompilation;
+    // Serialization mutex: ensures concurrent buildStart calls (from sharedPlugins
+    // Environment API builds running client + ssr via Promise.all) never execute
+    // performCompilation simultaneously, which would corrupt shared mutable state
+    // (builder, angularCompiler, nextProgram, transformers, outputFile, cachedHost).
+    let compilationLock = Promise.resolve();
     function angularPlugin() {
         let isProd = false;
         if (angularFullVersion < 190000 || isTest) {
@@ -618,6 +623,21 @@ export function angular(options) {
         compilation.close?.();
     }
     async function performCompilation(config, ids) {
+        // Serialize all compilation calls to prevent concurrent mutation of shared
+        // state (builder, angularCompiler, nextProgram, transformers, outputFile).
+        // This is critical when sharedPlugins:true causes buildStart to be called
+        // concurrently for the client and ssr environments via Promise.all.
+        let resolve;
+        const previousLock = compilationLock;
+        compilationLock = new Promise((r) => { resolve = r; });
+        try {
+            await previousLock;
+            await _doCompilation(config, ids);
+        } finally {
+            resolve();
+        }
+    }
+    async function _doCompilation(config, ids) {
         if (pluginOptions.useAngularCompilationAPI) {
             await performAngularCompilation(config);
             return;

```

- [x] Yes
- [ ] No

GuillaumeNury:
## PR Checklist

Closes #2103

- [ ] Yes
- [x] No

### DIFF
--- a/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
+++ b/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
@@ -200,6 +200,7 @@ export function angular(options?: PluginOptions): Plugin[] {
     filename: string,
   ) => Promise<vite.PreprocessCSSResult>;
   let pendingCompilation: Promise<void> | null;
+  let compilationLock = Promise.resolve();
 
   function angularPlugin(): Plugin {
     let isProd = false;
@@ -895,6 +896,24 @@ export function angular(options?: PluginOptions): Plugin[] {
   }
 
   async function performCompilation(config: ResolvedConfig, ids?: string[]) {
+    let resolve: (() => unknown) | undefined;
+    const previousLock = compilationLock;
+    compilationLock = new Promise<void>((r) => {
+      resolve = r;
+    });
+    try {
+      await previousLock;
+      await _doPerformCompilation(config, ids);
+    } finally {
+      resolve!();
+    }
+  }
+
+  /**
+   * This method share mutable state and performs the actual compilation work.
+   * It should not be called concurrently. Use `performCompilation` which wraps this method in a lock to ensure only one compilation runs at a time.
+   */
+  async function _doPerformCompilation(config: ResolvedConfig, ids?: string[]) {
     if (pluginOptions.useAngularCompilationAPI) {
       await performAngularCompilation(config);
       return;


brandonroberts:
Thanks @GuillaumeNury!

PR #2113, opened 2026-03-14
refactor: migrate to Nitro v3 beta and align with h3 v2 APIs

benpsnyder:
## PR Checklist

This PR updates Analog's Nitro integration from the earlier Nitro v3 alpha series to the current beta line and aligns the codebase with Nitro/h3 v2 APIs. It also fixes the `@analogjs/trpc` Nitro handler so it works with the newer h3 event model, where writing directly to `event.node.res` is no longer reliable in all runtimes.

## What is the new behavior?

Analog now targets Nitro `3.0.260311-beta` and consistently imports server utilities from `nitro/h3` instead of `h3`.

The Nitro integration has also been adjusted for the newer runtime and bundler behavior:
- `@analogjs/vite-plugin-nitro` now lets Nitro choose its default builder again.
- Nitro bundler config is sanitized to avoid alpha-era Rollup/Rolldown issues around `codeSplitting`, `manualChunks`, and route-param-derived chunk names.
- Vercel handling is simplified to Nitro v3's `vercel` preset model, and Cloudflare preset detection now accepts both hyphenated and underscored forms.
- `@analogjs/trpc` now returns a web `Response` instead of mutating `event.node.res`, while still preserving Node request access for tRPC request resolution and error handling.
- Examples, templates, and tests have been updated to reflect the new `nitro/h3` import path and the newer Nitro compatibility date.

- [x] Yes
- [ ] No

Existing applications or integrations that import server helpers from `h3` should migrate those imports to `nitro/h3`.

If consumers are still using the old Nitro Vercel preset naming, they should also move from `vercel-edge` to `vercel` in line with Nitro v3 behavior.

### DIFF
--- a/apps/analog-app-e2e-cypress/project.json
+++ b/apps/analog-app-e2e-cypress/project.json
@@ -19,7 +19,7 @@
     "e2e": {
       "executor": "nx:run-commands",
       "options": {
-        "command": "WAIT_ON_TIMEOUT=30000 start-server-and-test --expect 200 'nx run analog-app:serve-nitro' http://localhost:3000 'nx run analog-app-e2e-cypress:cypress'"
+        "command": "start-server-and-test --expect 200 'nx run analog-app:serve-nitro' http://localhost:3000 'nx run analog-app-e2e-cypress:cypress'"
       },
       "configurations": {
         "watch": {


benpsnyder:
> Not a fan of the patching of internal bundlers and general inlining of plugins

Thank you for putting this through. I will explore how to address your feedback on a follow-up.

PR #2132, opened 2026-03-16
build: pnpm workspace correctness, cross-platform builds, and Dagger CI

benpsnyder:
## PR Checklist

Closes #

## Affected scope

- Primary scope: `platform`
- Secondary scopes: `astro-angular`, `content`, `content-plugin`, `create-analog`, `nx-plugin`, `router`, `storybook-angular`, `vite-plugin-angular`, `vite-plugin-nitro`, `vitest-angular`

- [ ] Squash merge
- [x] Rebase merge
- [ ] Other

This PR spans multiple packages because the build/publish pipeline, workspace layout, and CI validation all cross package boundaries.

## What is the new behavior?

- pnpm now treats the full monorepo as a real workspace, which removes phantom dependency resolution and makes lockfile/install behavior reproducible.
- Library packages that previously built through `@nx/js:tsc` now build through Vite/rolldown with preserved module structure plus declaration-only TypeScript output.
- Cross-platform Vite build issues that only surfaced on Windows are fixed by handling absolute paths correctly, normalizing `preserveModulesRoot`, and using explicit object `lib.entry` mappings.
- Build and publish helper scripts were updated for cross-platform correctness, including Windows-safe process execution and shared artifact verification.
- Release-package builds now validate their generated artifact layout, including the maintainer-reported missing outputs for `astro-angular`, `platform`, `storybook-angular`, and `vite-plugin-angular`.
- Linux CI now runs through Dagger for reproducible build/test/e2e execution, while Windows and macOS continue validating native platform behavior.
- Build targets that write to `node_modules/` now avoid unsafe Nx cache hits that could skip restoring required outputs.
- The deprecated `@analogjs/trpc` package and demo app were removed.

## Test plan

- [x] `nx format:check`
- [x] `pnpm build`
- [x] `pnpm test`
- [x] Manual verification

Additional verification:

- Verified native CI/build behavior on Windows and macOS alongside Linux Dagger CI.
- Verified generated package artifacts for:
  - `@analogjs/astro-angular` exported `src/*` runtime files
  - `@analogjs/platform` nested `src/lib/nx-plugin` executors/generators output
  - `@analogjs/storybook-angular` `src/lib/*.js` builder/testing output
  - `@analogjs/vite-plugin-angular` `src/lib/tools` builder output
- Verified the new artifact validation script runs as part of the affected release-package build targets.

- [ ] Yes
- [x] No

## Other information

- This PR intentionally touches multiple packages because the build system, package output layout, workspace correctness, and CI behavior are tightly coupled.
- Maintainer feedback about missing build artifacts has been addressed and covered with regression validation in the build pipeline.


benpsnyder:
## AI Critical Review: `feat/pnpm-correctness` vs `analogjs/alpha`

**Stats**: 2 commits, 98 files changed

Review covers four main areas: bugs, missing dependency edges, behavioral changes, and Dagger CI issues.

## 🐛 Bugs / Errors to Fix (Must-fix before merge)

- [x] ~~`build:test` script in `package.json` references non-existent `ng-app` project~~
  → Fixed: changed to `blog-app` (a real project that validates the build pipeline)

- [x] ~~`@parcel/watcher` is **duplicated** in `pnpm.onlyBuiltDependencies`~~
  → Fixed: removed duplicate entry

- [x] ~~**Dagger lint step is fundamentally incomplete**~~
  → Fixed: `lint()` now runs `pnpm run lint` (all 3 passes: oxlint, type-aware oxlint, ESLint via Nx) instead of only the legacy ESLint path

- [x] ~~Node version mismatch (`.node-version` = `24.14.0`, Dagger hardcoded `24.11.0`)~~
  → Fixed: updated Dagger `base()` image to `node:24.14.0`

## ⚠️ Missing Dependency Graph Edges (Risk of broken isolated runs)

Implicit dependencies were removed without `dependsOn` replacements. Global `test`/`e2e` default `dependsOn: ["^build"]` was also dropped.

- [x] ~~`analog-preset-e2e` — lost `implicitDependencies: ["platform"]`~~
  → Fixed: added `dependsOn: ["platform:build"]` to e2e target

- [x] ~~`create-analog-e2e` — lost `implicitDependencies: ["create-analog"]`~~
  → Fixed: added `dependsOn: ["create-analog:build"]` to e2e target

- [x] ~~`nx-plugin-e2e` — lost `implicitDependencies: ["platform", "trpc", "content"]`~~
  → Fixed: added `dependsOn: ["platform:build", "trpc:build", "content:build"]` to e2e target

- [x] ~~`astro-app` — lost `implicitDependencies: ["astro-angular"]`~~
  → Fixed: added `dependsOn: ["astro-angular:build"]` to build and serve targets

- [ ] `analog-app` → storybook — no `dependsOn` for `storybook-angular:build`
  → Already has `dependsOn: ["^build"]` on storybook target — **no change needed** ✅

- [x] ~~`content/test`, `router/test` — per-target `^build` + global default removed~~
  → Fixed: restored `dependsOn: ["build"]` on both test targets

## 🔍 Behavioral Changes to Verify

11. `router-plugin.ts` — added change watcher
    → Previously only `add`/`unlink` triggered route invalidation
    → Now **every edit** to route/page/content file → full browser reload (bypasses Vite HMR)
    → **Verify**: is this intentional? DX impact could be significant

12. `cart.cy.ts` — relaxed Cypress assertion
    → `.should('have.been.calledOnce')` → `.should('have.been.called')`
    → Could mask regressions if exactly-once behavior was intended

13. `trpc-client.ts` — replaced `AppRouter` import with hand-written `TrpcAppClient` type
    → Workaround for build-graph issues
    → **Downside**: hand-written type will drift silently; `createTrpcClient<any>` loses compile-time safety

14. **All build/test/e2e caching disabled** (`nx.json` → `"cache": false`)
    → Reasonable for correctness validation phase
    → **But**: significantly slows CI & local dev → should document as temporary or revert before merge

## 🛠️ Dagger CI – Additional Issues

- [x] ~~`disableDefaultFunctionCaching: true` in `dagger.json` — no justification~~
  → Fixed: removed from `dagger.json`

- [x] ~~`@argument({ ignore })` copy-pasted across **8+ functions**~~
  → Fixed: extracted `SOURCE_IGNORE` constant shared by all functions

- [x] ~~`.daggerignore` vs `@argument({ ignore })` — overlap but not identical~~
  → Fixed: aligned both to the same set of patterns

- [x] ~~`prettier` command uses `pnpm exec` instead of `pnpm run`~~
  → Fixed: changed to `pnpm run prettier:check` so it tracks the script definition

- [ ] `e2e` now depends on full `ci-linux` (prettier + lint + build + test)
  → Original CI only depended on `build` → adds unnecessary latency
  → **Intentional design choice** — Dagger shares container cache across stages

## ✅ What's Correct and Valuable

- `pnpm-workspace.yaml` rewrite — old file was badly broken (only `docs-app` listed); new glob patterns are correct
- `pnpm` overrides using `$` syntax — properly deduplicates `vite`, `less`, `rolldown`, `sass`, `sass-embedded`
- `@nx/js:tsc` → Vite build migration — clean & consistent across packages
- `angular-vite-plugin.ts` bug fix — `invalidateTsconfigCaches()` on add/unlink is important for hot-added pages
- `normalizePath` fixes in `angular-vite-plugin.ts` & `live-reload-plugin.ts` — genuine cross-platform correctness
- `setup-testbed.ts` — fixed missing spread (`providers` → `...providers`) — real latent bug
- Export type housekeeping, `route-config.ts` type narrowing, `tsconfig.lib.json` path cleanup — all solid
- `sharedGlobals: [pnpm-lock.yaml]` in `nx.json` — critical for correct cache invalidation
- `sass` second commit — correctly adds as devDep + uses `$sass` override pattern
- `copy-src-assets.mts` — well-structured utility for assets/src handling

## Verdict

**Core workspace correctness** (pnpm workspace, overrides, Vite migration, tsconfig cleanup) is **solid and necessary**.
Most source changes are **genuine bug fixes**.

All critical bugs and missing dependency edges have been addressed. Remaining items are behavioral changes to verify (11–14) and the e2e dependency ordering (19) which is an intentional Dagger design choice.

benpsnyder:
This document explains what this PR changes compared to the current `alpha` branch, and why it matters for contributors, CI reliability, and downstream package consumers.

---

## 1. pnpm workspace correctness eliminates phantom dependencies

### How alpha works today

`pnpm-workspace.yaml` only lists `apps/docs-app`. Every other app and library is invisible to pnpm as a workspace member. They rely on implicit hoisting to find their dependencies at runtime.

This creates **phantom dependency** problems: a package can `import` something it never declared in its own `package.json`, and it works — until pnpm's hoisting algorithm changes between versions, or a CI cache shifts, or a contributor has a different lockfile state. The failure mode is silent and delayed.

### What this PR changes

Every app and library now has a `package.json` and is listed in `pnpm-workspace.yaml`. pnpm treats them as first-class workspace members and resolves their dependencies through the workspace protocol.

### Why this matters

- **Reproducibility**: `pnpm install` produces the same dependency tree on every machine, every time. No more "works on my machine because the hoisting happened to expose the right transitive dependency."
- **Catch missing dependencies early**: If a package uses `marked` but doesn't declare it, pnpm will error during install — not at runtime, not in CI, not when a user installs the published package.
- **Lockfile accuracy**: The lockfile now reflects the real dependency graph of the entire workspace, making upgrades and audits reliable.

---

## 2. Vite lib builds replace @nx/js:tsc for package compilation

### How alpha works today

Packages are compiled by the `@nx/js:tsc` executor, which runs `tsc` and copies the output to `node_modules/`. This has several limitations:

- **Output format is fragile**: The emitted JavaScript depends on `tsconfig.json` settings (`module`, `moduleResolution`, `target`) that can drift between packages or between the tsconfig used for building vs. the one used for IDE support.
- **No bundler-level optimizations**: tsc doesn't tree-shake, doesn't dead-code-eliminate, and doesn't produce source maps that align with the actual module boundaries consumers see.
- **Different toolchain from consumers**: Apps use Vite to build, but the libraries they consume were compiled by tsc. Mismatches between how tsc resolves modules and how Vite resolves them cause subtle issues (e.g., the `.js` → `.ts` extension rewriting that broke Windows builds).

### What this PR changes

Every package now builds with `vite build` using `preserveModules: true`. The output goes directly to `node_modules/@analogjs/<package>/` with the same module structure as the source.

```
packages/platform/src/index.ts          →  node_modules/@analogjs/platform/src/index.js
packages/platform/src/lib/router-plugin.ts  →  node_modules/@analogjs/platform/src/lib/router-plugin.js
```

Declaration files (`.d.ts`) are emitted separately by `tsc --emitDeclarationOnly`.

### Why this matters

- **Single toolchain**: Libraries and apps both build with Vite/rolldown. What works at build time works at serve time. No more "tsc compiled it differently than Vite resolves it."
- **Tree-shakeable output**: `preserveModules` keeps each source module as a separate output file. Bundlers that consume these packages (including Vite itself) can tree-shake at the module level, dropping unused exports.
- **Accurate source maps**: Each output `.js.map` points to exactly one source `.ts` file. Stack traces and debugger breakpoints land in the right place.
- **Faster iteration**: Vite's build with rolldown is significantly faster than tsc for the same file set, especially for packages with many internal modules.

---

## 3. New Vite lib build pipeline is Windows-safe from the start

### How alpha works today

On alpha, packages build with `@nx/js:tsc`, which works on Windows. The Windows CI job passes.

### What this PR introduces — and why it needed platform fixes

This PR replaces the tsc-based pipeline with Vite lib builds (section 2 above). The new pipeline initially broke on Windows because three patterns common in Vite/rolldown configs assume Unix paths:

1. The `external` function uses `!id.startsWith('/')` to detect absolute paths. On Windows, absolute paths start with a drive letter (`D:\`), not `/`. Internal modules get wrongly externalized.
2. `preserveModulesRoot: __dirname` passes a Windows backslash path to rolldown's Rust binding, which normalizes paths to forward slashes internally. The prefix stripping fails, and output files land at wrong paths.
3. String lib entries (e.g. `entry: 'src/index.ts'`) produce ambiguous chunk names that may not match `package.json` exports on Windows.

These are fixed with `path.isAbsolute(id)`, `normalizePath(__dirname)`, and explicit object entries respectively.

### Why this matters

- **The new build pipeline works everywhere from day one**: Rather than shipping a Unix-only pipeline and fixing Windows later, these fixes ensure the Vite lib build approach works on all platforms before it replaces tsc.
- **CI validates cross-platform correctness**: The Windows and macOS CI jobs catch platform-specific bugs before they reach published packages. This is especially important for a framework that runs in diverse environments (developer machines, CI servers, cloud deployments).
- **Contributors on Windows are not left behind**: The new toolchain works on Windows out of the box. No workarounds needed.

---

## 4. Explicit Nx caching strategy prevents stale builds

### How alpha works today

Nx caching defaults are in effect for most build targets. Some package builds write output to `node_modules/`, but Nx doesn't reliably restore cached output to `node_modules/` (it's designed for `dist/`). This creates a failure mode:

1. First CI run: package builds, output goes to `node_modules/`, Nx caches the task.
2. Second CI run: `node_modules/` is recreated by `pnpm install` (fresh). Nx sees a cache hit and skips the build. But Nx doesn't restore the output to `node_modules/`. The package is missing.

This manifests as intermittent CI failures — builds pass on the first run but fail on cache hits.

### What this PR changes

Build targets that write to `node_modules/` have `"cache": false` in their `project.json`. Targets that write to `dist/` keep caching enabled. The `AGENTS.md` documents this as a project convention.

### Why this matters

- **No more mystery CI failures**: The "worked yesterday, fails today" pattern from stale Nx cache hits is eliminated.
- **Clear convention for new packages**: When adding a new build target, the rule is simple — if it writes to `node_modules/`, set `cache: false`. This is documented and enforced by convention.
- **Caching still works where it's safe**: Targets that write to `dist/` (like the docs app) continue to benefit from Nx's remote and local caching.

---

## 5. Dagger CI provides reproducible Linux builds

### How alpha works today

All CI jobs run directly on GitHub Actions runners. The build environment depends on whatever Node version, system libraries, and global state the runner provides. Caching is managed through `actions/cache` and Nx Cloud.

### What this PR changes

Linux CI (build, test, e2e) runs inside Dagger containers with a pinned base image and explicit dependency installation. Windows and macOS jobs continue to use native runners (Dagger doesn't support them).

### Why this matters

- **Reproducible builds**: The same container, same Node version, same system libraries — every time. No more "the runner was updated and now builds fail."
- **Local CI debugging**: Contributors can run `dagger call ci-checks --source=.` locally to reproduce the exact CI environment, without pushing to GitHub and waiting for a runner.
- **Separation of concerns**: Linux validates correctness in a controlled container. Native runners (Windows, macOS) validate platform-specific behavior. Each job tests what it's best at.

---

## 6. Removal of @analogjs/trpc reduces maintenance surface

### How alpha works today

The `@analogjs/trpc` package, `trpc-app`, and `trpc-app-e2e-playwright` exist in the monorepo. They add to build time, dependency count, and maintenance burden even though the integration is no longer actively maintained.

### What this PR changes

All trpc-related code is removed.

### Why this matters

- **Fewer dependencies**: Removes tRPC-specific dependencies from the lockfile, reducing install size and audit surface.
- **Faster builds**: Nx no longer builds, tests, or lints the trpc package and its e2e app.
- **Clearer scope**: The framework's package list reflects what is actively maintained and supported.

---

## Summary: before and after

| Concern | Alpha (today) | This PR |
|---------|--------------|---------|
| pnpm workspace | Partial — only docs-app | Complete — all apps and libs |
| Package compiler | tsc via @nx/js:tsc | Vite lib build with rolldown |
| Windows builds | Works (tsc pipeline) | Works (new Vite pipeline, with platform fixes) |
| CI environment | Bare GitHub runners | Dagger containers (Linux) + native (Win/Mac) |
| Nx caching | Implicit, sometimes stale | Explicit — `cache: false` for node_modules outputs |
| trpc maintenance | Carried but unmaintained | Removed |
| Build scripts | Platform-specific | Cross-platform (copy-src-assets, assert-deps) |

benpsnyder:
@brandonroberts this PR really helps agentic development and testing loops due to dagger.
I will endevour to optimize the CI speed on future PRs.
I did not yet do a line-by-line manual review of the whole PR; I did ask Claude to help make scoped commits for a rebase merge so it is ready from that viewpoint.

brandonroberts:
Overall, the setup looks good. There are some missing artifacts from the builds

`astro-angular` -> missing all the build output in the `src` folder. 
`platform` -> missing the `src/lib/nx-plugin` folder that contains the executors/schematics/preset.
`storybook-angular` -> missing all the build output in the `src/lib` folder. There's only json files.
`vite-plugin-angular` -> missing the `src/lib/tools` output that contains the builders/schematics.

benpsnyder:
> Overall, the setup looks good. There are some missing artifacts from the builds
> 
> `astro-angular` -> missing all the build output in the `src` folder. `platform` -> missing the `src/lib/nx-plugin` folder that contains the executors/schematics/preset. `storybook-angular` -> missing all the build output in the `src/lib` folder. There's only json files. `vite-plugin-angular` -> missing the `src/lib/tools` output that contains the builders/schematics.

I addressed these concerns and added regression tests.

PR #2133, opened 2026-03-16
fix: support astro v6 using vite environment api

### LINKED ISSUE #2112: NG0401 error on the client after building an Astro 6 project

Reproduction: https://github.com/eblocha/astro-angular-experiments/tree/astro-6-vite-7

Run `pnpm build` then `pnpm preview` and go to `http://localhost:4321/guides/example`.

astro-angular

### Description

I recently updated to Astro 6 with vite 7, and the astro-angular integration is now failing on the client during hydration with NG0401 (missing platform).

It's hitting this condition in `@angular/core`: https://github.com/angular/angular/blob/main/packages/core/src/application/create_application.ts#L47

This is what the built code looks like:

Image

The `!i` is the `!platformRef` from the source.

This seems to suggest that `ngServerMode` is defined as truthy during the client code build, which folds the check to `!platformRef`.

When built with Astro 5 and vite 6, this check has been removed (stepping through with a debugger shows platformRef is still undefined here).

Image

```true
Uncaught (in promise) Error: NG0401
    S http://localhost:4321/_astro/common.OwyWT5em.js:3
    op http://localhost:4321/_astro/common.OwyWT5em.js:4
    he http://localhost:4321/_astro/client.DInVqwOQ.js:1
    Se http://localhost:4321/_astro/client.DInVqwOQ.js:1
    y http://localhost:4321/guides/example:103
    r http://localhost:4321/guides/example:103
common.OwyWT5em.js:3:13394
```

### Other information

_No response_

- [x] Yes
- [ ] No

eblocha:
## PR Checklist

Closes #2112

## What is the new behavior?

During the client build in astro projects, this will set the `ngServerMode` define global to `'false'`. This avoids building an invalid, broken client bundle.

- [ ] Yes
- [x] No

Flex seal

### DIFF
--- a/packages/astro-angular/src/index.ts
+++ b/packages/astro-angular/src/index.ts
@@ -14,7 +14,7 @@ function getRenderer(): AstroRenderer {
   };
 }
 
-function getViteConfiguration(vite?: PluginOptions) {
+function getViteConfiguration(vite?: PluginOptions): ViteUserConfig {
   return {
     esbuild: {
       jsxDev: true,
@@ -48,6 +48,20 @@ function getViteConfiguration(vite?: PluginOptions) {
           }
 
           return;
+        },
+      },
+      {
+        name: 'analogjs-astro-client-ngservermode',
+        configEnvironment(name: string) {
+          if (name === 'client') {
+            return {
+              define: {
+                ngServerMode: 'false',
+              },
+            };
+          }
+
+          return undefined;
         },
       },
     ],


eblocha:
@brandonroberts I'm not quite sure why the vercel preview failed. From the logs it appears to be downgrading a lot of packages, and failing a type check when building the `platform` project. The build works fine locally for me.

brandonroberts:
This isn't related to your changes @eblocha

PR #2145, opened 2026-03-21
refactor(nx-plugin): use default-only exports for generators

benpsnyder:
## Affected scope

- Primary scope: nx-plugin
- Secondary scopes: none

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

## What is the new behavior?

Converts four generator entry modules from dual (named + default) exports to default-only exports, eliminating `MIXED_EXPORTS` bundler warnings and removing dead code.

### Why

When a module has both `export async function foo()` and `export default foo`, Rolldown emits a `MIXED_EXPORTS` warning because it cannot determine whether the module should be treated as a single-value default export or a multi-value named export. This is relevant now that the nx-plugin builds through Vite/Rolldown as part of the Vite+OXC lib pipeline migration — these warnings clutter CI logs and mask real issues.

Nx resolves generators via the default export (configured in `generators.json` → `compat.ts` → `export default`), so the named exports on the generator functions themselves serve no purpose. Removing them:

- **Eliminates `MIXED_EXPORTS` warnings** from the nx-plugin build output, making CI logs cleaner and real warnings more visible.
- **Removes dead code** — `analogPageGeneratorSchematic` (a `convertNxGenerator` wrapper) was never referenced by any consumer in the codebase, published entry points, or `generators.json`. It was leftover scaffolding from the original Nx schematic compatibility layer.
- **Simplifies the module contract** — each generator file now has exactly one export (`export default`), making it immediately clear how the module is consumed.

### Changes

| File | Change |
|---|---|
| `app/generator.ts` | `export async function appGenerator` → `async function appGenerator` (default export retained) |
| `init/generator.ts` | `export async function setupAnalogGenerator` → `async function setupAnalogGenerator` |
| `setup-vitest/generator.ts` | `export async function setupVitestGenerator` → `async function setupVitestGenerator` |
| `page/generator.ts` | Same as above, plus removed dead `analogPageGeneratorSchematic` export and its `convertNxGenerator` import |
| `preset/generator.ts` | Updated dynamic import from `({ appGenerator }) => appGenerator(...)` to `(m) => m.default(...)` to align with default-only export |
| `page/generator.spec.ts` | `import { analogPageGenerator }` → `import analogPageGenerator` (default import) |

## Test plan

- [x] `nx test nx-plugin` — all 21 tests pass (app, init, page, preset, setup-vitest generators)
- [x] `nx build nx-plugin` — clean build with no `MIXED_EXPORTS` warnings

- [ ] Yes
- [x] No

The removed `analogPageGeneratorSchematic` export was dead code — no consumers reference it in the codebase, in `generators.json`, or in any published entry point. The default exports that Nx actually uses are unchanged.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,12 +35,24 @@ jobs:
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v4
-      - uses: actions/setup-node@v4
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
         with:
           node-version-file: .node-version
-      - run: corepack enable && corepack install
-      - uses: actions/setup-node@v4
+      - name: Clear hosted tool-cache for corepack
+        if: runner.os == 'Windows'
+        run: Remove-Item -Recurse -Force (Get-Command corepack.cmd).Path
+        shell: pwsh
+      - name: Install Corepack
+        run: npm install --global --force corepack@latest
+      - name: Enable Corepack
+        run: |
+          corepack enable
+          pnpm --version
+      - name: Setup Node.js
+        uses: actions/setup-node@v4
         with:
+          node-version-file: .node-version
           cache: 'pnpm'
           cache-dependency-path: '**/pnpm-lock.yaml'
       - name: Install
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,16 +28,18 @@ jobs:
         uses: actions/setup-node@v4
         with:
           node-version-file: .node-version
-      - run: npm install --global corepack@0.31.0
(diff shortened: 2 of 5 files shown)


brandonroberts:
Compatibility with Angular schematics needs to be preserved. Or rewrite alll Nx generators as schematics so they are compatible with both Angular/Nx by default without using the `@nx/devkit`.

benpsnyder:
> Compatibility with Angular schematics needs to be preserved. Or rewrite alll Nx generators as schematics so they are compatible with both Angular/Nx by default without using the `@nx/devkit`.

This can merge. Follow-up issue made https://github.com/analogjs/analog/issues/2168

PR #2148, opened 2026-03-21
feat(content): enhance content-file-resource and parse-raw-content-file with schema validation

benpsnyder:
## Affected scope

- Primary scope: content
- Secondary scopes: none

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

## What is the new behavior?

Adds [Standard Schema](https://github.com/standard-schema/standard-schema)-based frontmatter validation to the content pipeline. This allows consumers to validate and type-narrow frontmatter attributes at load time using any Standard Schema-compatible library (Valibot, Zod, ArkType, etc.).

### Why this matters

Today, `contentFileResource<BlogAttributes>()` provides TypeScript types through a generic parameter, but there's no runtime guarantee that the frontmatter in a markdown file actually matches those types. A missing `date` field or a misspelled `tags` property silently passes through as `undefined`, leading to runtime errors in templates that are hard to trace back to the source content file.

Schema validation closes this gap by:

- **Catching content errors at build/load time** rather than at render time — invalid frontmatter surfaces immediately with the filename and specific field that failed, not as a cryptic `Cannot read property of undefined` in a template.
- **Providing type-safe attributes without manual casting** — when a schema is provided, the returned `attributes` type is inferred from the schema output, eliminating the need for `as` casts or manual interface definitions that can drift from the actual content.
- **Enabling frontmatter transforms** — schemas can coerce, default, and derive values (e.g., auto-generating a `slug` from `title`, parsing date strings into `Date` objects, or adding computed `readingTime` fields) during the parse step, keeping components clean.

### What this unlocks

This is foundational infrastructure for several planned content features:

- **Content collections with schema enforcement** — define a schema per collection (blog, docs, changelog) and fail the build when any file in the collection has invalid frontmatter. This is the Analog equivalent of Astro's content collections or Contentlayer's schema definitions.
- **Type-safe `.agx` component metadata** — `.agx` file resolution already exists in the content pipeline; this PR extends schema validation to cover `.agx` module metadata (the `{ default, metadata }` export shape), enabling validated frontmatter for Angular component content files.
- **Prerender-time validation** — `parseRawContentFileAsync` supports async schemas, allowing validation that hits external services (e.g., checking that an `og:image` URL resolves) during SSG prerendering.
- **IDE-integrated content authoring** — with schemas as the source of truth for frontmatter shape, tooling can provide autocomplete and validation in markdown files.

### New APIs

- **`parseRawContentFileAsync(rawContent, schema, filename?)`** — async variant of `parseRawContentFile` that supports async schema validators.
- **`FrontmatterValidationError`** — structured error class with `issues` array and optional `filename` for clear error messages during content builds.
- **`contentFileResource({ schema })`** — new options-object overload that validates frontmatter against a schema before returning the resource. Returns typed attributes inferred from the schema output type.
- **`ContentFileResourceResult<Attributes>`** — exported interface extending `ContentFile` with a `toc` field, formalizing what the resource actually returns.

### Example

```typescript
import * as v from 'valibot';
import { contentFileResource } from '@analogjs/content/resources';

const BlogSchema = v.object({
  title: v.string(),
  date: v.pipe(v.string(), v.isoDate()),
  tags: v.optional(v.array(v.string()), []),
});

// Attributes are typed as { title: string; date: string; tags: string[] }
// Invalid frontmatter throws FrontmatterValidationError at load time
const post = contentFileResource({ schema: BlogSchema });
```

### Backward compatibility

The existing `contentFileResource(params?, fallback?)` positional API is preserved via function overloads. The existing `parseRawContentFile<Attributes>(raw)` generic API is unchanged. Consumers in `blog-app`, `opt-catchall-app`, and `@analogjs/router` continue to work without modification. No breaking changes.

### Changes

- **`content-file-resource.ts`** — added `ContentFileResourceResult` type, `validateAttributes` helper, options-object overload with schema support, module metadata validation for `.agx` files
- **`parse-raw-content-file.ts`** — added `FrontmatterValidationError` class, schema overload for `parseRawContentFile`, new `parseRawContentFileAsync` function
- **`src/index.ts`** — exported `parseRawContentFileAsync`, `FrontmatterValidationError`; changed `ContentFile` and `InjectContentFilesFilterFunction` to type-only exports (fixes workspace development resolution)
- **`package.json`** — added `@standard-schema/spec` as optional peer dependency
- **New:** `parse-raw-content-file.spec.ts` — 22 tests covering sync/async validation, error formatting, schema transforms, multiple issues, nested paths, edge cases
- **Updated:** `content-file-resource.spec.ts` — 3 new tests for async schema validation, module metadata validation, and validation error handling

## Test plan

- [x] `nx test content` — all 62 tests pass (37 existing + 25 new)
- [x] `nx build content`

- [ ] Yes
- [x] No

`@standard-schema/spec` is an optional peer dependency — consumers only need it if they use schema validation. The Standard Schema spec is a lightweight type-only contract (~1KB) adopted by Valibot, Zod 4, ArkType, and other validation libraries.

### DIFF
--- a/packages/content-plugin/eslint.config.mjs
+++ b/packages/content-plugin/eslint.config.mjs
@@ -23,7 +23,10 @@ export default [
   {
     files: ['**/*.json'],
     rules: {
-      '@nx/dependency-checks': 'error',
+      '@nx/dependency-checks': [
+        'error',
+        { ignoredFiles: ['**/vite.config.lib.ts'] },
+      ],
     },
     languageOptions: {
       parser: await import('jsonc-eslint-parser'),
--- a/packages/content-plugin/src/migrations/update-markdown-version/update-markdown-version.ts
+++ b/packages/content-plugin/src/migrations/update-markdown-version/update-markdown-version.ts
@@ -1,6 +1,6 @@
 import {
   addDependenciesToPackageJson,
-  Tree,
+  type Tree,
   formatFiles,
   installPackagesTask,
 } from '@nx/devkit';
--- a/packages/storybook-angular/src/lib/preset.spec.ts
+++ b/packages/storybook-angular/src/lib/preset.spec.ts
@@ -346,7 +346,7 @@ describe('viteFinal', () => {
       expect(transformPlugin).toBeDefined();
     });
 
-    it('should use esbuild config key when rolldownVersion is not available', async () => {
+    it('should use esbuild config key with keepNames on Vite 6-7', async () => {
       const options = createMockOptions();
       const result = await viteFinal(baseConfig, options);
 
@@ -360,30 +360,19 @@ describe('viteFinal', () => {
       expect(pluginConfig.esbuild.keepNames).toBe(true);
     });
 
-    it('should use oxc config key when rolldownVersion is available', async () => {
+    it('should use oxc config key with keepNames on Vite 8+ (Rolldown)', async () => {
       vi.resetModules();
-      vi.doMock('@storybook/angular/preset', () => ({
-        core: async () => ({
-          options: {},
-          channelOptions: { wsToken: 'mock-token' },
-        }),
-        addons: [],
-      }));
-      vi.doMock('storybook/internal/types', () => ({}));
-      vi.doMock('@storybook/angular', () => ({}));
-      vi.doMock('@storybook/builder-vite', () => ({}));
+      registerDependencyMocks();
       vi.doMock('vite', () => ({
         mergeConfig: (_base: unknown, override: unknown) => override,
         normalizePath: (p: string) => p,
         rolldownVersion: '1.0.0',
       }));
-      vi.doMock('@analogjs/vite-plugin-angular', () => ({
-        default: () => ({ name: 'angular-mock' }),
-      }));
-
       const mod = await import('./preset');
+      const freshViteFinal = mod.viteFinal;
+
       const options = createMockOptions();
-      const result = await mod.viteFinal(baseConfig, options);
+      const result = await freshViteFinal(baseConfig, options);
 
       const transformPlugin = result.plugins
         .flat()
--- a/packages/storybook-angular/src/lib/preset.ts
+++ b/packages/storybook-angular/src/lib/preset.ts
@@ -1,8 +1,8 @@
 import { resolve } from 'node:path';
 import { core as PresetCore } from '@storybook/angular/preset';
 import { fileURLToPath } from 'node:url';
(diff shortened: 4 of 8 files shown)


brandonroberts (approved):
Approved

benpsnyder:
> Approved

This is ready to merge.
I looked into whether we could pull in Valibot to simplify the schema mocks in the test files (`parse-raw-content-file.spec.ts` and `content-file-resource.spec.ts`). Here's a summary of the trade-offs.

### What Valibot would simplify

A handful of happy-path tests currently hand-roll `StandardSchemaV1` objects with inline validate functions. For example:

```ts
const schema = createMockSchema<{ title: string; date: string }>((value) => {
  const v = value as Record<string, unknown>;
  if (v.title && v.date && Array.isArray(v.tags)) {
    return { value: v as { title: string; date: string; tags: string[] } };
  }
  return { issues: [{ message: 'Invalid' }] };
});
```

With Valibot this becomes:

```ts
const schema = v.object({ title: v.string(), date: v.string(), tags: v.array(v.string()) });
```

More declarative, and it doubles as an integration smoke test proving the Standard Schema interface works with a real library.

### What it would not simplify

The majority of the test cases exist to exercise the Standard Schema *interface contract* itself, not any particular library's behavior. These still need hand-rolled mocks:

- **Error formatting** — tests that assert on specific issue shapes (e.g., nested path segments like `['meta', { key: 'og:image' }]`) need precise control over the `issues` array that Valibot wouldn't produce in the same form.
- **Async schema detection** — the sync `parseRawContentFile` has a guard that throws when it receives an async schema. Valibot schemas are synchronous, so this test case still requires a manual mock.
- **Value transformation** — tests that uppercase titles or inject computed fields are simpler as inline functions than Valibot `v.pipe()` chains.
- **Controlled failure paths** — tests for multiple validation issues, empty frontmatter against a schema, etc. rely on deterministic issue output that's easier to guarantee with a mock.

Roughly 4-5 tests would get cleaner, while 8-10 would still need the existing mocks.

### Recommendation

Keep the current approach. The hand-rolled mocks are intentionally library-agnostic and give full control over the edge cases being tested. Swapping in Valibot would add a `devDependency` for modest gains and split the test strategy between two patterns (real schemas + mocks). The doc comment already references Valibot as the suggested user-facing library, which is sufficient.

If we later want to add a dedicated integration test that exercises Valibot (or Zod) end-to-end, that could live as a separate test file without disrupting the existing coverage.

PR #2149, opened 2026-03-21
feat(router): add typed server primitives and TanStack Query integration

benpsnyder:
## Affected scope

- Primary scope: router
- Secondary scopes: none

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

## What is the new behavior?

This PR expands `@analogjs/router` with two tightly integrated systems:

1. **Typed server primitives** — `defineAction`, `defineServerRoute`, and
   `definePageLoad` handle request parsing, Standard Schema validation, and
   structured error responses at the server boundary.
2. **TanStack Query integration** — `@analogjs/router/query` bridges those
   server routes into TanStack Query with SSR hydration, end-to-end type
   inference, and zero manual type duplication between server and client.

### Why this matters

Angular lacks a first-party story for colocated server logic with client-managed
cache state. Developers currently hand-roll request parsing, duplicate
request/response types between server routes and client pages, and wire up
TanStack Query with manual `HttpClient` boilerplate. This PR provides:

- **Schema-validated server boundaries** — `defineServerRoute` validates query
  params, request bodies, and route params against any Standard Schema library
  (Valibot, Zod, ArkType). Invalid input returns 422 with structured issues.
- **End-to-end type safety** — `defineServerRoute` returns a branded
  `ServerRouteHandler<TQuery, TBody, TResult>` type. Client-side utilities
  (`serverQueryOptions`, `serverMutationOptions`) extract those types via
  `InferRouteQuery`, `InferRouteBody`, `InferRouteResult` — so query params,
  mutation inputs, and response shapes all flow from the server definition with
  no manual type duplication.
- **SSR hydration** — `provideAnalogQuery` dehydrates TanStack Query state
  during server rendering and rehydrates it on the client via Angular's
  `TransferState`, preventing duplicate fetches after navigation.

### TanStack Query Angular + Analog pairing

| TanStack Query Angular API              | Analog Integration                                                                     | What Analog adds                                                                               |
| --------------------------------------- | -------------------------------------------------------------------------------------- | ---------------------------------------------------------------------------------------------- |
| `provideTanStackQuery(client)`          | `provideAnalogQuery(client, ...features)`                                              | Wraps TanStack provider + wires SSR dehydration/hydration via `TransferState`                  |
| `QueryClient` + `dehydrate` / `hydrate` | `provideServerAnalogQuery()`                                                           | Hooks into `BEFORE_APP_SERIALIZED` to transfer query cache from server to client               |
| `injectQuery(() => options)`            | `injectQuery(() => serverQueryOptions<typeof route>(http, url, opts))`                 | Query params typed from route's `query` schema; response typed from handler return             |
| `injectMutation(() => options)`         | `injectMutation(() => serverMutationOptions<typeof route>(http, url, opts))`           | Mutation input typed from route's `body` schema; `onMutate`/`onError`/`onSettled` pass through |
| `injectInfiniteQuery(() => options)`    | `injectInfiniteQuery(() => serverInfiniteQueryOptions<typeof route>(http, url, opts))` | Pagination query params typed from route schema via `query: ({pageParam}) => ...`              |
| `queryOptions()`                        | `serverQueryOptions<typeof route>()`                                                   | Replaces manual `queryOptions` + `HttpClient` boilerplate with type-safe one-liner             |
| `mutationOptions()`                     | `serverMutationOptions<typeof route>()`                                                | Replaces manual `mutationOptions` + `HttpClient.post` with type-safe one-liner                 |

### New APIs

**Server-side primitives** (`@analogjs/router/server/actions`)

| Function            | Signature                                                | Purpose                                                                                                              |
| ------------------- | -------------------------------------------------------- | -------------------------------------------------------------------------------------------------------------------- |
| `defineAction`      | `({ schema?, handler })`                                 | Form/JSON action with optional Standard Schema validation. Handler receives validated `data`.                        |
| `defineServerRoute` | `({ query?, body?, params?, input?, output?, handler })` | API route handler with per-source schema validation. Returns branded `ServerRouteHandler<TQuery, TBody, TResult>`.   |
| `definePageLoad`    | `({ params?, query?, handler })`                         | Page data loader with route param/query validation. Return type flows to client via `injectLoadData<typeof load>()`. |

**Type extraction utilities** (`@analogjs/router/server/actions`)

| Type                                         | Extracts                                                    |
| -------------------------------------------- | ----------------------------------------------------------- |
| `ServerRouteHandler<TQuery, TBody, TResult>` | Branded handler type preserving schema and result generics  |
| `InferRouteQuery<typeof route>`              | Validated query param shape from the route's `query` schema |
| `InferRouteBody<typeof route>`               | Validated body shape from the route's `body` schema         |
| `InferRouteResult<typeof route>`             | Handler return type (excluding `Response`)                  |

**TanStack Query bridge** (`@analogjs/router/query`)

| Function                                                    | Purpose                                                                                                                                              |
| ----------------------------------------------------------- | ---------------------------------------------------------------------------------------------------------------------------------------------------- |
| `provideAnalogQuery(client, ...features)`                   | Client-side: provides TanStack Query + hydrates cache from SSR `TransferState`                                                                       |
| `provideServerAnalogQuery()`                                | Server-side: dehydrates query cache before HTML serialization. Isolated import path to avoid pulling `@angular/platform-server` into client bundles. |
| `serverQueryOptions<typeof route>(http, url, opts)`         | Creates typed `CreateQueryOptions` from a branded server route — query params and response inferred                                                  |
| `serverMutationOptions<typeof route>(http, url, opts)`      | Creates typed `CreateMutationOptions` — mutation body and response inferred                                                                          |
| `serverInfiniteQueryOptions<typeof route>(http, url, opts)` | Creates typed `CreateInfiniteQueryOptions` with pagination param mapping                                                                             |

**Validation and parsing** (`@analogjs/router/server/actions`)

- `validateWithSchema(schema, data)` — validates against any Standard Schema
  (Valibot, Zod, ArkType). Returns `{ value }` or `{ issues }`.
- `parseRequestData(event)` — auto-detects JSON, FormData, or URL params and
  parses into a plain object.
- `fail(status, issues)` / `json(data)` / `redirect(url)` — response helpers.

**Client-side**

- `FormAction` directive — intercepts form submissions, handles
  redirect/navigate detection, emits `(onSuccess)`, `(onError)`, `(state)`.
  Adds `data-state` and `aria-busy` host attributes.
- `injectLoadData<typeof load>()` — extracts typed page load data, excluding
  `Response` from the union.

### Recommended patterns

**Method-suffix route files** — for routes serving both GET and POST, split into
Nitro method files (`query-todos.get.ts` + `query-todos.post.ts`). Each has a
single return type, avoiding union narrowing on the client. Shared state lives in
`_` prefixed utility files (e.g. `_query-todos-state.ts`).

**`import type` from server routes** — client pages import `typeof route` from
server files using `import type`, which is erased at build time. No server code
leaks into the client bundle:

```typescript
import type { route as todosQuery } from '../../server/routes/api/v1/query-todos.get';
import { serverQueryOptions, injectQuery } from '@analogjs/router/query';

todosQuery = injectQuery(() =>
  serverQueryOptions<typeof todosQuery>(this.http, '/api/v1/query-todos', {
    queryKey: ['todos', this.scope()],
    query: { scope: this.scope() }, // ← typed from Valibot schema
    staleTime: 60_000,
  }),
);
```

### Examples included

| Page                         | Features demonstrated                                                            |
| ---------------------------- | -------------------------------------------------------------------------------- |
| `/contact`                   | `defineAction` with Valibot schema, `FormAction` directive, validation errors    |
| `/greet/:name`               | `definePageLoad` with params + query validation, `injectLoadData<typeof load>()` |
| `/tanstack-query`            | `serverQueryOptions` + `serverMutationOptions` with branded types, SSR hydration |
| `/tanstack-query-multi`      | Multiple independent queries + dependent query chain (`enabled` guard)           |
| `/tanstack-query-infinite`   | `serverInfiniteQueryOptions` with cursor pagination, client-side load-more       |
| `/tanstack-query-optimistic` | Optimistic mutation with `onMutate` snapshot, `onError` rollback                 |

### What this unlocks

- Colocated server logic with request parsing and schema validation at the
  framework boundary
- Shared server → client type contracts with zero manual duplication
- Progressive-enhancement-friendly forms with structured validation failures
- First-class TanStack Query integration with SSR hydration that fits Analog's
  router and request-context model
- Method-split routes for clean single-return-type handlers

## Test plan

- [x] `nx test router` — unit tests pass
- [x] `nx build router` — package builds
- [x] `nx build analog-app` — app builds with type-safe imports
- [x] `nx lint router` — lint passes
- [x] `nx lint analog-app` — lint passes
- [x] `nx e2e analog-app-e2e` — all 31 E2E tests pass
- [x] Manual verification of `/contact`, `/greet/analog`, `/tanstack-query`,
      `/tanstack-query-multi`, `/tanstack-query-infinite`,
      `/tanstack-query-optimistic`

- [ ] Yes
- [x] No

All new APIs are additive. The `FormAction` directive preserves the existing
`(state)` output name from `analogjs/alpha`.

## Other information

**Architecture decisions:**

- `provideServerAnalogQuery` is exported from a dedicated sub-path
  (`@analogjs/router/query/provide-server-analog-query`) rather than the main
  `@analogjs/router/query` barrel, so client-only imports never resolve
  `@angular/platform-server`.
- `@standard-schema/spec` added to workspace root dependencies for type
  resolution across Angular and Nitro build contexts.
- Server routes use Valibot for schema definitions, consistent with the existing
  `contact.server.ts` and `greet.[name].server.ts` patterns.
- `buildUrl` in `serverQueryOptions` embeds query params directly in the URL
  string as a workaround for a known `requestContextInterceptor` limitation
  where `HttpRequest.params` are not forwarded during SSR internal fetch.


brandonroberts:
defineApiRoute for something that's not an API route feels awkward

SOG-web:
> defineApiRoute for something that's not an API route feels awkward

I think this can be useful since it more typed, but at the same time it will create confusion with nitro api route

also this might lock analog on nitro 

but overall it worth exploring and maybe we could use it as the default api route for analog, users will only have to use the native nitro way if they need to do something defineApiRoute does not support

packages/router/server/actions/src/define-action.ts
```diff
@@ -0,0 +1,130 @@
...
+export interface DefineActionOptions<
+  TSchema extends StandardSchemaV1,
+  TParamsSchema extends OptionalSchema = undefined,
+> {
+  schema: TSchema;
+  params?: TParamsSchema;
+  handler: (
+    context: DefineActionContext<TSchema, TParamsSchema>,
+  ) => Promise<Response> | Response;
+}
```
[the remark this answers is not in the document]
benpsnyder: All 121 tests pass (including all 13 define-action tests). Here's what changed:

  - DefineActionContext<TSchema> — TSchema constraint relaxed from StandardSchemaV1 to OptionalSchema (defaults to undefined); data now uses InferSchema<TSchema, Record<string, unknown>> so it falls back to Record<string, unknown> when no schema is
  provided
  - DefineActionOptions<TSchema> — same constraint change; schema is now optional (schema?: TSchema)
  - defineAction / handleValidatedRequest — schema validation only runs when options.schema is provided; otherwise the raw parsed body is passed through as data

  Schema-less actions can now be created without a pass-through schema:
  const action = defineAction({
    handler: async ({ data }) => {
      // data is Record<string, unknown>
      return json({ ok: true });
    },
  });

packages/router/server/actions/src/define-page-load.ts
```diff
@@ -0,0 +1,131 @@
...
+        (ctx.event as H3Event & { request: Request }).request.url,
+        'http://localhost',
+      );
+      const rawQuery = parseSearchParams(url.searchParams);
+      const result = await validateWithSchema(options.query, rawQuery);
+      if (result.issues) {
+        return fail(422, result.issues);
+      }
+      query = result.value;
+    }
```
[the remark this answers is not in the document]
benpsnyder: Added a fix just now and tests

packages/router/server/actions/src/index.ts
```diff
@@ -1 +1,16 @@
...
+  DefineApiRouteContext,
+  DefineApiRouteOptions,
+  DefineApiRouteResult,
+} from './define-api-route';
+export { definePageLoad } from './define-page-load';
+export type {
+  PageLoadContext,
+  DefinePageLoadOptions,
+} from './define-page-load';
+export { validateWithSchema } from './validate';
```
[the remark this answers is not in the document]
benpsnyder: parseRequestData is an internal utility — it parses raw request bodies for defineAction and defineApiRoute. It's not something consumers would call directly; they use defineAction/defineApiRoute which call it internally.

  I don't think this should be exported. The public API (defineAction, defineApiRoute, definePageLoad) already handles request parsing for the user. Exposing parseRequestData would add surface area for an implementation detail that could change. No fix needed.

packages/router/server/actions/src/parse-request-data.ts
```diff
@@ -0,0 +1,110 @@
...
+
+function getContentType(event: { method: string; headers: Headers }): string {
+  const request = getRequest(event);
+
+  return (
+    request.headers.get('content-type') ??
+    event.headers.get('content-type') ??
+    event.headers.get('Content-Type') ??
+    ''
+  );
```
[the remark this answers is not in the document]
benpsnyder: No fix needed. The as unknown as cast in getRequest is intentional — H3Event doesn't expose request in its type declarations, but it exists at runtime. The loose parameter type is the correct approach here given the h3 type gap.

packages/router/server/actions/src/parse-request-data.ts
```diff
@@ -0,0 +1,110 @@
...
+
+  if (isFormContentType(contentType)) {
+    return parseFormData(await request.formData());
+  }
+
+  try {
+    return await request.json();
+  } catch {
+    return {};
+  }
```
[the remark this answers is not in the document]
benpsnyder: All tests pass. The JSON fast path now catches malformed JSON and returns {} instead of throwing a 500, matching the fallback behavior at the end of the function.

benpsnyder:
Thanks for the feedback @brandonroberts and @SOG-web — both valid points.

Renamed `defineApiRoute` → `defineServerRoute` across the board (function, types, tests, barrel exports). The new name scopes it as Analog's server-layer abstraction without colliding with Nitro's native `defineEventHandler` / API route convention.

On the Nitro coupling concern — this wrapper doesn't introduce new coupling beyond what's already there (`H3Event` is the runtime reality for all Analog server routes). It just makes the existing contract type-safe. Users who need raw Nitro access can still use `defineEventHandler` directly, and the typed validation layer (`input`/`query`/`body`/`params`/`output`) is framework-agnostic via Standard Schema.

Changes in this update:
- `defineApiRoute` → `defineServerRoute`
- `DefineApiRouteContext` → `DefineServerRouteContext`
- `DefineApiRouteOptions` → `DefineServerRouteOptions`
- `DefineApiRouteResult` → `DefineServerRouteResult`
- `defineAction` schema is now optional — handlers without a schema receive the raw parsed body as `Record<string, unknown>`
- `definePageLoad` now always parses URL query params, even without a query schema — handlers previously received `{}` instead of actual query values
- JSON fast-path in `parseRequestData` now catches malformed bodies instead of throwing a 500

benpsnyder:
I'm working on E2E tests and refactoring this. Hang tight.

benpsnyder:
# Idiomatic Angular Report

Assessment of `feat/router-server-actions` against Angular main (framework source)
and modern Angular idioms. Intended for maintainer review.

---

## Angular Idiom Alignment Summary

| Area | Alignment | Notes |
|------|-----------|-------|
| Standalone components | Excellent | All components use `standalone: true`, no NgModules |
| Signal-based state | Excellent | `signal()`, `computed()`, `toSignal()` throughout — no legacy `@Input`/`@Output` decorators |
| Modern control flow | Excellent | `@if`/`@for` everywhere — no `*ngIf`/`*ngFor` |
| `ChangeDetectionStrategy.OnPush` | Excellent | Every component |
| Functional DI (`inject()`) | Excellent | No constructor injection anywhere |
| Provider composition | Excellent | `makeEnvironmentProviders()`, functional interceptors via `withInterceptors()` |
| SSR/hydration | Excellent | `TransferState`, `BEFORE_APP_SERIALIZED`, `provideClientHydration(withEventReplay())` |
| Branded/phantom types | Excellent | `ServerRouteHandler<TQuery, TBody, TResult>` mirrors Angular's own `StateKey<T>` pattern |
| RxJS → Signal bridge | Good | `toSignal()` used correctly; `lastValueFrom()` bridges where TanStack expects promises |
| No lifecycle hooks | Excellent | Zero `OnInit`/`OnDestroy` — all reactive via signals and effects |

**Overall: the PR follows Angular's latest idioms (v19+) with no deprecated patterns.**

---

## Angular Main Patterns vs This PR

### 1. `resource()` / `httpResource()` — Angular's Signal-Based Async

Angular 19.2 introduces `httpResource()` — a signal-based replacement for
Observable-based HTTP calls with built-in transfer cache:

```typescript
// Angular's new pattern
todo = httpResource(() => `/api/todos/${this.id()}`);
// todo.value() — signal with response data
// todo.status() — 'idle' | 'loading' | 'resolved' | 'error'
// todo.isLoading() — computed boolean
// todo.reload() — imperative refetch
```

**How this PR relates:**

| Angular `httpResource` | This PR (TanStack Query) | Trade-off |
|---|---|---|
| `httpResource(() => url)` | `injectQuery(() => serverQueryOptions(...))` | httpResource is simpler; TanStack provides caching, invalidation, background refetch |
| `todo.value()` (signal) | `todosQuery.data()` (signal) | Same signal-based consumption |
| `todo.status()` | `todosQuery.status()` | Same status signals |
| `todo.isLoading()` | `todosQuery.isPending()` | Same pattern, different name |
| `todo.reload()` | `queryClient.invalidateQueries()` | TanStack offers key-based selective invalidation |
| No mutation support | `injectMutation(() => serverMutationOptions(...))` | TanStack adds optimistic updates, rollback, cache manipulation |
| No infinite/pagination | `injectInfiniteQuery(() => serverInfiniteQueryOptions(...))` | TanStack provides cursor-based pagination natively |
| Built-in transfer cache | `provideAnalogQuery()` with `TransferState` dehydrate/hydrate | Both achieve SSR state transfer; httpResource does it per-request, TanStack does it for the whole cache |

**Recommendation for maintainer:** `httpResource` is excellent for simple
read-only data. TanStack Query is the right choice when apps need cache
invalidation, mutations, optimistic updates, or infinite queries — which is
exactly what this PR's demo pages exercise. The two approaches are complementary,
not competing. A future PR could add an `analogResource()` wrapper that bridges
`defineServerRoute` types into Angular's `resource()` API for simpler use cases.

### 2. `TransferState` Usage

Angular's own pattern:
```typescript
// Server: register lazy serialization callback
transferState.onSerialize(key, () => computeValue());

// Client: read and remove
const value = transferState.get(key, defaultValue);
transferState.remove(key);
```

This PR's pattern:
```typescript
// Server (BEFORE_APP_SERIALIZED):
transferState.set(ANALOG_QUERY_STATE_KEY, dehydrate(queryClient));

// Client (ENVIRONMENT_INITIALIZER):
const state = transferState.get(ANALOG_QUERY_STATE_KEY, null);
if (state) { hydrate(client, state); transferState.remove(key); }
```

**Assessment:** Correct usage. Uses `set()` during serialization (appropriate
since dehydrate must run eagerly to capture all query state) rather than
`onSerialize()` (which is for lazy computation). The `ENVIRONMENT_INITIALIZER`
hydration runs before component creation, ensuring queries find cached data.

### 3. `ENVIRONMENT_INITIALIZER` vs `provideEnvironmentInitializer()`

Angular now provides `provideEnvironmentInitializer(fn)` as a cleaner wrapper:
```typescript
// Angular's preferred pattern (v19+)
provideEnvironmentInitializer(() => { /* init logic */ })

// This PR's pattern
{ provide: ENVIRONMENT_INITIALIZER, multi: true, useValue() { ... } }
```

**Assessment:** Both work. The raw token approach is used because `inject()` must
be called inside `useValue()` (which runs in injection context), while
`provideEnvironmentInitializer` would need a separate closure. No functional
difference. Could migrate if desired but not required.

### 4. Signals vs Observables

Angular's direction is signal-first. This PR:

| Pattern | This PR | Angular's preference |
|---|---|---|
| Route query params | `toSignal(route.queryParamMap)` → `computed()` | Correct — bridges router Observable to signal |
| TanStack Query results | Signal-based (TanStack Angular returns signals) | Correct — aligns with Angular signals |
| HttpClient in queryFn | `lastValueFrom(http.get(...))` | Acceptable — HttpClient still returns Observable; `lastValueFrom` converts to Promise for TanStack |
| Page load data | `toSignal(injectLoadData<typeof load>())` | Could use `resource()` in future, but Observable + `toSignal` is current best practice for router data |

### 5. Branded/Phantom Types

Angular uses this pattern internally:
```typescript
// Angular's StateKey<T> — branded string with phantom type
type StateKey<T> = string & { __not_a_string: never; __value_type?: T };
```

This PR's `ServerRouteHandler`:
```typescript
interface ServerRouteHandler<TQuery, TBody, TResult> {
  (event: H3Event): Promise<Response>;
  readonly _types: { readonly query: TQuery; readonly body: TBody; readonly result: TResult };
}
```

**Assessment:** Follows the same phantom/branded type pattern Angular uses for
`StateKey`, `ɵproviders`, and internal type markers. The `_types` approach is
clean and idiomatic.

---

## File-Level Assessment

### Providers & DI

| File | Pattern | Idiomatic? |
|------|---------|------------|
| `provide-analog-query.ts` | `makeEnvironmentProviders` + `ENVIRONMENT_INITIALIZER` | Yes |
| `provide-server-analog-query.ts` | `makeEnvironmentProviders` + `BEFORE_APP_SERIALIZED` with factory + deps | Yes |
| `app.config.ts` | `provideFileRouter(withDebugRoutes(), withExtraRoutes(...))` | Yes — composition pattern |
| `app.config.server.ts` | `mergeApplicationConfig(appConfig, serverConfig)` | Yes |

### Components

| File | Standalone | OnPush | Signals | Control Flow | inject() | Assessment |
|------|-----------|--------|---------|-------------|----------|------------|
| `tanstack-query.page.ts` | Yes | Yes | Yes | `@if`/`@for` | Yes | Excellent |
| `tanstack-query-multi.page.ts` | Yes | Yes | Yes | `@if`/`@for` | Yes | Excellent |
| `tanstack-query-infinite.page.ts` | Yes | Yes | Yes | `@if`/`@for` | Yes | Excellent |
| `tanstack-query-optimistic.page.ts` | Yes | Yes | Yes | `@if`/`@for` | Yes | Excellent |
| `contact.page.ts` | Yes | Yes | Yes | `@if`/`@for` | Yes | Excellent |
| `greet.[name].page.ts` | Yes | Yes | Yes | N/A | Yes | Excellent |

### Directives

| File | Pattern | Assessment |
|------|---------|------------|
| `form-action.directive.ts` | `input()`, `output()`, `signal()`, standalone, host bindings | Exemplary modern directive |

---

## Opportunities for Future Alignment

These are NOT issues — they are areas where future Angular evolution could
inform the next iteration:

1. **`analogResource()` wrapper** — Bridge `defineServerRoute` branded types
   into Angular's `resource()` API for read-only data that doesn't need TanStack
   Query's caching/invalidation features. Would provide a signal-native
   alternative to `injectQuery + serverQueryOptions` for simpler cases.

2. **`httpResource` integration** — Angular's `httpResource` has built-in
   transfer cache support. A future `analogHttpResource<typeof route>()` could
   combine Angular's native SSR transfer cache with Analog's type-safe route
   inference, eliminating the need for manual `provideAnalogQuery` hydration for
   simple queries.

3. **`linkedSignal` for query state** — Angular's `linkedSignal` could model
   query state transitions (idle → loading → resolved → error) more explicitly
   than TanStack's internal signal management, if Analog ever builds its own
   query layer.

4. **`provideEnvironmentInitializer`** — Minor: could replace the raw
   `ENVIRONMENT_INITIALIZER` token usage with the wrapper function. No
   functional difference.

---

## Conclusion

This PR is well-aligned with modern Angular idioms. Every component, directive,
and provider follows Angular v19+ best practices. The branded type system for
server routes mirrors Angular's own phantom type patterns. The signal-based
consumption of TanStack Query results is idiomatic. The SSR hydration approach
correctly uses Angular's `TransferState` infrastructure.

The main architectural choice — TanStack Query over Angular's newer
`httpResource` — is justified by the feature requirements (cache invalidation,
mutations, optimistic updates, infinite queries) that `httpResource` doesn't
address. The two approaches can coexist, and the type-safe bridge utilities
(`serverQueryOptions`, etc.) make TanStack Query usage as ergonomic as possible
within Angular's ecosystem.

benpsnyder:
@coderabbitai resume

packages/router/query/ng-package.json
```diff
@@ -0,0 +1,6 @@
+{
```
brandonroberts: Why is this here?

packages/router/query/src/index.ts
```diff
@@ -0,0 +1,9 @@
+export * from '@tanstack/angular-query-experimental';
```
brandonroberts: Let's not do this. This is meant to be an integration, not a wrapper. We don't need to re-export thes APIs.

```suggestion
```

tsconfig.base.json
```diff
@@ -31,6 +31,10 @@
       "@analogjs/nx": ["./packages/nx-plugin/src/index.ts"],
       "@analogjs/platform": ["./packages/platform/src/index.ts"],
       "@analogjs/router": ["./packages/router/src/index.ts"],
+      "@analogjs/router/query": ["./packages/router/query/src/index.ts"],
+      "@analogjs/router/query/provide-server-analog-query": [
+        "./packages/router/query/src/provide-server-analog-query.ts"
+      ],
```
brandonroberts: Let's be more explicit with the integration

```suggestion
      "@analogjs/router/tanstack-query": ["./packages/router/query/src/index.ts"],
      "@analogjs/router/tanstack-query/server": [
        "./packages/router/query/src/provide-server-analog-query.ts"
      ],
```

packages/router/query/src/provide-analog-query.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export const ANALOG_QUERY_STATE_KEY: StateKey<DehydratedState> =
+  makeStateKey<DehydratedState>('analog_query_state');
+
+export function provideAnalogQuery(
+  queryClient: QueryClient | InjectionToken<QueryClient>,
+  ...features: QueryFeatures[]
+): EnvironmentProviders {
+  return makeEnvironmentProviders([
+    provideTanStackQuery(queryClient, ...features),
```
brandonroberts: This can be provided directly in the app.config

```suggestion
```

packages/router/query/src/provide-analog-query.ts
```diff
@@ -0,0 +1,52 @@
...
+  dehydrate,
+  hydrate,
+  provideTanStackQuery,
+} from '@tanstack/angular-query-experimental';
+
+export const ANALOG_QUERY_STATE_KEY: StateKey<DehydratedState> =
+  makeStateKey<DehydratedState>('analog_query_state');
+
+export function provideAnalogQuery(
+  queryClient: QueryClient | InjectionToken<QueryClient>,
```
brandonroberts: This can be provided directly in the app.config

```suggestion
```

packages/router/tanstack-query/src/server-query.ts
```diff
@@ -0,0 +1,146 @@
+import type { HttpClient } from '@angular/common/http';
```
brandonroberts, line 1: All this feels like it should be upstreamed to TanStack Query for Angular

apps/analog-app/index.html
```diff
@@ -1,11 +1,17 @@
...
     <title>AnalogApp</title>
     <base href="/" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
     <link rel="icon" type="image/x-icon" href="/favicon.ico" />
+    <link
+      href="https://cdn.jsdelivr.net/npm/daisyui@5"
+      rel="stylesheet"
+      type="text/css"
+    />
+    <script src="https://cdn.jsdelivr.net/npm/@tailwindcss/browser@4"></script>
```
brandonroberts: Use SCSS imports through node_modules instead of CDN links

```suggestion

```

packages/router/package.json
```diff
@@ -47,9 +55,20 @@
   "peerDependencies": {
     "@analogjs/content": "^3.0.0-alpha.9",
     "@angular/core": "^15.0.0 || ^16.0.0 || ^17.0.0 || ^18.0.0 || ^19.0.0 || ^20.0.0 || ^21.0.0",
-    "@angular/router": "^15.0.0 || ^16.0.0 || ^17.0.0 || ^18.0.0 || ^19.0.0 || ^20.0.0 || ^21.0.0"
+    "@angular/platform-server": "^15.0.0 || ^16.0.0 || ^17.0.0 || ^18.0.0 || ^19.0.0 || ^20.0.0 || ^21.0.0",
+    "@angular/router": "^15.0.0 || ^16.0.0 || ^17.0.0 || ^18.0.0 || ^19.0.0 || ^20.0.0 || ^21.0.0",
+    "@tanstack/angular-query-experimental": "^5.95.0"
```
brandonroberts: This needs to be less strict in case they are already using this package. What's the minimum version needed?

```suggestion
    "@tanstack/angular-query-experimental": ">=5.95.0"
```

packages/router/package.json
```diff
@@ -9,6 +9,14 @@
       "import": "./src/index.js",
       "default": "./src/index.js"
     },
+    "./query": {
```
brandonroberts:
```suggestion
    "./tanstack-query": {
```

packages/router/package.json
```diff
@@ -9,6 +9,14 @@
...
+      "import": "./query/src/index.js",
+      "default": "./query/src/index.js"
+    },
+    "./query/provide-server-analog-query": {
```
brandonroberts:
```suggestion
    "./tanstack-query/server": {
```

packages/router/README.md
```diff
@@ -1,5 +1,42 @@
-# Analog
+# Analog Router
 
-The fullstack meta-framework for Angular
+Filesystem-based routing for Angular with server-aware page loads, form actions,
```
brandonroberts: This should be in the documentation and this file left as-is.

apps/analog-app/index.html
```diff
@@ -1,11 +1,17 @@
...
     <link
```
brandonroberts, line 9: I don't like the look and feel of this app. Feels too AI-generated.

apps/docs-app/docs/features/data-fetching/server-side-data-fetching.md
```diff
@@ -124,3 +135,71 @@ VITE_ANALOG_PUBLIC_BASE_URL="http://localhost:5173"
 ```
 
 The environment variable must also be set when building for deployment.
+
+## Using TanStack Query with Analog
```
brandonroberts: This should go in the `Integrations` documentation with an entry for `TanStack Query`

packages/router/vite.config.lib.ts
```diff
@@ -23,6 +23,10 @@ export default defineConfig({
     lib: {
       entry: {
         'analogjs-router': resolve(import.meta.dirname, 'src/index.ts'),
+        'analogjs-router-query': resolve(
```
brandonroberts:
```suggestion
        'analogjs-router-tanstack-query': resolve(
```

packages/router/vite.config.lib.ts
```diff
@@ -23,6 +23,10 @@ export default defineConfig({
...
+          import.meta.dirname,
+          'query/src/index.ts',
```
brandonroberts:
```suggestion
          'tanstack-query/src/index.ts',
```

apps/analog-app/index.html
benpsnyder: Agreed -- will update

apps/analog-app/index.html
benpsnyder: I have done a lot of DaisyUI; this was just a temporary measure until I open a PR specifically on Tailwind v4

packages/router/query/ng-package.json
benpsnyder: This was added for ng-packagr to recognize the `query` directory as a secondary entry point during local development. But since this sub-path is consumed via TypeScript path aliases and package.json `exports` (not ng-packagr), it's not needed. I'll remove it.

packages/router/query/src/index.ts
benpsnyder: Agreed — removed. The re-export was from an earlier iteration. Consumers should import TanStack Query APIs directly from `@tanstack/angular-query-experimental`. The barrel will only export the Analog-specific bridge utilities (`serverQueryOptions`, `serverMutationOptions`, `serverInfiniteQueryOptions`, `provideAnalogQuery`).

tsconfig.base.json
benpsnyder: Makes sense — `tanstack-query` is more explicit about what the sub-path provides. I'll update the path aliases and all corresponding imports.

packages/router/query/src/provide-analog-query.ts
benpsnyder: Understood — I'll remove `provideAnalogQuery` as a wrapper. The hydration/dehydration logic can be wired directly in `app.config.ts` and `app.config.server.ts` using `provideTanStackQuery`, `TransferState`, and `BEFORE_APP_SERIALIZED` without an Analog-specific wrapper function. This keeps the integration transparent and avoids hiding TanStack's own provider setup behind an abstraction.

packages/router/query/src/provide-analog-query.ts
benpsnyder: Acknowledged — removing the wrapper. Users will configure TanStack Query directly in their app config.

packages/router/tanstack-query/src/server-query.ts
benpsnyder: I agree this would be valuable upstream. The `serverQueryOptions` / `serverMutationOptions` / `serverInfiniteQueryOptions` utilities are generic enough — they just bridge branded route types into TanStack Query option creators with `HttpClient`. The Analog-specific part is really just the `ServerRouteHandler` branded type and the `buildUrl` workaround for SSR internal fetch.

Would you prefer to: (a) keep these in Analog for now as a proving ground and upstream later once the API stabilizes, or (b) hold this part of the PR and open a TanStack Query RFC first? Happy to do either.

**EDIT**: I prefer (a) then lets open an issue for (b)

apps/analog-app/index.html
benpsnyder: Good call — I'll remove the CDN links for daisyUI and Tailwind CSS browser and set them up through local imports instead.

packages/router/package.json
benpsnyder: You're right — `>=5.95.0` is better to avoid conflicts with existing installations. 5.95.0 is the minimum because that's when `angular-query-experimental` stabilized the signal-based API surface we depend on (`injectQuery`, `injectMutation`, `injectInfiniteQuery` returning signals). I'll update to `>=5.95.0`. I should also make this an optional peer dependency since not all router users will use TanStack Query.

packages/router/package.json
benpsnyder: Agreed — renaming from `./query` to `./tanstack-query` to match the tsconfig path aliases. I'll update this.

packages/router/package.json
benpsnyder: Agreed — renaming from `./query/provide-server-analog-query` to `./tanstack-query/server`. I'll update this.

benpsnyder:
Already addressed — renamed to `defineServerRoute` across the board (function, types, tests, barrel exports) in the latest push. The new name scopes it as Analog's server-layer abstraction without implying it's limited to API routes or colliding with Nitro's native route conventions.

packages/router/tanstack-query/src/server-query.ts
brandonroberts: A

packages/router/query/ng-package.json
benpsnyder: Removed in previous commit — the `query/` directory has been renamed to `tanstack-query/` and `ng-package.json` was dropped since the sub-path is consumed via TypeScript path aliases and package.json exports, not ng-packagr.

packages/router/query/src/index.ts
benpsnyder: Done — TanStack Query re-exports were removed in a prior commit. The barrel now only exports Analog-specific bridge utilities (`ANALOG_QUERY_STATE_KEY`, `provideAnalogQuery`, `serverQueryOptions`, `serverMutationOptions`, `serverInfiniteQueryOptions`).

tsconfig.base.json
benpsnyder: Done — path aliases updated to `@analogjs/router/tanstack-query` and `@analogjs/router/tanstack-query/server` in a prior commit.

packages/router/query/src/provide-analog-query.ts
benpsnyder: Kept `provideAnalogQuery()` as a convenience wrapper for the `TransferState` rehydration. The hydration logic is wired up automatically so users only need one provider call in their app config.

packages/router/query/src/provide-analog-query.ts
benpsnyder: Same as above — keeping the wrapper for convenience. The inline pattern is shown in the new TanStack Query integration docs as an alternative.

apps/analog-app/index.html
benpsnyder: CDN links were already removed in a prior commit. The `index.html` now matches upstream.

packages/router/package.json
benpsnyder: Updated to `>=5.95.0` and marked as optional in `peerDependenciesMeta` in a prior commit.

packages/router/package.json
benpsnyder: Done — renamed to `./tanstack-query` in a prior commit.

packages/router/package.json
benpsnyder: Done — renamed to `./tanstack-query/server` in a prior commit.

packages/router/README.md
benpsnyder: Done — reverted `README.md` to upstream content. TanStack Query documentation moved to a dedicated integrations page at `docs/integrations/tanstack-query/index.md`.

apps/analog-app/index.html
benpsnyder: CDN styling removed and `index.html` restored to upstream. Will address the look and feel in a follow-up PR focused on Tailwind v4.

apps/docs-app/docs/features/data-fetching/server-side-data-fetching.md
benpsnyder: Done — moved TanStack Query documentation out of `server-side-data-fetching.md` and into a new `docs/integrations/tanstack-query/index.md` page in this commit (640ce12e).

packages/router/vite.config.lib.ts
benpsnyder: Done — renamed to `analogjs-router-tanstack-query` in a prior commit.

packages/router/vite.config.lib.ts
benpsnyder: Done — updated to `tanstack-query/src/index.ts` in a prior commit.

packages/router/tanstack-query/src/server-query.ts
benpsnyder: Confirmed — keeping in Analog as a proving ground (option A). Will open an issue to track upstreaming to TanStack Query once the API stabilizes.

benpsnyder:
Restored `analog-app`
image
Scaffolded functional `tanstack-query-app`
image
UI overhaul will be done on follow-up so it doesn't look AI generated

PR #2150, opened 2026-03-21
feat(platform): simplify deps-plugin and fix shiki highlighter init

benpsnyder:
## Affected scope

- Primary scope: platform
- Secondary scopes: none

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

### deps-plugin simplification

The deps-plugin previously used a runtime `rolldownVersion` check to decide whether to return an `oxc` or `esbuild` transform config key. Since Vite 8+ ships Rolldown by default and the monorepo no longer supports Vite < 8 for development, the esbuild branch was dead code. This PR:

- Removes the `import * as vite` namespace import and the `rolldownVersion` conditional
- Always returns the `oxc` transform config key via the existing `getJsTransformConfigKey()` utility
- Consolidates `useAngularCompilationAPI` detection to check both `options.experimental` and `options.vite.experimental` paths, fixing a case where the top-level experimental option was ignored
- Removes the `vi.mock('vite')` and esbuild-specific test branches, cutting test complexity in half (10 tests → 4 tests, same coverage)

### shiki-highlighter fix

Moves the `getHighlighter()` call from a class field initializer into the constructor body. Field initializers run before constructor parameters are assigned in the prototype chain, so `this.highlighterOptions` was `undefined` when `getHighlighter` was called. Moving it to the constructor body ensures the options are available.

## Test plan

- [x] `nx test platform` — all 19 tests pass
- [x] `nx build platform`

- [ ] Yes
- [x] No

### DIFF
--- a/packages/content-plugin/eslint.config.mjs
+++ b/packages/content-plugin/eslint.config.mjs
@@ -23,7 +23,10 @@ export default [
   {
     files: ['**/*.json'],
     rules: {
-      '@nx/dependency-checks': 'error',
+      '@nx/dependency-checks': [
+        'error',
+        { ignoredFiles: ['**/vite.config.lib.ts'] },
+      ],
     },
     languageOptions: {
       parser: await import('jsonc-eslint-parser'),
--- a/packages/content-plugin/src/migrations/update-markdown-version/update-markdown-version.ts
+++ b/packages/content-plugin/src/migrations/update-markdown-version/update-markdown-version.ts
@@ -1,6 +1,6 @@
 import {
   addDependenciesToPackageJson,
-  Tree,
+  type Tree,
   formatFiles,
   installPackagesTask,
 } from '@nx/devkit';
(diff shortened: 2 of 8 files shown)


brandonroberts (approved):
Approved

benpsnyder:
> Approved

Ready to merge

The reason for "Removes the import * as vite namespace import and the rolldownVersion conditional": Vite 8+ ships Rolldown by default, so the runtime check (rolldownVersion conditional) is no longer needed. The code
   always uses the oxc transform key now instead of branching between oxc (Rolldown/Vite 8+) and esbuild (older Vite).                                           
                                                                                                                                                                 
Concretely:                                                                                                                                                    
  - The import * as vite from 'vite' namespace import existed solely to access vite.rolldownVersion at runtime.                                                  
  - The rolldownVersion conditional decided whether to use the oxc or esbuild config key.                                                                        
  - Since Analog now targets Vite 8+ (which always has Rolldown), that branch is dead code — it will always be oxc. So both the namespace import and the conditional were removed to simplify the plugin.

PR #2152, opened 2026-03-21
fix(vite-plugin-angular): forward OXC compilation API and fix transform lang

benpsnyder:
## Affected scope

- Secondary scopes: none

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

## What is the new behavior?

Three targeted fixes for the Angular Vite plugin's OXC/Rolldown integration:

### 1. JavaScriptTransformer resource leak fix

`createRolldownCompilerPlugin` now accepts a `closeTransformer` parameter and calls `javascriptTransformer.close()` in its `buildEnd` hook when enabled. This is passed as `true` for non-Astro builds (where the plugin controls the transformer lifecycle) and `false` for Astro integration (where the transformer is shared across multiple builds).

Without this, the dep optimizer's Rolldown plugin keeps the Angular compiler's worker pool alive after optimization completes, leaking file handles and memory in long-running dev sessions.

### 2. Transform language correction (ts, not js)

`angularVitestSourcemapPlugin` was passing `lang: "js"` / `loader: "js"` when transforming `.ts` files through OXC and esbuild respectively. This worked by accident in most cases because both tools can infer the language from the file extension, but it fails for files with TypeScript-only syntax (type annotations, enums, etc.) when the extension is stripped or aliased. Changed to `lang: "ts"` / `loader: "ts"` to match the actual source language.

### 3. OXC ESM sourceType for Angular CDK

Added `sourceType: "module"` to the OXC transform call that handles `@angular/cdk` and related packages. Unlike esbuild (which defaults to ESM when it sees `import`/`export` statements), OXC requires explicit `sourceType` to guarantee ESM output. Without this, OXC may produce CommonJS-style output for some CDK modules, causing runtime import errors in the browser.

### Changes

| File | Change |
|---|---|
| `compiler-plugin.ts` | Added `closeTransformer` param, `buildEnd` hook calling `javascriptTransformer.close()` |
| `angular-vite-plugin.ts` | Pass `!isAstroIntegration` as `closeTransformer` to `createRolldownCompilerPlugin` |
| `angular-vitest-plugin.ts` | `lang: "ts"` / `loader: "ts"` fix, added `sourceType: "module"` for OXC CDK transform |
| `angular-vitest-plugin-oxc.spec.ts` | Updated expectations to match `lang: "ts"` / `loader: "ts"` |

## Test plan

- [x] `nx test vite-plugin-angular` — all 66 tests pass

- [ ] Yes
- [x] No

### DIFF
--- a/packages/nx-plugin/src/generators/app/generator.ts
+++ b/packages/nx-plugin/src/generators/app/generator.ts
@@ -67,7 +67,7 @@ function normalizeOptions(
   };
 }
 
-export async function appGenerator(
+async function appGenerator(
   tree: Tree,
   options: AnalogNxApplicationGeneratorOptions,
 ): Promise<() => void> {
--- a/packages/nx-plugin/src/generators/init/generator.ts
+++ b/packages/nx-plugin/src/generators/init/generator.ts
@@ -60,7 +60,7 @@ function addFiles(
   }
 }
 
-export async function setupAnalogGenerator(
+async function setupAnalogGenerator(
   tree: Tree,
   options: SetupAnalogGeneratorSchema,
 ): Promise<() => void> {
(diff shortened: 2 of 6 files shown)


benpsnyder:
> Approved

Ready to merge

PR #2154, opened 2026-03-21
fix(content-plugin): use type-only Tree import and update eslint config

benpsnyder:
## Affected scope

- Primary scope: content-plugin
- Secondary scopes: none

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

### Type-only import fix

Changes `Tree` from a value import to a type-only import (`import type`) in the `update-markdown-version` migration. `Tree` is only used as a parameter type annotation — importing it as a value causes the CJS output to emit a runtime `require('@nx/devkit').Tree` that fails when `Tree` is not re-exported as a value from the CommonJS build of `@nx/devkit`.

### ESLint config update

Adds `vite.config.lib.ts` and `tsdown.config.ts` to the `@nx/dependency-checks` ignore list. These are build-time config files that import dev tools (`oxc-transform`, `tsdown`, etc.) which are workspace root dependencies, not package-level dependencies. Without ignoring them, the lint rule falsely flags missing dependencies.

## Test plan

- [x] `nx build content-plugin`
- [x] `nx lint content-plugin`

- [ ] Yes
- [x] No

### DIFF
--- a/packages/nx-plugin/src/generators/app/generator.ts
+++ b/packages/nx-plugin/src/generators/app/generator.ts
@@ -67,7 +67,7 @@ function normalizeOptions(
   };
 }
 
-export async function appGenerator(
+async function appGenerator(
   tree: Tree,
   options: AnalogNxApplicationGeneratorOptions,
 ): Promise<() => void> {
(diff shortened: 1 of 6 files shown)


brandonroberts (approved):
Approved

PR #2155, opened 2026-03-21
chore(vitest-angular): esm-compatible require for builders

benpsnyder:
## Affected scope

- Primary scope: vitest-angular
- Secondary scopes: vitest-angular-tools

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

### ESM-compatible require() in builders

Replaces bare `require()` calls with `createRequire(import.meta.url)` in `devkit.ts` and `vitest.impl.ts`. Vite 8 resolves builder entry points as ESM by default, and bare `require()` is not available in ESM modules. `createRequire` provides a CJS-compatible `require` function that works in both ESM and CJS contexts, maintaining the dynamic import behavior these builders need for loading `@angular-devkit/build-angular` and `@angular/build/private` at runtime.

### Test collection path update

Updates the `vitest-angular-tools` test to resolve `collection.json` from the new `node_modules/@analogjs/vitest-angular/src/lib/tools/` output location (was `dist/packages/vitest-angular-tools/`).

## Test plan

- [x] `nx test vitest-angular-tools` — all 16 tests pass
- [x] `nx build vitest-angular`

- [ ] Yes
- [x] No


brandonroberts:
Not a fan of removing the idiomatic Angular decorators for this

benpsnyder:
> Not a fan of removing the idiomatic Angular decorators for this

My mistake. Reverted that.

packages/vitest-angular/src/lib/builders/build/vitest.impl.ts
```diff
@@ -194,7 +195,8 @@ function findIncludes(options: {
   include: string[];
   exclude: string[];
 }) {
-  const { normalizePath } = require('vite');
+  const _require = createRequire(import.meta.url);
```
brandonroberts: This usage is inconsistent with the other usage, where _require is created in the module scope.

benpsnyder: Good catch — moved `_require` to module scope to match `devkit.ts`. Fixed.

PR #2162, opened 2026-03-22
Improve astro rendering: move component styles to document head

eblocha:
## PR Checklist

This changes the Astro server renderer to produce valid HTML documents.

Closes https://github.com/analogjs/analog/issues/2207

Today, `@analogjs/astro-angular` renders components like so:

```html
<html>
    <head>
    </head>
    <body>
        <astro-island>
            <html>
                <head>
                    <style><!-- component styles --></style>
                </head>
                <body>
                    <ng-component>
                    </ng-component>
                </body>
            <html>
        </astro-island>
    </body>
</html>
```

This change will render the component as:

```html
<html>
    <head>
    </head>
    <body>
        <astro-island>
            <style><!-- component styles --></style>
            <ng-component>
            </ng-component>
        </astro-island>
    </body>
</html>
```

Additionally, it includes an optional `strictStylePlacement` flag to move the style tag to the head (disables SSR streaming).

```html
<html>
    <head>
        <style><!-- component styles --></style>
    </head>
    <body>
        <astro-island>
            <ng-component>
            </ng-component>
        </astro-island>
    </body>
</html>
```

## Affected scope

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

## What is the new behavior?

See above

## Test plan

- [x] `nx format:check`
- [x] `nx test astro-angular`
- [x] Manual verification: built the astro-app with the new strictStylePlacement option, verified the output was valid html.

- [ ] Yes
- [x] No

### DIFF
--- a/packages/astro-angular/src/index.ts
+++ b/packages/astro-angular/src/index.ts
@@ -1,3 +1,4 @@
+import { fileURLToPath } from 'node:url';
 import viteAngular, { PluginOptions } from '@analogjs/vite-plugin-angular';
 import { enableProdMode } from '@angular/core';
 import type { AstroIntegration, AstroRenderer, ViteUserConfig } from 'astro';
@@ -94,7 +95,7 @@ export default function (options?: AngularOptions): AstroIntegration {
         if (options?.strictStylePlacement) {
           addMiddleware({
             order: 'pre',
-            entrypoint: '@analogjs/astro-angular/middleware.js',
+            entrypoint: fileURLToPath(import.meta.resolve('./middleware.js')),
           });
         }
       },
--- a/packages/astro-angular/package.json
+++ b/packages/astro-angular/package.json
@@ -9,7 +9,6 @@
     "./utils": "./src/utils.js",
     "./client.js": "./src/client.js",
     "./server.js": "./src/server.js",
-    "./middleware.js": "./src/middleware.js",
     "./package.json": "./package.json"
   },
   "keywords": [


eblocha:
@brandonroberts I'd like to guage your interest in this change. It's inspired by the discussion with Astro here: https://github.com/withastro/roadmap/discussions/1322 and this library I found: https://inox-tools.fryuni.dev/portal-gun

I'm definitely open to changing the new option, or making on by default. I left it off to avoid changing the default behavior, which currently allows SSR streaming.

I also opened this issue with Angular to see if it's possible to use Angular's client hydration in Astro: https://github.com/angular/angular/issues/67785

brandonroberts:
@eblocha I've asked about this change previously also. I think its good to add it behind a flag. Please open an issue to associate with the PR that describes the original issue similar to what you did in the Angular repo.

brandonroberts:
Thank @eblocha. One question on the middleware: right now it moves **all** `<style>` tags found in the body to the `<head>`, not just the ones Angular generated. Could this be scoped to only target Angular-rendered style tags? For example, Angular's emulated encapsulation produces styles with `_ngcontent-`/`_nghost-` attributes — or the styles could be identified by proximity to `<astro-island>` elements. This would avoid accidentally hoisting styles that other integrations or user code intentionally placed in the body.

packages/astro-angular/src/index.ts
```diff
@@ -77,13 +84,19 @@ export default function (options?: AngularOptions): AstroIntegration {
...
         addRenderer(getRenderer());
         updateConfig({
           vite: getViteConfiguration(
             options?.vite,
           ) as unknown as ViteUserConfig,
         });
+        if (options?.strictStylePlacement) {
+          addMiddleware({
+            order: 'pre',
+            entrypoint: '@analogjs/astro-angular/middleware.js',
```
brandonroberts: Does this entrypoint need to be public API or can it point to the middleware file relative to this index.js file?

eblocha: I was able to remove it from the public API by using `import.meta.resolve`.

eblocha:
> Thank @eblocha. One question on the middleware: right now it moves **all** `<style>` tags found in the body to the `<head>`, not just the ones Angular generated. Could this be scoped to only target Angular-rendered style tags? For example, Angular's emulated encapsulation produces styles with `_ngcontent-`/`_nghost-` attributes — or the styles could be identified by proximity to `<astro-island>` elements. This would avoid accidentally hoisting styles that other integrations or user code intentionally placed in the body.

@brandonroberts I added a check to only move style tags with an `ng-app-id` attribute. I also added a component to the astro-app demo to verify this works for nested components, not just the bootstrap component.

packages/astro-angular/README.md
```diff
@@ -173,6 +173,24 @@ export default defineConfig({
...
+
+```js
+import { defineConfig } from 'astro/config';
+
+import angular from '@analogjs/astro-angular';
+
+export default defineConfig({
+  integrations: [angular({ strictStylePlacement: true })],
+});
+```
```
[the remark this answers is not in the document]
eblocha: I didn't consider this to be breaking.

- The previous behavior produced an entire html document in the body, which is likely to cause issues in browsers and assistive technology.
- This PR changes the default behavior to unwrap this document into style tags + component content, which is less likely to cause issues.
- This new option _would_ be a breaking change if enabled by default, but it is off by default.

packages/astro-angular/src/middleware.spec.ts
```diff
@@ -0,0 +1,200 @@
...
+      `
+      <!DOCTYPE html>
+      <html>
+        <head>
+        </head>
+        <body>
+          <astro-island>
+            <!-- comment -->
+            <style ng-app-id="ng">.card{color:red}</style>
+            <style ng-app-id="ng">.card{background:blue}</style>
```
eblocha, line 16: I put 2 style tags here to verify order is preserved.

brandonroberts (dismissed):

brandonroberts:
Thanks @eblocha!

PR #2181, opened 2026-03-25
fix(platform): enhance route manifest generation to handle pathless layouts and prevent collisions

### LINKED ISSUE #2174: bug(router): pathless parent layouts are treated as route collisions during typed route generation
## Environment

- Reproducible in the current typed-routes work for Analog v3 / PR #2125
- Package area: `platform` — `generateRouteManifest()` in `packages/platform/src/lib/route-manifest.ts`
- OS: macOS 25.3.0
- Node: `v25.5.0`
- pnpm: `10.32.0`

## Description

Pathless parent layouts are currently treated as duplicate routes when the typed route manifest is generated.

Today, `generateRouteManifest()` detects collisions strictly by `fullPath` before it computes any parent-child relationships. That works for true duplicate routes, but it produces a false positive for pathless layouts because files like `(auth).page.ts` and `(home).page.ts` both normalize to `"/"` even though they are structural parent routes, not competing leaf routes.

This means the current implementation does not match Analog's documented pathless layout behavior. Semantically, these files should coexist as layout parents with distinct structural identities, but one of them is warned on and skipped.

## Small reproduction

Create a page structure like this:

```text
src/app/pages/
├── (auth).page.ts
├── (auth)/login.page.ts
├── (home).page.ts
└── index.page.ts
```

Both `(auth).page.ts` and `(home).page.ts` normalize to `/` at `filenameToRoutePath()` time.

The current manifest logic in `packages/platform/src/lib/route-manifest.ts` then warns and drops one of them:

```ts
if (seenByFullPath.has(fullPath)) {
  const winningFilename = seenByFullPath.get(fullPath);
  console.warn(
    `[Analog] Route collision: '${fullPath}' is defined by both ` +
      `'${winningFilename}' and '${filename}'. ` +
      `Keeping '${winningFilename}' based on route source precedence and skipping duplicate.`,
  );
  continue;
}
```

A closely related expectation is also encoded in the current test suite (`packages/platform/src/lib/route-manifest.spec.ts`):

```ts
it('should warn on route collisions and skip duplicates', () => {
  const manifest = generateRouteManifest([
    '/app/routes/index.ts',
    '/app/routes/(home).ts',
  ]);

  expect(spy).toHaveBeenCalledWith(
    expect.stringContaining('Route collision'),
  );
  expect(manifest.routes.filter((r) => r.fullPath === '/').length).toBe(1);
});
```

## Problem statement

The manifest builder already distinguishes structural identity (`id`) from navigation identity (`fullPath`), which is the right model for pathless/layout routes. The bug is that collision detection currently happens too early and uses only `fullPath` as the uniqueness key.

As a result:

- valid pathless parent layouts are reported as collisions
- one of the layouts is removed from the manifest
- typed route generation is inconsistent with the documented layout-route model
- current warnings make it look like user code is invalid when the structure is actually supported by Analog

## Expected behavior

Pathless parent layouts should be preserved as distinct structural routes even when they resolve to the same navigation path.

A warning should only be emitted for true competing routes, not for layout/group parents that intentionally share a resolved `fullPath`.

## Proposed solution

Treat structural route identity and navigation path identity separately inside `generateRouteManifest()`:

1. Keep `id` / filename-derived structural identity as the primary manifest identity.
2. Allow multiple manifest entries to share the same `fullPath` when they are pathless layout/group parents.
3. Move or refine collision detection so it runs with enough structural context to detect only true route conflicts.
4. Continue choosing a deterministic canonical route for any APIs that must stay unique by `fullPath`, but do not drop the structural layout entries from the manifest.
5. Add regression tests for pathless-layout cases such as:
   - `(auth).page.ts` + `(auth)/login.page.ts`
   - `(auth).page.ts` + `(home).page.ts`
   - `index.page.ts` coexisting with one or more pathless parent layouts

One possible implementation approach would be:

- build manifest entries first using `id` as the guaranteed-unique key
- derive parent-child/layout relationships
- perform collision checks only for entries that truly compete for the same resolved route slot
- keep all layout/group nodes in `byId` metadata even if `byFullPath` remains canonicalized for navigation helpers

## Why this matters

Pathless parent layouts are a documented routing feature in Analog, so the current warning is confusing for users and makes valid route structures look broken during dev/build. Fixing this would align typed route generation with the existing runtime/layout model and remove a noisy false positive.

## Other information

Relevant docs describe pathless layout routes as valid and supported:
- https://analogjs.org/docs/features/routing/overview#pathless-layout-routes

benpsnyder:
## PR Checklist

Closes #2174

## Affected scope

- Primary scope: platform
- Secondary scopes: router (typed route generation consumers)

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

## What is the new behavior?

Pathless layout files (e.g. `(auth).page.ts`, `(home).page.ts`) are no longer treated as route collisions when they share the same resolved `fullPath`. Previously, `generateRouteManifest()` used `fullPath` as the sole uniqueness key during collision detection, which caused valid pathless layouts to be warned on and dropped from the manifest.

Changes in `packages/platform/src/lib/route-manifest.ts`:

- **`isPathlessLayoutId()`** — detects routes whose id leaf segment is a group pattern (e.g. `/(auth)`, `/(home)`), identifying structural layout parents rather than competing navigation targets.
- **Collision detection** — pathless layouts skip the `seenByFullPath` check, so multiple group layouts and a non-group route can coexist at the same `fullPath`. True collisions (two non-group files at the same path) still warn and deduplicate.
- **`canonicalRoutesByFullPath()`** — for generated declarations that must be unique by `fullPath` (`AnalogRouteTable`, `AnalogFileRoutesByFullPath`, `byFullPath`), selects a canonical entry preferring non-group routes over group layouts.
- **`isGroup` flag narrowed** — previously `id.includes('/(')` matched any route nested under a group (e.g. `/(auth)/sign-up` was incorrectly `isGroup: true`). Now uses `isPathlessLayoutId()` so only actual pathless layout routes are flagged.
- **`findNearestParentById()`** — new structural parent lookup that wires group children (e.g. `/(auth)/sign-up`) to their pathless layout parent (`/(auth)`) via id-based hierarchy, falling back to the existing `fullPath`-based lookup for non-group routes.
- **`hasJsonLd` accuracy** — renamed `jsonLdPaths` → `jsonLdFiles` and matches against `route.filename` instead of `route.fullPath`. Previously, when multiple routes shared a `fullPath` (e.g. pathless layouts at `/`), all got `hasJsonLd: true` even if only one exported JSON-LD.

## Test plan

- [x] `pnpm nx test platform` — 177 tests pass
- [x] Existing collision test updated to use a true collision case (duplicate non-group files from different source dirs)
- [x] New test: `should preserve pathless layouts that share the same fullPath` — verifies `(auth).page.ts`, `(home).page.ts`, and `index.page.ts` all coexist without collision warnings
- [x] New test: `should preserve pathless layout with its nested children` — verifies `(auth).page.ts` and `(auth)/login.page.ts` are both present with correct parent-child relationship
- [x] `pnpm nx build analog-app` — verified generated `routeTree.gen.ts` output:
  - `/(auth)` has `children: ["/(auth)/sign-up"]`, `hasJsonLd: false`
  - `/(auth)/sign-up` has `parentId: "/(auth)"`, `isGroup: false`
  - `/(home)` has `hasJsonLd: true` (the only layout exporting JSON-LD)

- [ ] Yes
- [x] No

## Other information

Cross-ref: the issue documents that this aligns with Analog's [pathless layout route docs](https://analogjs.org/docs/features/routing/overview#pathless-layout-routes). The `byId` metadata preserves all layout/group nodes while `byFullPath` remains canonicalized for navigation helpers.

The `jsonLdPaths` → `jsonLdFiles` rename is an internal API change between `typed-routes-plugin.ts` and `route-manifest.ts` — no public API surface is affected.

### DIFF
--- a/apps/blog-app/src/app/pages/blog/index.page.ts
+++ b/apps/blog-app/src/app/pages/blog/index.page.ts
@@ -2,12 +2,9 @@ import { isPlatformBrowser } from '@angular/common';
 import { Component, inject, OnInit, PLATFORM_ID } from '@angular/core';
 import { Router, RouterLink } from '@angular/router';
 import { contentFilesResource } from '@analogjs/content/resources';
-import { analogRouteTree } from '../../../routeTree.gen';
 import { routePath } from '@analogjs/router';
 
 import { PostAttributes } from './models';
-
-const blogPostRoute = analogRouteTree.byId['/blog/[slug]'];
 
 @Component({
   standalone: true,
@@ -19,7 +16,7 @@ const blogPostRoute = analogRouteTree.byId['/blog/[slug]'];
       @for (post of contentFilesResource.value(); track post.slug) {
         <li>
           @let postLink =
-            routePath(blogPostRoute.fullPath, {
+            routePath('/blog/[slug]', {
               params: { slug: post.slug },
             });
           <a [routerLink]="postLink.path"> {{ post.attributes.title }}</a>
@@ -30,7 +27,6 @@ const blogPostRoute = analogRouteTree.byId['/blog/[slug]'];
 })
 export default class BlogComponent implements OnInit {
   readonly routePath = routePath;
-  readonly blogPostRoute = blogPostRoute;
   private readonly platformId = inject(PLATFORM_ID);
   private readonly router = inject(Router);
   readonly contentFilesResource = contentFilesResource<PostAttributes>(


brandonroberts:
## `isPathlessLayoutId` regex matches empty group `()`

```typescript
function isPathlessLayoutId(id: string): boolean {
  const segments = id.split('/').filter(Boolean);
  if (segments.length === 0) return false;
  return /^\([^.[\]]*\)$/.test(segments[segments.length - 1]);
}
```

The `*` quantifier means an empty group name `()` matches — a file literally named `().page.ts` would be treated as a valid pathless layout and skip collision detection. Should be `+` instead of `*`:

```typescript
return /^\([^.[\]]+\)$/.test(segments[segments.length - 1]);
```

brandonroberts:
## `findNearestParentById` takes priority for all routes, not just group children

```typescript
const structuralParent = findNearestParentById(route.id, routeById);
const fullPathParent = findNearestParentRoute(route.fullPath, routeByFullPath);
const parent = structuralParent ?? fullPathParent;
```

The comment says "this correctly wires group children (e.g. `/(auth)/sign-up`) to their pathless layout parent" — but the code applies the id-based lookup unconditionally for every route, not just group children.

For non-group routes this may accidentally prefer an id-based parent over the intended fullPath-based parent if the id segments happen to match a different route than the canonical fullPath lookup would find. Either scope `findNearestParentById` to only run when the route's id contains a group segment, or add a test demonstrating that the unconditional behavior is safe for non-group routes with nested paths.

brandonroberts:
## No test for "only pathless layouts, no index" at a given fullPath

`canonicalRoutesByFullPath` picks the canonical route for `AnalogRouteTable` and `byFullPath`:

```typescript
if (!existing || (existing.isGroup && !route.isGroup)) {
  map.set(route.fullPath, route);
}
```

If **all** routes at a given `fullPath` are pathless layouts and there's no `index.page.ts`, the first one in iteration order wins. That order depends on the prior sort, which is by route weight then `fullPath` — and since all pathless layouts sharing a `fullPath` have the same weight and `fullPath`, the tiebreaker is `fullPath.localeCompare(fullPath)` which is always `0`, making the winner depend on sort stability.

The existing tests only cover cases where an `index.page.ts` exists alongside the layouts. Adding a test for the "only layouts, no index" case would either confirm the intended behavior or reveal that an explicit tiebreaker is needed (e.g., alphabetical by id or filename).

benpsnyder:
> ## `isPathlessLayoutId` regex matches empty group `()`
> ```ts
> function isPathlessLayoutId(id: string): boolean {
>   const segments = id.split('/').filter(Boolean);
>   if (segments.length === 0) return false;
>   return /^\([^.[\]]*\)$/.test(segments[segments.length - 1]);
> }
> ```
> 
> The `*` quantifier means an empty group name `()` matches — a file literally named `().page.ts` would be treated as a valid pathless layout and skip collision detection. Should be `+` instead of `*`:
> 
> ```ts
> return /^\([^.[\]]+\)$/.test(segments[segments.length - 1]);
> ```

isPathlessLayoutId regex (route-manifest.ts:310): Changed * to + so an empty group () is no longer treated as a valid pathless layout.

benpsnyder:
> ## `findNearestParentById` takes priority for all routes, not just group children
> ```ts
> const structuralParent = findNearestParentById(route.id, routeById);
> const fullPathParent = findNearestParentRoute(route.fullPath, routeByFullPath);
> const parent = structuralParent ?? fullPathParent;
> ```
> 
> The comment says "this correctly wires group children (e.g. `/(auth)/sign-up`) to their pathless layout parent" — but the code applies the id-based lookup unconditionally for every route, not just group children.
> 
> For non-group routes this may accidentally prefer an id-based parent over the intended fullPath-based parent if the id segments happen to match a different route than the canonical fullPath lookup would find. Either scope `findNearestParentById` to only run when the route's id contains a group segment, or add a test demonstrating that the unconditional behavior is safe for non-group routes with nested paths.

findNearestParentById scoped to group routes (route-manifest.ts:245-252): The structural id-based parent lookup now only runs when the route's id contains a group segment (/( pattern). Non-group routes always use the canonical fullPath-based lookup, preventing accidental id-based parent mismatches.

benpsnyder:
> ## No test for "only pathless layouts, no index" at a given fullPath
> `canonicalRoutesByFullPath` picks the canonical route for `AnalogRouteTable` and `byFullPath`:
> 
> ```ts
> if (!existing || (existing.isGroup && !route.isGroup)) {
>   map.set(route.fullPath, route);
> }
> ```
> 
> If **all** routes at a given `fullPath` are pathless layouts and there's no `index.page.ts`, the first one in iteration order wins. That order depends on the prior sort, which is by route weight then `fullPath` — and since all pathless layouts sharing a `fullPath` have the same weight and `fullPath`, the tiebreaker is `fullPath.localeCompare(fullPath)` which is always `0`, making the winner depend on sort stability.
> 
> The existing tests only cover cases where an `index.page.ts` exists alongside the layouts. Adding a test for the "only layouts, no index" case would either confirm the intended behavior or reveal that an explicit tiebreaker is needed (e.g., alphabetical by id or filename).

New test for "only layouts, no index" (route-manifest.spec.ts): Added a test confirming that when only pathless layouts exist at a given fullPath (no index.page.ts), the canonical selection for AnalogRouteTable and byFullPath picks exactly one entry deterministically.

brandonroberts:
**Pitfall 1: Non-deterministic canonical selection when all routes are groups**

`canonicalRoutesByFullPath()` prefers non-group over group, but when multiple group layouts share a `fullPath` and there's no non-group route, the "winner" is whichever appears first in the `routes` array. The test asserts determinism but relies on `routes` being sorted by `fullPath` — and since all candidates have the *same* `fullPath`, `localeCompare` doesn't guarantee a stable order between them. If the filesystem or glob returns files in a different order across platforms/runs, the canonical pick could flip, causing noisy diffs in `routeTree.gen.ts`.

**Suggested fix:** Add a tiebreaker to `canonicalRoutesByFullPath` (e.g., compare by `id` alphabetically).

brandonroberts:
**Pitfall 2: Parent wiring heuristic is broader than `isGroup`**

The parent lookup decides the strategy based on `route.id.includes('/(')` — the same broad check this PR explicitly narrowed for `isGroup`. A route like `/(auth)/sign-up` correctly gets `findNearestParentById`, but a deeply nested route like `/dashboard/(settings)/profile` would also trigger the id-based lookup. If `/(settings)` doesn't exist as a route (e.g. no `(settings).page.ts` file), `findNearestParentById` returns `undefined` and the fallback to `fullPathParent` kicks in — which is fine. But if there *is* a `/dashboard` route, the id-based lookup would walk up to `/dashboard` anyway, potentially skipping the fullPath-based parent that would have been more semantically correct.

brandonroberts:
**Pitfall 3: `canonicalRoutesByFullPath` is called multiple times redundantly**

It's called once in `generateRouteManifest`, then again in `generateRouteTableDeclaration`, and twice in `generateRouteTreeDeclaration`. This is a minor efficiency concern but also a consistency risk — if the route list is mutated between calls, different parts of the generated output could disagree on the canonical route.

brandonroberts:
**Pitfall 4: Collision between a pathless layout and `index.page.ts` is silently allowed**

If a user has both `(auth).page.ts` and `index.page.ts` at the same level, both resolve to `/` and both are kept. The test verifies this as expected behavior, but it's worth considering whether the Angular router can actually handle two competing components for the same path. Pathless layouts in frameworks like Remix/TanStack are *wrappers* (they render an `<Outlet>`), but in Analog's Angular routing, these are `*.page.ts` files that are full page components. If the Angular router receives multiple route configs for `/`, behavior depends on registration order.

brandonroberts:
**Pitfall 5: `isPathlessLayoutId` regex is narrow**

The regex `^\([^.[\]]+\)$` excludes dots and brackets — but doesn't handle edge cases like `(auth.v2)` with dots (which would NOT be recognized as a pathless layout). This may be intentional but isn't documented.

brandonroberts:
**Pitfall 6: `jsonLdFiles` matching depends on exact filename normalization**

The switch from `fullPath` to `filename` matching is correct for disambiguation, but `filename` values come from the raw file list input (e.g. `/src/app/pages/(home).page.ts`). If the `jsonLdEntries` in `typed-routes-plugin.ts` produces `sourceFile` values with slightly different normalization (trailing slashes, relative vs absolute), the match will silently fail. The old `fullPath` approach was more forgiving since paths were normalized during route generation.

benpsnyder:
## Addressing review pitfalls

| Pitfall | Action |
|---------|--------|
| **1**: Non-deterministic canonical selection | Added `id.localeCompare()` tiebreaker in `canonicalRoutesByFullPath` when both candidates are groups. Updated test to prove input-order independence. |
| **2**: Parent wiring heuristic broader than `isGroup` | Expanded comment explaining why the unconditional `includes('/(')` is safe (walks up to group parent or falls through to fullPath). Added test for nested group case (`/dashboard/(settings)/profile`). |
| **3**: Redundant `canonicalRoutesByFullPath` calls | Computed once in `generateRouteManifest`, stored on `RouteManifest.canonicalByFullPath`, reused in `generateRouteTableDeclaration` and `generateRouteTreeDeclaration`. |
| **4**: Pathless layout + index.page.ts coexistence | Added clarifying comment that pathless layouts are structural wrappers with `<router-outlet>`, not competing page components. |
| **5**: `isPathlessLayoutId` regex excludes dots | Added comment documenting that `(auth.v2)` is intentionally excluded — dot-containing names collide with dynamic-segment syntax. |
| **6**: jsonLdFiles normalization | Verified safe — both `route.filename` and `entry.sourceFile` originate from the same `discovery` source with no intermediate normalization. No code change needed. |

brandonroberts:
General note, there aren't any examples in the analog-app/blog-app that actually use the generated type-safe routes to ensure the type safety is working.

benpsnyder:
> General note, there aren't any examples in the analog-app/blog-app that actually use the generated type-safe routes to ensure the type safety is working.

image
image
image

Also 9aee409edd210220aa7e437aa6de67b50848ac8a needs tobe called out; might need a cherry-pick

PR #2193, opened 2026-03-27
fix(astro-angular): move vite-plugin-angular to peerDependencies

benpsnyder:
## Summary
- Moves `@analogjs/vite-plugin-angular` from `dependencies` to `peerDependencies` in `astro-angular` with a broad range (`^3.0.0-alpha.0`)
- Removes `packages/astro-angular/package.json` from `replacementFiles` in `release.config.ts` since the version no longer needs per-release bumps
- Eliminates `pnpm-lock.yaml` conflicts on long-lived branches caused by semantic-release bumping the hard dep on every release

## Test plan
- [ ] Verify `pnpm i` resolves correctly with the peer dep
- [ ] Verify `nx build astro-angular` succeeds
- [ ] Confirm `astro-angular` is no longer in `replacementFiles` regex targets

### DIFF
--- a/packages/astro-angular/package.json
+++ b/packages/astro-angular/package.json
@@ -31,10 +31,8 @@
     "type": "github",
     "url": "https://github.com/sponsors/brandonroberts"
   },
-  "dependencies": {
-    "@analogjs/vite-plugin-angular": "^3.0.0-alpha.18"
-  },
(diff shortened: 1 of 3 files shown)


brandonroberts:
Is this a temporary removal?

PR #2212, opened 2026-03-30
feat(astro-angular): use Angular client hydration

eblocha:
## PR Checklist

Closes [#2213](https://github.com/analogjs/analog/issues/2213)

This is an experimental feature to use Angular's client hydration for Astro islands.
See related Angular feature request: https://github.com/angular/angular/issues/67785

## Affected scope

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

## What is the new behavior?

This adds a new configuration option to the astro-angular integration: `experimental.useAngularHydration`.

Setting the new option to `true` will switch the server and client to use Angular's [`provideClientHydration`](https://angular.dev/guide/hydration) feature to hydrate rendered components. Using Angular hydration will reuse the DOM structure rendered on the server, improving performance and the user experience.

Included is support for Angular's hydration features, such as [event replay](https://angular.dev/guide/hydration#how-event-replay-works), [incremental hydration](https://angular.dev/guide/incremental-hydration), and the [http transfer cache](https://angular.dev/guide/ssr#caching-data-when-using-httpclient). These features are configurable per-island.

At the present time, it relies on some internal details inside Angular to function. I've submitted a [feature request](https://github.com/angular/angular/issues/67785) to the Angular team to which will reduce the coupling to Angular implementation details.

Also included: added `continuous: true` to `serve` and `preview` targets. This more accurately reflects the nature of these tasks, allows other tasks to depend on it without waiting for it to complete, and removes the timer from the nx TUI when the task is running.

## Test plan

- [x] `nx format:check`
- [x] `pnpm build`
- [x] `pnpm test`
- [x] Manual verification: added lots of new test scenarios to the astro-app demo.

- [ ] Yes
- [x] No

### DIFF
--- a/apps/astro-app/astro.config.mjs
+++ b/apps/astro-app/astro.config.mjs
@@ -18,9 +18,7 @@ export default defineConfig({
   integrations: [
     angular({
       strictStylePlacement: true,
-      experimental: {
-        useAngularHydration: true,
-      },
+      useAngularHydration: true,
     }),
     react(),
     mdx({ syntaxHighlight: 'prism' }),
--- a/packages/astro-angular/README.md
+++ b/packages/astro-angular/README.md
@@ -203,7 +203,7 @@ import { defineConfig } from 'astro/config';
 import angular from '@analogjs/astro-angular';
 
 export default defineConfig({
-  integrations: [angular({ experimental: { useAngularHydration: true } })],
+  integrations: [angular({ useAngularHydration: true })],
 });
 ```
 


packages/astro-angular/src/server.ts
```diff
@@ -88,10 +90,11 @@ async function renderToStaticMarkup(
   const mirror = reflectComponentType(Component);
   const appId =
     mirror?.selector.split(',')[0] || Component.name.toString().toLowerCase();
+  const ngAppId = 'ng-' + Math.random().toString().slice(2, 9);
```
eblocha: I don't like how this generates a different output each time. It makes it hard to test against because you can't assert against a snapshot.

I'm hoping https://github.com/angular/angular/issues/67785 may help by scoping the state element search to the containing element.

packages/astro-angular/src/server.ts
```diff
@@ -106,6 +109,11 @@ async function renderToStaticMarkup(
           provideServerRendering(),
           { provide: ɵSERVER_CONTEXT, useValue: 'analog' },
           provideZonelessChangeDetection(),
+          provideClientHydration(),
+          {
+            provide: APP_ID,
+            useValue: ngAppId,
+          },
```
eblocha: Angular adds `id="{APP_ID}-state"` to the `<script>` tag containing the client state, then uses `document.getElementById` to recover it. This means we need to ensure there is a unique APP_ID for each rendered component.

packages/astro-angular/tsconfig.json
```diff
@@ -1,14 +1,15 @@
...
+    "module": "nodenext",
     "forceConsistentCasingInFileNames": true,
     "strict": true,
     "noImplicitOverride": true,
     "noPropertyAccessFromIndexSignature": true,
     "noImplicitReturns": true,
     "noFallthroughCasesInSwitch": true,
-    "moduleResolution": "bundler",
+    "moduleResolution": "nodenext",
+    "rewriteRelativeImportExtensions": true,
```
eblocha, line 12: This is needed to be able to import other files and have it work after the package is built. `rewriteRelativeImportExtensions` will let us use `.ts` extensions in the import, which makes the typescript files runnable directly with node (as long as we only use erasable syntax).

apps/astro-app/src/components/todos.component.ts
```diff
@@ -10,27 +9,30 @@ interface Todo {
...
+        <li>
+          {{ todo.title }}
+          @if (todo.completed) {
+            ✅
+          }
+        </li>
+      }
     </ul>
   `,
 })
```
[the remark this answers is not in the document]
eblocha: Standalone is the default.

packages/astro-angular/src/client-ngh.ts
```diff
@@ -0,0 +1,87 @@
...
+      // Not an Angular component
+      return;
+    }
+
+    // Insert Angular client hydration marker
+    // See https://github.com/angular/angular/issues/67785
+    document.body.prepend(
+      document.createComment(ɵSSR_CONTENT_INTEGRITY_MARKER),
+    );
+    document.body.setAttribute(ɵCLIENT_RENDER_MODE_FLAG, '');
```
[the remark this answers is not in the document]
eblocha: We can't be sure that some other island also prepends to the body, so it's safest to always insert a new comment node. This doesn't contribute to bundle size or incur a large performance penalty.

packages/astro-angular/src/server-ngh.ts
```diff
@@ -0,0 +1,132 @@
+import { readFile, readFileSync } from 'node:fs';
```
brandonroberts: Unused import: `readFile` is imported but only `readFileSync` is used. With `"module": "nodenext"` this may cause a build error or ships dead code. Remove `readFile` from the import.

packages/astro-angular/src/client-ngh.ts
```diff
@@ -0,0 +1,87 @@
...
+
+    let hostElement = element.querySelector(mirror.selector);
+    let reuseDom = true;
```
brandonroberts, line 42: `createApplication({...}).then(...)` has no `.catch()` handler. If application creation fails (missing provider, metadata error), the error is silently swallowed as an unhandled promise rejection. Add a `.catch()` or use `async`/`await` with try/catch so hydration failures surface to the developer.

packages/astro-angular/src/index.ts
```diff
@@ -12,13 +12,26 @@ interface AngularOptions {
...
+     * Use Angular's `provideClientHydration` to hydrate components.
+     */
+    useAngularHydration?: boolean;
+  };
 }
 
-function getRenderer(): AstroRenderer {
+function getRenderer(ngHydration: boolean | undefined): AstroRenderer {
   return {
     name: '@analogjs/astro-angular',
```
brandonroberts, line 23: `getViteConfiguration()` still hardcodes `@analogjs/astro-angular/client.js` in `optimizeDeps.include` regardless of `useAngularHydration`. When the flag is on, the actual entrypoints are `client-ngh.js` and `server-ngh.js`. Pass the flag into `getViteConfiguration` and conditionally include the correct paths.

packages/astro-angular/src/client-ngh.ts
```diff
@@ -0,0 +1,87 @@
...
+} from './create-component.ts';
+import { ID_PROP_NAME } from './id.ts';
+
+export default (element: HTMLElement) => {
+  return (
+    Component: Type<unknown> & {
+      clientProviders?: (Provider | EnvironmentProviders)[];
+      hydrationFeatures?: () => HydrationFeature<HydrationFeatureKind>[];
+    },
+    props?: Record<string, unknown>,
```
brandonroberts, line 30: These markers are prepended/set on every hydration call. Multiple Angular islands on the same page will accumulate duplicate markers. Guard with an existence check:

```ts
if (!document.body.hasAttribute(ɵCLIENT_RENDER_MODE_FLAG)) {
  document.body.prepend(document.createComment(ɵSSR_CONTENT_INTEGRITY_MARKER));
  document.body.setAttribute(ɵCLIENT_RENDER_MODE_FLAG, "");
}
```

packages/astro-angular/src/server-ngh.ts
```diff
@@ -0,0 +1,132 @@
...
+    document,
+  });
+
+  // Since astro renders components in parallel, we must reset the ngServerMode global to `true` after rendering the component.
+  // The `renderApplication` function resets this back to `undefined` before returning.
+  globalThis.ngServerMode = true;
+
+  document.documentElement.innerHTML = html;
+  let styleTags = '';
+
```
brandonroberts, line 111: This couples to an undocumented Angular internal behavior. Add a comment referencing the specific Angular source location this depends on so it can be tracked during Angular upgrades. Ideally this should be part of the feature request at angular/angular#67785.

packages/astro-angular/src/index.ts
```diff
@@ -12,13 +12,26 @@ interface AngularOptions {
    * Enabling this option disables astro's streaming under SSR.
    */
   strictStylePlacement?: boolean;
+  /**
+   * Experimental options
+   */
+  experimental?: {
+    /**
+     * Use Angular's `provideClientHydration` to hydrate components.
+     */
```
brandonroberts: The new entrypoints (`client-ngh.ts`, `server-ngh.ts`) and their dependencies (`context.ts`, `create-component.ts`, `server-providers.ts`, `id.ts`) should be exposed through `package.json` exports (e.g., `./client-ngh`, `./server-ngh`) so consumers and Vite can resolve them as proper package subpaths rather than relying on `fileURLToPath` filesystem resolution. This also ensures they are included in the published package.

brandonroberts:
@eblocha this doesnt need to be an experimental feature if its working. Hydration is stable in Angular, so we can fix any issues going forward

packages/astro-angular/src/server-providers.ts
```diff
@@ -0,0 +1,52 @@
...
+
+        return () => {
+          // Note: appRef.tick() can cause NG0100 errors here
+          appRef.components.forEach((compRef) =>
+            compRef.changeDetectorRef.detectChanges(),
+          );
+        };
+      },
+      multi: true,
+    },
```
eblocha: I noticed this causes some issues with hydration in some complex components I tested this on. I originally put this here because the todos component was not rendering its fetched data, but I think it was just incompatible with zoneless change detection.

After I updated the todos component to store the todos in a signal, it is now properly rendering the list.

I'll remove this provider.

apps/astro-app/astro.config.mjs
```diff
@@ -16,8 +16,24 @@ export default defineConfig({
     },
   },
   integrations: [
-    angular({ strictStylePlacement: true }),
+    angular({
+      strictStylePlacement: true,
+      experimental: {
+        useAngularHydration: true,
+      },
```
brandonroberts:
```suggestion
      useAngularHydration: true,
```

packages/astro-angular/README.md
```diff
@@ -191,6 +191,26 @@ export default defineConfig({
...
+
+To opt-in to Angular's client hydration, enable the `experimental.useAngularHydration` option in the integration config. This will switch the hydration strategy to use [provideClientHydration](https://angular.dev/api/platform-browser/provideClientHydration).
+
+```js
+import { defineConfig } from 'astro/config';
+
+import angular from '@analogjs/astro-angular';
+
+export default defineConfig({
+  integrations: [angular({ experimental: { useAngularHydration: true } })],
```
brandonroberts:
```suggestion
  integrations: [angular({ useAngularHydration: true })],
```

eblocha: Good catch, thanks. Pushed a commit to fix this.

brandonroberts:
Thanks @eblocha!

Issue #12, opened 2022-07-29
[FEAT]: Add Git initialization to template creation


brandonroberts:
When running `yarn create analog`, initialize the git repository with an initial commit

brandonroberts:
Fixed via aadf01835a0c6bfd184795a1f821baf8caec56c0

Issue #11, opened 2022-07-27
[FEAT]: Add support for HMR in development mode


brandonroberts:
Vite uses HMR to enable refreshing the app without reloading the whole page. Vite uses HMR closest to the code that accepts HMR.

Angular has some support for this but not as granular as the root app needs to be destroyed and recreated for this to work.

External styles outside of component templates and HTML work as intended.

Prior art: https://github.com/angular/angular-cli/blob/main/packages/angular_devkit/build_angular/src/webpack/plugins/hmr/hmr-accept.ts

brandonroberts:
Plugin as-is when used in Vite environment does a flash on each HMR reload. Need to investigate this so it appears native.

omerman:
👍

brandonroberts:
Will revisit this when better HMR support lands in Angular

Issue #10, opened 2022-07-22
[FEAT]: Investigate bundle size differences between Webpack build and Vite build


brandonroberts:
Currently, the Vite build is significantly larger than a Webpack build for production

The Webpack-based builder does a 2-pass transform with esbuild/terser

Vite can also do a 2-pass transform with esbuild/terser to render chunks, but it doesn't happen for each file.

brandonroberts:
Angular mjs files needed to be transformed to enable better treeshaking
Along with esbuild optimization flags, the production build is on par with the Webpack build

Fixed via b78cfad3e8b49a1c04565a228022676c222d9b92

PR #2226, opened 2026-04-03
refactor: overhaul stylesheet pipeline and improve Tailwind HMR across Analog

benpsnyder:
## PR Checklist

Related:
- analogjs/analog#2229
- analogjs/analog#2227
- tailwindlabs/tailwindcss#19903
- tailwindlabs/tailwindcss#19904

## Affected scope

- Secondary scopes: `platform`, `router`, `vite-plugin-nitro`, `content`, `create-analog`, docs, repo tooling, examples, test apps

- [ ] Squash merge
- [x] Rebase merge
- [ ] Other

This branch preserves intentional commit boundaries across multiple packages and follow-up review fixes. A rebase merge keeps that history readable without adding an extra merge commit.

## What is the new behavior?

This branch is broader than the PR title alone suggests.

The center of gravity is still Angular component stylesheet HMR in `@analogjs/vite-plugin-angular`, but the branch also includes the dedicated reproduction app, route/content/runtime fixes needed to validate that work in-repo, the Tailwind full-reload repro alignment, and the follow-up maintainer review fixes that landed on top.

Highlights:
- `@analogjs/vite-plugin-angular` centralizes stylesheet handling for HMR, expands debug logging controls, improves component stylesheet HMR, restores expected `optimizeDeps` prebundling, cleans up stale metadata on file deletion, removes duplicate Vite-ignore injection, fixes class-binding conflict false positives, and limits component metadata extraction to actual `@Component(...)` decorators.
- The dedicated `tailwind-debug-app` and its e2e coverage now reproduce the current upstream Tailwind full-reload behavior from `tailwindlabs/tailwindcss#19903` instead of asserting the opposite behavior, while the supporting debug stream and diagnostics flow were tightened based on review feedback.
- `@analogjs/platform` adds dev-time route idiom diagnostics, clearer debug reload reasons, cached page-route discovery for diagnostics, and full rescans when route graph shape changes.
- `@analogjs/router` keeps explicit app-route precedence while preserving historical last-wins behavior for equal-priority collisions, and route metadata listeners are now active during app init instead of being deferred past first navigation.
- `@analogjs/vite-plugin-nitro` no longer requires an SSR entry for `ssr: false` apps that explicitly opt out of prerender routes, which unblocks production builds for the repro app.
- `@analogjs/content` keeps the devtools API on the dedicated `@analogjs/content/devtools` entrypoint instead of the main package entrypoint.
- `create-analog` scaffolds the Tailwind v4 PostCSS config needed by the debug workflow.
- Repo dependency/catalog cleanup and docs/test updates are included as supporting work.

## Test plan

This PR description reflects targeted validation that was actually run on the branch; it does not claim a fresh full-workspace validation pass after the latest commit stack rewrite.

- [ ] `nx format:check`
- [ ] `pnpm build`
- [ ] `pnpm test`
- [x] Targeted verification

Targeted verification run on this branch:
- `pnpm vitest run packages/vite-plugin-angular/src/lib/component-resolvers.spec.ts packages/vite-plugin-angular/src/lib/stylesheet-registry.spec.ts packages/platform/src/lib/router-plugin.spec.ts packages/platform/src/lib/utils/debug.spec.ts`
- `pnpm nx test tailwind-debug-app --runInBand`
- `pnpm vitest run --config packages/vite-plugin-angular/vite.config.ts packages/vite-plugin-angular/src/lib/angular-vite-plugin.spec.ts`
- `pnpm vitest run --config packages/vite-plugin-angular/vite.config.ts packages/vite-plugin-angular/src/lib/angular-vite-plugin-live-reload.spec.ts`
- `CI=1 pnpm nx run tailwind-debug-app-e2e:e2e`
- `pnpm nx run tailwind-debug-app:build:production`
- `pnpm nx test vite-plugin-nitro --runInBand`
- `pnpm oxlint -c oxlint.config.ts --quiet apps/tailwind-debug-app/src/app/debug/debug-stream.service.ts apps/tailwind-debug-app/src/app/debug/hmr-diagnostics.ts apps/tailwind-debug-app/src/app/probes/debug-stream-panel.component.ts apps/tailwind-debug-app/src/app/probes/style-probe.component.spec.ts apps/tailwind-debug-app/vite.config.ts apps/tailwind-debug-app-e2e/tests/component-css-hmr.spec.ts packages/platform/src/lib/content-plugin.ts packages/platform/src/lib/router-plugin.ts packages/platform/src/lib/router-plugin.spec.ts packages/platform/src/lib/utils/debug.spec.ts packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts packages/vite-plugin-angular/src/lib/component-resolvers.ts packages/vite-plugin-angular/src/lib/component-resolvers.spec.ts packages/vite-plugin-angular/src/lib/stylesheet-registry.spec.ts packages/vite-plugin-nitro/src/lib/vite-plugin-nitro.spec.ts`

- [x] Yes
- [ ] No

Breaking change:
- `@analogjs/content` devtools exports are intentionally no longer re-exported from the main `@analogjs/content` entrypoint. Consumers must import them from `@analogjs/content/devtools`.

## Other information

Current branch stack relative to `alpha` (`33` commits):

| Commit | Scope | Summary |
| --- | --- | --- |
| `6ce1d364` | `vite-plugin-angular` | centralize stylesheet handling for hmr |
| `eba6e08e` | `vite-plugin-angular` | expand debug logging controls |
| `2c49f0a2` | `content` | split devtools into a dedicated entrypoint |
| `8cbb8fd4` | `router` | prefer app routes over shared duplicates |
| `f23ce5a2` | `platform` | restore analog-app action and content examples |
| `8eec739f` | `router` | cover legacy PageServerAction compatibility |
| `84002ee3` | `create-analog` | scaffold Tailwind v4 PostCSS config |
| `b9b60be7` | `router` | preserve app route precedence in production |
| `09482a9e` | `platform` | add dev-time route idiom diagnostics |
| `d8742e39` | `vite-plugin-angular` | improve component stylesheet hmr |
| `87ad7d21` | `test` | align fixtures with explicit selector enforcement |
| `2f2f7171` | `router` | stabilize legacy page action fixture |
| `162043de` | `platform` | emit debug reload reasons for content changes |
| `605d2150` | `build` | adopt workspace catalogs for repo deps |
| `5156354a` | `router` | defer route metadata setup during bootstrap |
| `7c9c2e9e` | `chore` | add tailwind debug coverage and stabilize e2e |
| `677a4a69` | `vite-plugin-angular` | handle `>` inside quoted class bindings |
| `1683e4ee` | `vite-plugin-angular` | ignore `[class.foo]` in `ngClass` conflict guard |
| `6a6efc4a` | `router` | preserve last-wins behavior for equal-priority collisions |
| `7eafecb9` | `router` | register route metadata listeners during app init |
| `b55a4366` | `vite-plugin-angular` | restore rxjs `optimizeDeps` prebundling |
| `8a93d0cd` | `vite-plugin-angular` | evict stale metadata on file deletion |
| `002de6c9` | `platform` | cache page-route discovery for diagnostics |
| `61807f34` | `vite-plugin-angular` | dedupe vite-ignore hmr injection |
| `ba044b0f` | `chore` | remove investigation snapshot artifacts |
| `135c883e` | `build` | move all root package deps to workspace catalogs |
| `fb701b8f` | `docs` | note Angular HMR version requirements |
| `9c6a2966` | `fix` | harden tailwind debug app HMR diagnostics |
| `779c9c8f` | `platform` | refresh route diagnostics on route graph changes |
| `c37ff35f` | `vite-plugin-angular` | restrict metadata extraction to component decorators |
| `bcadc13e` | `vite-plugin-angular` | tighten resolver coverage docs |
| `9e0be352` | `vite-plugin-angular` | align tailwind debug app e2e with upstream full-reload repro |
| `c091cf7a` | `vite-plugin-nitro` | avoid SSR entry lookup for `ssr: false` apps with empty prerender routes |

### DIFF
--- a/apps/tailwind-debug-app/vite.config.ts
+++ b/apps/tailwind-debug-app/vite.config.ts
@@ -102,6 +102,9 @@ export default defineConfig(({ mode }) => ({
     analog({
       apiPrefix: 'api',
       hmr: true,
+      prerender: {
+        routes: [],
+      },
       ssr: false,
       nitro: {
         routeRules: {
--- a/packages/vite-plugin-nitro/src/lib/vite-plugin-nitro.spec.ts
+++ b/packages/vite-plugin-nitro/src/lib/vite-plugin-nitro.spec.ts
@@ -409,6 +409,63 @@ describe('nitro', () => {
     }
   });
 
+  it('does not require an SSR entry for client-only apps with explicit empty prerender routes', async () => {
+    vi.stubEnv('VITEST', '');
+    vi.stubEnv('NODE_ENV', 'production');
+    const { buildServerImportSpy } = await mockBuildFunctions();
+    const workspaceRoot = mkdtempSync(join(tmpdir(), 'analog-nitro-'));
+
+    try {
+      writeBuiltClientIndexHtml(
+        workspaceRoot,
+        '<html>client only explicit prerender opt-out</html>',
+      );
+
+      const plugin = nitro({
+        workspaceRoot,
+        ssr: false,
+        prerender: {
+          routes: [],
+        },
+      });
+      const result = await (plugin[1].config as any)(
+        {},
+        { command: 'build', mode: 'production' },
+      );
+
+      const builderBuild = vi.fn().mockResolvedValue(undefined);
+      await result.builder.buildApp({
+        build: builderBuild,
+        environments: {
+          client: {},
+          ssr: {},
+        },
+      });
+
+      expect(builderBuild).toHaveBeenCalledTimes(1);
+      expect(builderBuild).not.toHaveBeenCalledWith(
+        expect.objectContaining({
+          config: expect.objectContaining({
+            build: expect.objectContaining({
+              ssr: true,
+            }),
+          }),
+        }),
+      );
+
+      const nitroConfig = buildServerImportSpy.mock.calls[0][1];
+      expect(nitroConfig.alias?.['#analog/ssr']).toBeUndefined();
+      expect(nitroConfig.virtual?.['#ANALOG_CLIENT_RENDERER']).toContain(
+        "import template from '#analog/index';",
+      );
+      expect(nitroConfig.virtual?.['#analog/index']).toBe(
+        'export default "<html>client only explicit prerender opt-out</html>";',
+      );
+    } finally {
+      rmSync(workspaceRoot, { recursive: true, force: true });
+    }
+  });
+
   it('should resolve client output path correctly for nested roots without explicit build.outDir', async () => {
     vi.stubEnv('VITEST', '');
     vi.stubEnv('NODE_ENV', 'production');


benpsnyder:
Maintainer context from `apps/analog-app` while validating this branch:

## What We Fixed And Why

This app exposed three separate regressions while exercising production build output through `analog-app-e2e`.

### 1. The browser bundle pulled in Vite-only content devtools code

What broke:

The `@analogjs/content` main entrypoint re-exported devtools symbols. Those devtools modules depend on `vite`, which is a build-time dependency and not safe to ship into the browser bundle.

In workspace development, that meant the app client bundle could resolve through source, pull Vite internals into runtime code, and fail during hydration with errors like `createRequire is not a function`.

What we changed:

- Removed devtools exports from the main `@analogjs/content` entrypoint.
- Added a dedicated `@analogjs/content/devtools` secondary entrypoint.
- Added the matching library build output and package export so the devtools API is still available intentionally.

Why this is the right fix:

The main content package should stay browser-safe. Devtools are optional and build-time oriented, so they need an explicit opt-in entrypoint instead of piggybacking on the default one.

### 2. The newsletter page used the old action path

What broke:

The newsletter example still used the older server action shape and had an unsupported template binding. That left the example out of step with the current `defineAction(...)` flow and caused the newsletter e2e to fail.

What we changed:

- Converted the newsletter server handler to `defineAction(...)`.
- Added schema validation for the email field.
- Removed the unsupported `(onStateChange)` binding from the template.

Why this is the right fix:

This example app is part of the regression surface for Analog features. It should demonstrate the current action API, not a legacy path that quietly drifts out of coverage.

### 3. The no-SSR `/client` route depended on server-only rendering

What broke:

The `/client` route is intentionally client-only, but it imported and rendered a server-only component. In the production app bootstrap path that led to a runtime failure on navigation, the page never rendered, and the JSON-LD test failed because the route content never mounted.

What we changed:

- Simplified the `/client` page so it is fully client-safe.
- Kept the page title and JSON-LD metadata.
- Kept a small interactive signal example so the page still demonstrates client rendering.

Why this is the right fix:

A no-SSR route has to be self-consistent. If the route is meant to render only on the client, its example code cannot depend on server-only components and still serve as a reliable regression target.

## Regression coverage

- `analog-app-e2e` covers the client-only `/client` route and the newsletter flow.
- `analog-app-e2e/tests/legacy-server-action.spec.ts` keeps the supported `PageServerAction` flow covered at the app level because that compatibility path is still intentionally present in the example app.
- `packages/content/src/index.spec.ts` guards the public entrypoint split so devtools exports do not accidentally leak back into the main browser-safe entrypoint.
- `packages/router/server/actions/src/page-server-action-compat.spec.ts` keeps the same `PageServerAction` compatibility pattern covered at the package level until that API is explicitly deprecated and removed.

benpsnyder:
@coderabbitai
Verified the findings against the current branch and fixed the ones that still applied.

The main code fixes were in packages/router/src/lib/routes.spec.ts:978, where the duplicate-route test now traverses the nested blog -> :slug route tree and the console.warn spy is cleaned up with beforeEach/afterEach; packages/vite-plugin-angular/src/lib/angular-jit-plugin.ts, which now emits a visible warning as well as structured debug output on JIT CSS preprocess failures; packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts, which now warns when Angular <19 forces HMR off; and packages/vite-plugin-angular/src/lib/stylesheet-registry.ts:70, which now rewrites @import url(...) forms and removes the collision-prone basename alias. I also removed the unused Nitro/debug imports, switched the app config to prefer hmr while keeping liveReload as a deprecated alias, fixed the legacy action label/input association, fixed the flaky E2E readiness selector, and corrected the Angular Material docs install command.

One review note was stale against current code: the angularVitestPlugins() helpers are not the HMR toggle. I updated packages/vite-plugin-angular/src/lib/angular-vite-plugin.spec.ts:24 to verify the actual HMR helper plugins under a development-like env instead of asserting against always-present Vitest plugins, and I updated packages/vite-plugin-angular/src/lib/stylesheet-registry.spec.ts to match the deliberate removal of basename lookup.

Verified with pnpm nx test router --skip-nx-cache and pnpm nx test vite-plugin-angular --skip-nx-cache. I did not run the Playwright E2E suite or docs/site validation.

packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
```diff
@@ -947,9 +1346,9 @@ export function angular(options?: PluginOptions): Plugin[] {
     const workspaceRoot = normalizePath(resolve(pluginOptions.workspaceRoot));
 
     // Map include patterns to absolute workspace paths
-    const globs = [
-      ...pluginOptions.include.map((glob) => `${workspaceRoot}${glob}`),
-    ];
+    const globs = pluginOptions.include.map(
+      (glob) => `${workspaceRoot}${glob}`,
+    );
```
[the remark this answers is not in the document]
benpsnyder: Latest commit should address if it is a concern

packages/router/src/lib/route-builder.ts
```diff
@@ -31,7 +31,16 @@ export function createRoutes<TFile>(
...
+  const filenames = Object.keys(files).sort((a, b) => {
+    const aPriority = getCollisionPriority(a);
+    const bPriority = getCollisionPriority(b);
+
+    if (aPriority !== bPriority) {
+      return aPriority - bPriority;
+    }
+
+    return a.localeCompare(b);
+  });
```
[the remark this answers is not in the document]
benpsnyder: Latest commit should address if it is a concern

apps/docs-app/docs/guides/migrating.md
```diff
@@ -194,8 +194,35 @@ export default defineConfig(({ mode }) => ({
...
+      vite: {
+        tailwindCss: {
+          rootStylesheet: resolve(__dirname, 'src/styles.css'),
+        },
+      },
+    }),
+    tailwindcss(),
+  ],
+}));
+```
```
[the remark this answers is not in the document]
benpsnyder: Latest commit should address if it is a concern

apps/docs-app/docs/integrations/angular-material/index.md
```diff
@@ -164,7 +164,7 @@ export default defineConfig({
 @import 'tailwindcss';
 ```
 
-> **Note:** Analog's default Tailwind v4 setup uses `@tailwindcss/vite`. You do not need a `.postcssrc.json` file or a generated `tailwind.config.*` file for the standard Vite-based setup.
+> **Note:** Analog's current Tailwind v4 setup uses `@tailwindcss/vite` and also commonly includes a `postcss.config.mjs` with `@tailwindcss/postcss` so the build path and related integrations use the same Tailwind pipeline.
 >
 > The scaffolded Tailwind v4 flow also expects a plain CSS entry file such as `src/styles.css`. If your app currently uses Sass or Less for the global entry point, keep your existing setup or migrate that entry file to CSS before adopting the default Tailwind v4 flow.
```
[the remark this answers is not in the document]
benpsnyder: Latest commit should address if it is a concern

packages/vite-plugin-nitro/src/lib/vite-plugin-nitro.ts
```diff
@@ -32,7 +32,7 @@ import {
   apiMiddleware,
 } from './utils/renderers.js';
 import { getBundleOptionsKey, isRolldown } from './utils/rolldown.js';
-import { debugNitro, debugSsr } from './utils/debug.js';
+import { debugNitro, debugSsr, debugPrerender } from './utils/debug.js';
```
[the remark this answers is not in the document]
benpsnyder: Latest commit should address if it is a concern

packages/router/src/lib/route-builder.ts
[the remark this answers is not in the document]
benpsnyder: Addressed in 20859121a01fdf01d18d8d87a4b4a921d3ecf67e. Duplicate route precedence now short-circuits in all modes so production preserves the first winning file just like dev; the dev-only behavior that remains is the warning, not the precedence decision.

packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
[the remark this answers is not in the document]
benpsnyder: Addressed in 20859121a01fdf01d18d8d87a4b4a921d3ecf67e. `findIncludes()` now normalizes all three supported forms: already-rooted workspace globs, `/`-prefixed workspace-relative globs, and bare relative globs like `libs/**`. I also added unit coverage for those cases in `angular-vite-plugin.spec.ts`.

apps/docs-app/docs/guides/migrating.md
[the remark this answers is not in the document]
benpsnyder: Addressed in 20859121a01fdf01d18d8d87a4b4a921d3ecf67e. The migration guide example now uses `resolve(import.meta.dirname, 'src/styles.css')` instead of `__dirname`, which matches the repo's supported Node/ESM baseline.

packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
brandonroberts, line 1: **[1/12] Logic bug: Tag-matching regex can produce false positives and block builds**

`findStaticClassAndBoundClassConflicts` uses `/<([a-zA-Z][\w:-]*)([\s\S]*?)>/g` to match open tags. The `[\s\S]*?` group can match across `>` characters inside attribute values containing `>` (e.g., `[class]="a > b ? 'x' : 'y'"`), producing false tag boundaries and incorrect conflict detection.

Since this guard runs `enforce: 'pre'` and **throws errors** (not warnings), false positives will block builds for valid templates. Consider:
- Making the static-class + `[class]` check a warning instead of a hard error, or
- Using a proper HTML tokenizer (oxc-parser is already a dependency for route diagnostics).

brandonroberts, line 1: **[2/12] Logic bug: `findBoundClassAndNgClassConflicts` false positive on `[class.foo]` bindings**

The function gates on `snippet.includes('[class]')`, but `[class.active]` also contains the substring `[class]`. This means an element like:

```html
<div [class.active]="isActive" [ngClass]="{'bold': isBold}"></div>
```

will be flagged as a conflict even though `[class.foo]` and `[ngClass]` are safe to combine. The guard should use a pattern that excludes `[class.` — e.g., check for `[class]="` or use a regex like `/\[class\]\s*=/`.

packages/router/src/lib/route-builder.ts
brandonroberts, line 1: **[3/12] Behavior change: Route collision resolution changed from "last wins" to "first wins"**

The new `getCollisionPriority` sort + early `return acc` on collision silently changes which route file wins when two files resolve to the same path. Previously the last file in iteration order would overwrite the first; now the first (after sort) wins and duplicates are discarded.

The priority heuristic checks fixed path fragments like `/src/app/pages/` and `/app/routes/`. Monorepo layouts with non-standard directory structures (e.g., `packages/feature-x/pages/`) won't match any of these patterns and will all get priority `1`, falling back to `localeCompare` order — which may not be the intuitive winner.

The warning message and the `return acc` early-exit are correct for the new logic, but this is a silent behavior change for existing users whose apps happen to have colliding route files.

packages/content/src/index.ts
brandonroberts, line 1: **[4/12] Breaking change: Devtools exports removed from main entrypoint**

`contentDevToolsPlugin`, `DevToolsContentRenderer`, and `withContentDevTools` are removed from `@analogjs/content` and moved to `@analogjs/content/devtools`. Any consumer currently importing these from the main entrypoint will get a build error after upgrading.

The PR body says "No" to breaking changes, but this qualifies as one. Should either be documented as a breaking change or re-exported from the main entrypoint with a deprecation notice.

packages/router/src/lib/provide-file-router-base.ts
brandonroberts, line 1: **[5/12] Timing change: Meta tags and JSON-LD now initialize later (post-bootstrap microtask)**

Switching from `ENVIRONMENT_INITIALIZER` to `APP_BOOTSTRAP_LISTENER` with `queueMicrotask` defers `updateMetaTagsOnRouteChange()` and `updateJsonLdOnRouteChange()` to after the bootstrap listener fires and after the current microtask queue drains.

The injection-context fix in `json-ld.ts` / `meta-tags.ts` (`runInInjectionContext`) is sound and addresses a real issue. But the timing change could cause meta tags and JSON-LD to be absent during the first SSR render if the server resolves the HTML response before the microtask fires. This needs validation with SSR prerendering — if the initial `NavigationEnd` fires during bootstrap, the subscription won't be active yet.

packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
brandonroberts, line 1: **[6/12] `optimizeDeps.include` changed from `['rxjs/operators', 'rxjs']` to `['tslib']`**

Dropping rxjs from `optimizeDeps.include` means Vite won't pre-bundle rxjs on dev-server cold start. Since rxjs is a heavy CJS dependency with many submodules, this can measurably increase cold-start time as Vite discovers and transforms rxjs modules on demand. Was this intentional?

brandonroberts, line 1: **[7/12] Module-scoped mutable state grows unboundedly during dev session**

`activeGraphComponentMetadata`, `selectorOwners`, `classNameOwners`, `transformedStyleOwnerMetadata`, and `styleSourceOwners` are Maps that grow as files are transformed but are only cleaned up when a file is re-transformed (via `removeActiveGraphMetadata` / `removeStyleOwnerMetadata`). Deleted files never get their entries removed because `removeActiveGraphMetadata` is only called inside `registerActiveGraphMetadata`, which only fires during transform.

For long-running dev sessions with file churn (creating/deleting components), stale entries will accumulate and could cause false-positive duplicate-selector errors for selectors that no longer exist in the codebase.

packages/platform/src/lib/router-plugin.ts
brandonroberts, line 1: **[8/12] Performance: Synchronous file reads and re-globbing on every `.page.ts` change**

`reportRouteDiagnostics` calls `discoverRouteFiles()` (which runs `globSync`) and then `readFileSync` + `parseSync` (oxc-parser) for a single file. On server start, it also iterates all `.page.ts` files and runs this for each one.

For large apps with many route files, this adds startup latency. More importantly, `discoverRouteFiles()` is called on every single `.page.ts` file change — even edits that don't affect the route graph shape. Consider caching the route file list and only re-globbing when a file is added/removed (the `invalidateDiscoveryCaches()` call already tracks this).

packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
brandonroberts, line 1: **[9/12] Duplicate `@vite-ignore` injection logic**

The `@vite-ignore` regex replacement for `getReplaceMetadataURL` appears in two places:
1. Inside `angularPlugin().transform` (runs during normal transform)
2. In the dedicated `@analogjs/vite-plugin-angular:hmr-vite-ignore` plugin (`enforce: 'post'`)

Both use the same regex pattern. Since the `enforce: 'post'` plugin exists specifically for this purpose and runs after all other transforms, the inline copy in `angularPlugin().transform` appears redundant. Having both means the regex runs twice on every component file.

changes/2026-04-04-reset-investigation/README.md
brandonroberts, line 1: **[10/12] Investigation log appears accidentally included**

This file looks like a personal investigation/debugging log. It probably shouldn't be committed to the repo as part of this PR.

package.json
brandonroberts, line 1: **[11/12] `@angular/platform-server` version specifier inconsistency**

Changed to `~21.2.6` (tilde range) while all other Angular packages use exact `21.2.6`. This could cause version drift between `@angular/platform-server` and the rest of the Angular packages if a patch is published. Was this intentional?

brandonroberts, line 1: **[12/12] `workspaces` field removed**

The `workspaces` array was removed from `package.json`. While pnpm uses `pnpm-workspace.yaml` for workspace discovery, some tooling (e.g., npm-based tools, certain IDE features) may depend on the `workspaces` field in `package.json`. Was this removal intentional?

brandonroberts:
## Review Summary

12 findings posted as inline comments. Here's the overview:

### Contribution Guidelines

| Guideline | Status |
|-----------|--------|
| **PR title format** | ✅ `fix(vite-plugin-angular): ...` — valid type and scope |
| **Scope accuracy** | ⚠️ Title scope is `vite-plugin-angular`, but substantial changes span `router`, `content`, `platform`, `create-analog`, `storybook-angular`. Squash merge will misattribute. |
| **Tests included** | ⚠️ Tests present for new functionality, but test plan has unchecked items (`nx format:check`, `pnpm build`, `pnpm test`). |
| **Related commits** | ⚠️ 14 commits span ~6 distinct concerns (HMR, route diagnostics, content devtools split, router precedence, legacy action compat, workspace catalogs). |
| **Breaking changes** | ❌ PR says "No", but content devtools export removal (comment #4) is breaking. |
| **Squash merge** | ✅ Selected, but see scope accuracy above. |

### Verdict

| Area | Verdict |
|------|---------|
| Tag-matching regex false positives (#1, #2) | Should fix before merge |
| Content devtools breaking change (#4) | Should fix before merge |
| Investigation file (#10) | Should fix before merge |
| Scope / commit disjointness | Should fix before merge |
| Bootstrap timing change (#5) | Minor concern — validate with SSR |
| Route collision behavior change (#3) | Minor concern |
| optimizeDeps change (#6) | Minor concern |
| Module-scoped unbounded state (#7) | Minor concern |
| Route diagnostics perf (#8) | Minor concern |
| Duplicate `@vite-ignore` logic (#9) | Minor concern |
| Version specifier inconsistency (#11) | Minor concern |
| `workspaces` removal (#12) | Minor concern |

benpsnyder:
@coderabbitai resume

benpsnyder:
> ## Review Summary
> 12 findings posted as inline comments. Here's the overview:
> 
> ### Contribution Guidelines
> Guideline	Status
> **PR title format**	✅ `fix(vite-plugin-angular): ...` — valid type and scope
> **Scope accuracy**	⚠️ Title scope is `vite-plugin-angular`, but substantial changes span `router`, `content`, `platform`, `create-analog`, `storybook-angular`. Squash merge will misattribute.
> **Tests included**	⚠️ Tests present for new functionality, but test plan has unchecked items (`nx format:check`, `pnpm build`, `pnpm test`).
> **Related commits**	⚠️ 14 commits span ~6 distinct concerns (HMR, route diagnostics, content devtools split, router precedence, legacy action compat, workspace catalogs).
> **Breaking changes**	❌ PR says "No", but content devtools export removal (comment #4) is breaking.
> **Squash merge**	✅ Selected, but see scope accuracy above.
> ### Verdict
> Area	Verdict
> Tag-matching regex false positives (#1, #2)	Should fix before merge
> Content devtools breaking change (#4)	Should fix before merge
> Investigation file (#10)	Should fix before merge
> Scope / commit disjointness	Should fix before merge
> Bootstrap timing change (#5)	Minor concern — validate with SSR
> Route collision behavior change (#3)	Minor concern
> optimizeDeps change (#6)	Minor concern
> Module-scoped unbounded state (#7)	Minor concern
> Route diagnostics perf (#8)	Minor concern
> Duplicate `@vite-ignore` logic (#9)	Minor concern
> Version specifier inconsistency (#11)	Minor concern
> `workspaces` removal (#12)	Minor concern

Updated PR Description; recommend rebase merge. Subsequent commits will be based on maintainer feedback and worthy of rebase merge.

packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
benpsnyder: @coderabbitai fixed this 677a4a692e5961be04113b60386721ac687ef780

benpsnyder: Try now

benpsnyder: @coderabbitai review 677a4a692e5961be04113b60386721ac687ef780 if it addresses this issue

benpsnyder: @coderabbitai this is addressed in 1683e4ee0b844cb6b101dbddd6af7f123b2c78df

packages/router/src/lib/route-builder.ts
benpsnyder: What changed:

  - Cross-priority collisions still preserve the intended explicit precedence, so app-local routes continue to beat shared/additional routes.
  - Same-priority collisions now keep the historical overwrite behavior, so non-standard layouts don’t silently switch winner just because of filename sorting.
  - The warning message now reports the actual winner in both cases.

  I added coverage in packages/router/src/lib/routes.spec.ts for both:

  - app-local route beats shared route
  - same-priority shared-route collision remains “last wins”

  Verification:

  - pnpm vitest run -c packages/router/vite.config.ts packages/router/src/lib/routes.spec.ts -t "duplicate route precedence": passed
  - pnpm oxlint -c oxlint.config.ts --quiet packages/router/src/lib/route-builder.ts packages/router/src/lib/routes.spec.ts: 0 errors

packages/content/src/index.ts
benpsnyder: This was recently shipped within days and nobody is using it. I don't see the need for documenting a breaking change.

packages/router/src/lib/provide-file-router-base.ts
benpsnyder: I changed `packages/router/src/lib/provide-file-router-base.ts` to register meta tag and JSON-LD listeners via `provideAppInitializer(...)` instead of the deferred `APP_BOOTSTRAP_LISTENER` + `queueMicrotask()` path, and removed the dev-only skip so behavior is consistent across environments. I also added a provider-level regression test in `packages/router/src/lib/provide-file-router.spec.ts` that verifies those listeners are active on first navigation.
7eafecb9d0dfccace3580befd760f12b7b0f30fd

packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
benpsnyder: I restored RxJS prebundling in `packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts` so `optimizeDeps.include` is now:

  - rxjs/operators
  - rxjs
  - tslib

  I also added a focused config-hook test in `packages/vite-plugin-angular/src/lib/angular-vite-plugin.spec.ts` that asserts the plugin still emits that `optimizeDeps.include` list

benpsnyder: Addressed in `packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts`.

The stale-metadata concern on deleted files was valid. Cleanup previously happened on re-register paths, which meant a deleted component file could leave selector/style ownership behind until something else happened to overwrite it.

This branch now evicts that state on the watcher `unlink` path before recompilation:

- active component metadata
- selector ownership
- component class ownership
- style-owner metadata
- per-file transform/class caches

I made the deletion behavior explicit with `evictDeletedFileMetadata(...)` and added a focused regression test in `packages/vite-plugin-angular/src/lib/angular-vite-plugin.spec.ts` to verify the deleted file’s metadata is removed immediately.

packages/platform/src/lib/router-plugin.ts
benpsnyder: Addressed in `packages/platform/src/lib/router-plugin.ts`.

The repeated sync work concern was valid. The route/content/endpoint discovery caches were already being reused for normal edits, but the route-idiom diagnostics path was still rebuilding the `.page.ts` subset on every changed page.

This branch now caches that page-route subset explicitly and reuses it for both:

- startup diagnostics
- subsequent `.page.ts` change diagnostics

The cache is still invalidated on add/unlink, which is the point where the route graph shape actually changes. We still synchronously read/parse the changed file itself for diagnostics, but plain edits no longer trigger another route glob.

I also added a regression test in `packages/platform/src/lib/router-plugin.spec.ts` that verifies a `.page.ts` change reuses the cached discovery result and does not call `globSync` again.

Validation:

- `pnpm vitest run packages/platform/src/lib/router-plugin.spec.ts -t "reuses cached page-route discovery for diagnostics on change"`
- `pnpm oxlint -c oxlint.config.ts --quiet packages/platform/src/lib/router-plugin.ts packages/platform/src/lib/router-plugin.spec.ts`

packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
benpsnyder: Addressed in `packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts`.

The duplication was real. The same `@vite-ignore` rewrite existed both:

- inside the main Angular transform path
- in the dedicated post-transform `@analogjs/vite-plugin-angular:hmr-vite-ignore` plugin

This branch removes the extra post-transform plugin and keeps a single shared rewrite path in the main Angular transform. I pulled the regex into `injectViteIgnoreForHmrMetadata(...)` so the behavior is explicit and directly testable instead of being duplicated across two transform hooks.

That keeps the HMR metadata import patch in one place and avoids running the same rewrite twice on component output.

Validation:

- `pnpm vitest run packages/vite-plugin-angular/src/lib/angular-vite-plugin.spec.ts -t "injectViteIgnoreForHmrMetadata"`
- `pnpm oxlint -c oxlint.config.ts --quiet packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts packages/vite-plugin-angular/src/lib/angular-vite-plugin.spec.ts`

changes/2026-04-04-reset-investigation/README.md
benpsnyder: ba044b0f00df1faa425ee918726d985044f101a3

package.json
benpsnyder: 135c883eee050db9257f392dcca00e2f05fc4d82

benpsnyder: I understand the concern, but I don’t currently see a concrete repo-local regression from removing `package.json.workspaces`.

This repo’s workspace and release flow is driven by:

- `pnpm-workspace.yaml`
- `pnpm`
- `nx`
- the existing CI/release scripts

That path still works end to end without `package.json.workspaces`, including install, task orchestration, and release automation. So within this repo’s actual pnpm/Nx workflow, it is not obvious what this removal would break.

The weaker version of the concern is still fair: some external npm-oriented tools or IDE features can inspect `package.json.workspaces`. But that is an ecosystem-compatibility question, not an identified break in this repository’s current workflow.

Unless we can point to a specific tool in this repo or release path that depends on `package.json.workspaces`, I don’t think there is enough evidence here to treat the removal itself as a concrete bug.

benpsnyder:
@coderabbitai
I verified the current code against the remaining review findings and only kept the fixes that were still needed.

Still-valid findings that I fixed:
- `apps/docs-app/docs/guides/migrating.md`: added the Angular version compatibility note so the `hmr` / `liveReload` section now states that Analog requires Angular v19+, and that Angular v16-v18 disable HMR at runtime with a console warning.
- `apps/tailwind-debug-app-e2e/tests/component-css-hmr.spec.ts`: changed teardown to restore the exact checked-in stylesheet contents instead of resetting to a hard-coded class.
- `apps/tailwind-debug-app/project.json`: made `serve-nitro` depend on `build` so the expected dist output exists before serving.
- `apps/tailwind-debug-app/src/app/app.component.css`: suppressed the `scss/at-rule-no-unknown` warning for Tailwind v4 `@reference`.
- `apps/tailwind-debug-app/src/app/debug/debug-stream.service.ts`: deduplicates merged stream entries by `id` before trimming so snapshots do not re-add duplicates.
- `apps/tailwind-debug-app/src/app/debug/hmr-diagnostics.ts`: wraps persistent storage writes in `try/catch` and warns without breaking in-memory diagnostics.
- `apps/tailwind-debug-app/src/app/probes/debug-stream-panel.component.ts`: Clear now clears both the live stream and the frozen paused view.
- `packages/platform/src/lib/router-plugin.ts`: add/unlink route changes now rescan all surviving `.page.ts` files for diagnostics, with regression coverage in `packages/platform/src/lib/router-plugin.spec.ts`.
- `packages/vite-plugin-angular/src/lib/component-resolvers.ts`: component metadata extraction is now limited to actual `@Component(...)` decorator objects, with regression coverage in `packages/vite-plugin-angular/src/lib/component-resolvers.spec.ts`.
- `apps/tailwind-debug-app/src/app/probes/style-probe.component.spec.ts`: renamed the test so the title matches what it actually asserts.
- `packages/platform/src/lib/content-plugin.ts`: debug reload payloads now emit `normalizedPath` instead of the raw path.
- `packages/platform/src/lib/utils/debug.spec.ts`: added an explicit assertion for the first `wrapInstancesForFileLog` argument.
- `packages/vite-plugin-angular/src/lib/stylesheet-registry.spec.ts`: added coverage for rewriting `@import url(...)` CSS imports.
- `packages/vite-plugin-angular/src/lib/angular-vite-plugin.spec.ts`: updated the behavior-doc test naming/commentary for the `hasComponent` detection nitpick.

Findings that were already resolved and did not need another fix:
- `package.json`: the `@angular/platform-server` comment was already superseded because the root dependencies now come from the workspace catalog.
- `packages/router/src/lib/provide-file-router-base.ts`: the dev-only bootstrap short-circuit had already been removed, and the metadata subscriptions are initialized during app init.

Targeted verification run on this branch:
- `pnpm vitest run packages/vite-plugin-angular/src/lib/component-resolvers.spec.ts packages/vite-plugin-angular/src/lib/stylesheet-registry.spec.ts packages/platform/src/lib/router-plugin.spec.ts packages/platform/src/lib/utils/debug.spec.ts`
- `pnpm nx test tailwind-debug-app --runInBand`
- `pnpm vitest run --config packages/vite-plugin-angular/vite.config.ts packages/vite-plugin-angular/src/lib/angular-vite-plugin.spec.ts`
- `pnpm oxlint -c oxlint.config.ts --quiet ...` on the touched TS files: `0 errors`

Follow-up commits:
- fb701b8f3 `docs: note Angular HMR version requirements`
- 9c6a2966b `fix: harden tailwind debug app HMR diagnostics`
- 779c9c8f4 `fix(platform): refresh route diagnostics on route graph changes`
- c37ff35f2 `fix(vite-plugin-angular): restrict metadata extraction to component decorators`
- bcadc13e0 `test(vite-plugin-angular): tighten resolver coverage docs`

benpsnyder:
@coderabbitai 9e0be352489cf09c7786bbd3b41792bd85987df2 adjusted the repro to match [tailwindlabs/tailwindcss#19903](https://github.com/tailwindlabs/tailwindcss/issues/19903) instead of asserting the opposite behavior.

What changed:

- `apps/tailwind-debug-app/vite.config.ts`
  - the HMR wiretap now logs both `server.ws.send` and `environment.hot.send`
  - full-reload payloads also capture a stack so the source matches the upstream report more closely
- `apps/tailwind-debug-app-e2e/tests/component-css-hmr.spec.ts`
  - renamed the test to reflect that it reproduces the Tailwind-triggered full reload
  - stopped asserting “red without reload”
  - now asserts the actual repro signal:
    - the stylesheet edit is detected
    - `environment.hot.send` emits `full-reload`
    - app state resets after reload
    - the stack includes `updateModules`
- `packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts`
  - unchanged-but-already-valid external component stylesheets are now still registered in the stylesheet registry, which keeps the Angular stylesheet HMR bookkeeping honest even when preprocessing makes no textual change

Verification:

- `CI=1 pnpm nx run tailwind-debug-app-e2e:e2e` passed
- `pnpm vitest run --config packages/vite-plugin-angular/vite.config.ts packages/vite-plugin-angular/src/lib/angular-vite-plugin-live-reload.spec.ts` passed
- `pnpm oxlint -c oxlint.config.ts --quiet ...` on touched files: `0 errors`

benpsnyder:
image

PR #2228, opened 2026-04-03
feat(platform,vite-plugin-angular,vite-plugin-nitro): add logFile option to debug system

benpsnyder:
## Summary

- Adds `logFile` option to `DebugModeOptions` so consumers can opt into writing debug output to log files in `tmp/debug/`
- Supports `true`/`'single'` (all output to `tmp/debug/analog.log`) and `'scoped'` (one file per namespace, e.g. `tmp/debug/analog.angular.hmr.log`)
- Consolidates duplicated debug infrastructure from `platform` and `vite-plugin-angular` into a new private `@analogjs/cross-utils` package
- Fixes workspace protocol drift: restores `workspace:^` in platform and router source `package.json` files that semantic-release was overwriting with concrete versions

### Usage

```typescript
analog({
  debug: { logFile: true }                                    // all scopes → tmp/debug/analog.log
  debug: { scopes: ['analog:platform'], logFile: true }       // specific scopes + file
  debug: { scopes: true, logFile: 'scoped' }                  // one file per scope
  debug: { scopes: true, mode: 'dev', logFile: true }         // mode-restricted + file
})
```

### Motivation

Agents (Claude Code, Cursor, etc.) monitoring Analog builds benefit from a stable file to read/tail. stderr output is ephemeral and hard for agents to capture, but log files in the workspace are straightforward to monitor.

The debug utilities in `platform` and `vite-plugin-angular` were byte-for-byte identical (`debug-log-file.ts`) or ~80% duplicated (`debug.ts`). Rather than maintaining two copies, shared logic now lives in `@analogjs/cross-utils` — a private internal package that gets bundled into each consumer at build time (no runtime dependency for end users, no references in published `.d.ts` files).

### Files changed

| Area | Files | Change |
|------|-------|--------|
| **cross-utils** (new) | `packages/cross-utils/` | Private shared package: `debug-log-file.ts` (truncation, ANSI stripping, `appendFileSync` wrapper), `debug-harness.ts` (`createDebugHarness` factory for `applyDebugOption`/`activateDeferredDebug`) |
| **platform** | `debug.ts`, `debug.spec.ts` | Rewritten from 185→60 lines using `createDebugHarness`; deleted `debug-log-file.ts` + spec |
| **platform** | `platform-plugin.ts` | Passes `workspaceRoot` to `applyDebugOption` |
| **platform** | `vite.config.ts`, `vite.config.lib.ts` | Resolve alias + bundle `@analogjs/cross-utils` |
| **vite-plugin-angular** | `debug.ts`, `debug.spec.ts` | Rewritten from 176→55 lines using `createDebugHarness`; deleted `debug-log-file.ts` + spec |
| **vite-plugin-angular** | `angular-vite-plugin.ts` | Passes `workspaceRoot` to `applyDebugOption` |
| **vite-plugin-angular** | `vite.config.ts`, `vite.config.lib.ts` | Resolve alias + bundle `@analogjs/cross-utils` |
| **vite-plugin-nitro** | `debug.ts`, `index.ts` | Exports `debugInstances` array for platform to wrap |
| **release infra** | `release.config.ts`, `platform/package.json`, `router/package.json` | Removes source packages from `replacementFiles`; restores `workspace:^` references |
| **workspace** | `tsconfig.base.json` | Adds `@analogjs/cross-utils` path alias |

## Test plan

- [x] `nx test cross-utils` — 42 tests pass (debug-log-file + debug-harness)
- [x] `nx test vite-plugin-angular` — 83 tests pass
- [x] `nx test platform` — 213 tests pass
- [x] `nx build platform` — builds, artifact validation passes
- [x] `nx build vite-plugin-angular` — builds, artifact validation passes
- [x] Verified no `@analogjs/cross-utils` references in published `.d.ts` files
- [ ] Manual: add `debug: { logFile: true }` to an app's `vite.config.ts`, run `pnpm dev`, verify `tmp/debug/analog.log` appears with readable output

### DIFF
--- a/packages/platform/package.json
+++ b/packages/platform/package.json
@@ -35,8 +35,8 @@
     "es-toolkit": "catalog:",
     "tinyglobby": "catalog:",
     "nitro": "catalog:",
-    "@analogjs/vite-plugin-angular": "workspace:^",
-    "@analogjs/vite-plugin-nitro": "workspace:^",
+    "@analogjs/vite-plugin-angular": "workspace:*",
+    "@analogjs/vite-plugin-nitro": "workspace:*",
     "rolldown": "catalog:",
     "obug": "catalog:",
     "vitefu": "catalog:"
--- a/packages/platform/src/lib/deps-plugin.ts
+++ b/packages/platform/src/lib/deps-plugin.ts
@@ -34,6 +34,7 @@ export function depsPlugin(options?: Options): Plugin[] {
           ssr: {
             noExternal: [
               '@analogjs/**',
+              'es-toolkit',
               'firebase/**',
               'firebase-admin/**',
               'rxfire',
--- a/packages/router/package.json
+++ b/packages/router/package.json
@@ -64,7 +64,7 @@
     "url": "https://github.com/sponsors/brandonroberts"
   },
   "peerDependencies": {
-    "@analogjs/content": "workspace:^",
+    "@analogjs/content": "workspace:*",
     "@angular/core": "^15.0.0 || ^16.0.0 || ^17.0.0 || ^18.0.0 || ^19.0.0 || ^20.0.0 || ^21.0.0",
     "@angular/platform-server": "^15.0.0 || ^16.0.0 || ^17.0.0 || ^18.0.0 || ^19.0.0 || ^20.0.0 || ^21.0.0",
     "@angular/router": "^15.0.0 || ^16.0.0 || ^17.0.0 || ^18.0.0 || ^19.0.0 || ^20.0.0 || ^21.0.0",
@@ -91,7 +91,7 @@
     "tslib": "catalog:"
   },
   "devDependencies": {
-    "@analogjs/vite-plugin-angular": "workspace:^"
+    "@analogjs/vite-plugin-angular": "workspace:*"
   },
   "ng-update": {
     "packageGroup": [
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -1014,10 +1014,10 @@ importers:
   packages/platform:
     dependencies:
       '@analogjs/vite-plugin-angular':
-        specifier: workspace:^
+        specifier: workspace:*
         version: link:../vite-plugin-angular
       '@analogjs/vite-plugin-nitro':
-        specifier: workspace:^
+        specifier: workspace:*
         version: link:../vite-plugin-nitro
       '@nx/angular':
         specifier: catalog:peerCompat
@@ -1074,7 +1074,7 @@ importers:
   packages/router:
     dependencies:
       '@analogjs/content':
-        specifier: workspace:^
+        specifier: workspace:*
         version: link:../content
       '@angular/core':
         specifier: ^15.0.0 || ^16.0.0 || ^17.0.0 || ^18.0.0 || ^19.0.0 || ^20.0.0 || ^21.0.0
@@ -1102,7 +1102,7 @@ importers:
         version: 2.8.1
     devDependencies:
       '@analogjs/vite-plugin-angular':
-        specifier: workspace:^
+        specifier: workspace:*
         version: link:../vite-plugin-angular
 
   packages/storybook-angular:
--- a/tools/build/resolve-catalogs.ts
+++ b/tools/build/resolve-catalogs.ts
@@ -9,7 +9,7 @@
  * YAML and substituting every `catalog:` / `catalog:<name>` reference with
  * the resolved version.
  *
- * Also resolves `workspace:` protocol references (`workspace:*`, `workspace:^`,
+ * Also resolves `workspace:` protocol references (`workspace:*`, `workspace:*`,
  * `workspace:~`) by reading the target package's version from the workspace.
  * pnpm normally resolves these during `pnpm publish`, but the smoke test
  * pipeline uses `npm pack` which does not understand the workspace protocol.
@@ -202,9 +202,9 @@ function resolveDependencyMap(
     } else if (version.startsWith('workspace:')) {
       // Resolve workspace: protocol the same way pnpm does at publish time:
       //   workspace:*  → exact version  (e.g. "3.0.0-alpha.18")
-      //   workspace:^  → caret range    (e.g. "^3.0.0-alpha.18")
+      //   workspace:*  → caret range    (e.g. "^3.0.0-alpha.18")
       //   workspace:~  → tilde range    (e.g. "~3.0.0-alpha.18")
-      //   workspace:^1.0.0 → "^1.0.0"  (pass-through semver range)
+      //   workspace:*1.0.0 → "^1.0.0"  (pass-through semver range)
       const specifier = version.slice('workspace:'.length);
       const targetVersion = resolveWorkspacePackageVersion(name, workspaceRoot);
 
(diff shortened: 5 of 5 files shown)


benpsnyder:
During agentic development, there are cases where the scoped debug isn't enough within one terminal, and we need streamed log output that agents can tail/monitor.
image

benpsnyder:
@coderabbitai 
Fixed (2):
- ensureTruncated now wraps mkdirSync/writeFileSync in try/catch so a read-only filesystem won't crash the build (falls through to append mode)
- Test now asserts appendFileSync receives the formatted string "hello world" instead of just checking it was called
- Added test covering writeFileSync throwing during truncation

Skipped (not applicable or already handled):
- TRUNCATED_KEY "global boolean" — already a Set<string> keyed by filePath; per-path tracking works correctly
- WRAPPED_KEY per-filePath re-wrapping — current per-instance design is correct; applyDebugOption is called once at init
- POSIX path assertions — Vite normalizes to POSIX; not relevant for this build-tool project
- Missing mockClear — already fixed in the cross-utils spec
- NX_WORKSPACE_ROOT fallback tests — trivial ?? chain, not worth the test complexity
- angular-vite-plugin.ts comment/implementation mismatch — pre-existing code not introduced by this PR; the comments describe planned external stylesheet preprocessing but the implementation only maps raw paths

benpsnyder:
## Confirming this fixes a real-world Bun + Tailwind CSS v4 issue

I hit the exact drift problem this PR fixes while working on Tailwind CSS v4 `@reference` injection with `useAngularCompilationAPI: true` in a Bun workspace.

### Setup

Our `package.json` uses Bun's `catalog:` with `file:` references to a local Analog fork:

```json
"@analogjs/vite-plugin-angular": "file:../../@benpsnyder/analog/packages/vite-plugin-angular/dist"
```

### The problem

After building `vite-plugin-angular` locally with fixes for external stylesheet preprocessing (reading CSS from disk and injecting `@reference` for Tailwind v4 prefix utilities), **the dev server kept loading the stale npm-published copy** instead of our patched build.

Debugging revealed Bun resolved **5 separate copies** of `angular-vite-plugin.js`:

| Copy | Inode | Source |
|------|-------|--------|
| `@file+...+42683fd4/` | 269572070 | ✅ Our `file:` build (44KB, with fixes) |
| `@file+...+92401b28/` | 269572070 | ✅ Same hard link |
| `@3.0.0-alpha.22+42683fd4/` | 275065715 | ❌ npm (34KB, no fixes) |
| `@3.0.0-alpha.23+92401b28/` | 275736668 | ❌ npm (38KB, no fixes) |
| `@3.0.0-alpha.22+92401b28/` | 274937921 | ❌ npm (34KB, no fixes) |

The npm copies existed because `packages/platform/package.json` declared:

```json
"@analogjs/vite-plugin-angular": "^3.0.0-alpha.23"
```

Bun saw this concrete version spec from `@analogjs/platform` and resolved it from npm, **ignoring the `file:` reference** in the root catalog. The runtime loaded the npm copy (which lacked our Tailwind preprocessing code), so `@reference` was never injected and every component CSS failed with:

```
Cannot apply utility class 'sa:grid' because the 'sa' variant does not exist
```

### Why this PR fixes it

The two-line removal from `release.config.ts` is the key fix — it prevents semantic-release from overwriting `workspace:^` with concrete versions in the source `package.json` files after each release. Combined with restoring `workspace:^` in `platform/package.json` and `router/package.json`, this ensures that local `file:` builds resolve transitively through the workspace instead of pulling stale copies from npm.

### The Tailwind context

For anyone curious about the underlying work: Angular's Compilation API with `externalRuntimeStyles` does **not** call `transformStylesheet` for external `styleUrls` (there's an explicit assert in `angular-host.js`). This means `buildStylePreprocessor` never injects `@reference` for component CSS files, which breaks `@tailwindcss/vite` prefix resolution. The fix reads each external stylesheet from `compilationResult.externalStylesheets`, runs the preprocessor, resolves relative `@import` paths to absolute (since the virtual hash-based module ID has no meaningful directory), and stores the result in `inlineComponentStyles`. None of this was visible until the drift was fixed because the runtime was loading the npm copy without these changes.

Node v25 / Bun 1.x / Angular 21 / Tailwind CSS v4 / Analog alpha.23

brandonroberts:
**API surface concern:** `debugInstances` is now exported from `@analogjs/vite-plugin-nitro`'s public `index.ts`, which exposes an internal `Debugger[]` to consumers. It's only used by `platform` internally — consider marking it `@internal` in the JSDoc or moving the cross-package wiring so it doesn't need to be on the public export surface.

benpsnyder:
@coderabbitai

Inline comments (verified and fixed):

1. debug-harness.ts — deferred wrapping eagerness: Confirmed the bug. extractLogFile + wrapping ran immediately even for mode-gated entries. Restructured so
applyEntry carries logFile and root into the pending queue, and activateDeferredDebug calls installFileWrappers only when the mode actually matches. Added 3
regression tests.
2. debug-log-file.ts — WRAPPED_KEY as boolean: Confirmed. Changed WRAPPED_KEY from storing true to storing the active filePath. When a new path differs, the
wrapper is replaced using the preserved __analogOriginalLog. Added a "re-wraps when target path changes" test.

Outside diff (verified and fixed):

3. platform/vite.config.ts — external predicate: Confirmed. Added early false returns for @analogjs/cross-utils and @analogjs/cross-utils/* so the bare
specifier isn't externalized before alias resolution kicks in.

Nitpicks (verified and fixed):

4. cross-utils/project.json — cache comment: JSON doesn't supports and no other project.json in the repo uses them, so I left this as-is.
5. debug-harness.spec.ts — regression test: Added "does not wrap when deferred mode does not match" + two positive-path tests for deferred single and scoped
wrapping.

CI lint fix:

6. debug-log-file.ts — no-control-regex: Added // eslint-disable-next-line no-control-regex above the ANSI stripping regex since the control characters are
intentional.

benpsnyder:
> **API surface concern:** `debugInstances` is now exported from `@analogjs/vite-plugin-nitro`'s public `index.ts`, which exposes an internal `Debugger[]` to consumers. It's only used by `platform` internally — consider marking it `@internal` in the JSDoc or moving the cross-package wiring so it doesn't need to be on the public export surface.

## Problem

`debugInstances` was exported from `@analogjs/vite-plugin-nitro`'s public barrel (`index.ts`), leaking an internal `Debugger[]` array to consumers even though it's only used by `@analogjs/platform` for debug harness wiring.

## Fix

Moved the cross-package wiring off the public export surface via a `./internal` subpath export:

1. `packages/vite-plugin-nitro/src/index.ts` — removed `debugInstances` re-export
2. `packages/vite-plugin-nitro/package.json` — added `"./internal"` subpath pointing to `dist/src/lib/utils/debug.{d.ts,js}`
3. `tsconfig.base.json` — added path alias for `@analogjs/vite-plugin-nitro/internal`
4. `packages/platform/tsconfig.lib.json` — added matching path alias
5. `packages/platform/src/lib/utils/debug.ts` — import now uses `@analogjs/vite-plugin-nitro/internal`
6. Test mocks — updated to mock the new import path

The public `.` export is now clean — only the Nitro plugin default export and option types. The `debug.js` file still gets emitted in the dist (it's in the module graph via `vite-plugin-nitro.ts`), so the `./internal` subpath works without needing a separate build entry.

benpsnyder:
> **API surface concern:** `debugInstances` is now exported from `@analogjs/vite-plugin-nitro`'s public `index.ts`, which exposes an internal `Debugger[]` to consumers. It's only used by `platform` internally — consider marking it `@internal` in the JSDoc or moving the cross-package wiring so it doesn't need to be on the public export surface.

All set to merge

benpsnyder:
Per maintainer feedback, removed the shared `@analogjs/cross-utils` library and copied the debug harness and log-file utilities directly into each consuming package (`platform` and `vite-plugin-angular`).

Each package now owns its own copy of `debug-harness.ts`, `debug-log-file.ts`, and their specs under `src/lib/utils/`. This eliminates the cross-package dependency, the extra build target, and the `./internal` subpath export that was needed to keep `debugInstances` off the public API surface.

See commit 96b77bec.

benpsnyder:
image

PR #2239, opened 2026-04-05
feat(platform): add shiki skipLangs option for analog v3

### LINKED ISSUE #2029: Ability to exclude Mermaid blocks from Shiki to avoid OOM in CI environments

create-analog

### Summary

When using Analog content routing with Markdown, Shiki syntax highlighting, and Mermaid diagrams, there is currently no supported way to exclude Mermaid blocks from the Shiki pipeline. In CI environments with limited memory (e.g. Cloudflare Pages), this leads to consistent out-of-memory (OOM) failures during build/SSR.

This appears to be a design/API limitation combined with incomplete documentation rather than a misuse or a runtime bug.

___

### Context

- 	Analog.js content routing
- 	Markdown content
- 	Shiki enabled as syntax highlighter
- 	Mermaid used for diagrams in Markdown
- 	Build executed via vite build
- 	CI environment: Cloudflare Pages (~2GB heap)

___

### Problem

To support Mermaid blocks in Markdown, the documentation suggests adding:

```ts
shikiOptions: {
  highlighter: {
    additionalLangs: ['mermaid'],
  },
}
```

### However:

- Including mermaid in Shiki causes high memory usage and frequently results in OOM during CI builds.
- Removing mermaid from Shiki causes the build to fail with:

``Language `mermaid` not found``

because marked-shiki attempts to highlight all fenced code blocks, and Analog does not currently expose a way to intercept or exclude specific languages before Shiki runs.

This creates a dead-end configuration:

- 	Mermaid included in Shiki → OOM in CI
- 	Mermaid excluded from Shiki → build error
- 	No supported hook or API to selectively bypass Shiki for Mermaid blocks

### Observations

- This is not specific to Analog alone, but Analog’s current API does not expose:
   - 	a way to exclude specific fenced languages from Shiki
   - 	a way to mark certain blocks as “raw” or “client-only”
- WithMarkdownRenderer does not support marked extensions or token filters.
- The documentation shows additionalLangs: ['mermaid'] as a valid example, but does not warn about the memory implications in constrained CI environments.
Expected / Requested Feature

One of the following would resolve the issue cleanly:

- A supported way to exclude specific languages (e.g. mermaid) from the Shiki pipeline.
- A configuration option such as:
  -  skipLanguages: ['mermaid']
  - rawLanguages: ['mermaid']
  - clientOnlyLanguages: ['mermaid']
- Or explicit documentation recommending client-side-only Mermaid rendering when using Shiki in SSR/CI environments.

### Why this matters

This affects:

- 	Content-heavy sites (blogs, documentation)
- 	DevRel / educational use cases
- 	Cloudflare Pages and other CI environments with limited memory

Without a supported approach, users are forced to choose between:

- unstable builds
- undocumented hacks
- or removing Mermaid from Markdown entirely

### Conclusion

This does not appear to be a bug, but rather a design limitation and documentation gap.
Clarifying or extending the API around Shiki + Mermaid would significantly improve the developer experience for real-world content use cases.

Happy to help further with testing or validation if needed.

At the moment, the only stable workaround is to avoid using Mermaid in the Markdown + Shiki pipeline altogether.

- [ ] Yes
- [ ] No

benpsnyder:
## PR Checklist

Closes analogjs/analog#2029

## Affected scope

- Primary scope: `platform`
- Secondary scopes:

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

## What is the new behavior?

This adds `shikiOptions.highlighter.skipLangs` to the platform content Shiki integration.

When a fenced code block language is listed in `skipLangs`, the block bypasses Shiki and is rendered as a plain fenced code block instead of going through Shiki's highlighter pipeline.

That gives users a supported way to exclude languages like `mermaid` from Shiki while still keeping Shiki enabled for the rest of the markdown pipeline. This follows the maintainer direction in the issue thread that `skipLangs` looked like the right option.

## Test plan

- [ ] `nx format:check`
- [x] `pnpm build`
- [x] `pnpm test`
- [ ] Manual verification

Commands run:

- `pnpm nx test platform --runTestsByPath packages/platform/src/lib/content/shiki/index.spec.ts`
- `pnpm nx build platform`

- [ ] Yes
- [x] No

## Other information

The behavior is intentionally narrow:
- skipped languages are rendered as plain fenced code blocks
- existing Mermaid handling stays unchanged when `mermaid` is not listed in `skipLangs`

### DIFF
--- a/apps/docs-app/docs/features/routing/content.md
+++ b/apps/docs-app/docs/features/routing/content.md
@@ -197,11 +197,11 @@ export default defineConfig({
         shikiOptions: {
           highlight: {
             // alternate theme
-            theme: 'ayu-dark'
-          }
+            theme: 'ayu-dark',
+          },
           highlighter: {
-             // add more languages
-            additionalLangs: ['mermaid'],
+            // add more languages for Shiki itself
+            additionalLangs: ['diff'],
           },
         },
       },
@@ -209,6 +209,31 @@ export default defineConfig({
   ],
 });
 ```
+
+For Mermaid-heavy content, keep the existing `loadMermaid` runtime path and skip Mermaid grammar loading in Shiki to avoid unnecessary server-side highlighting work in constrained CI environments:
+
+```ts
+import { defineConfig } from 'vite';
+import analog from '@analogjs/platform';
+
+export default defineConfig({
+  plugins: [
+    analog({
+      content: {
+        highlighter: 'shiki',
+        shikiOptions: {
+          highlighter: {
+            additionalLangs: ['mermaid'],
+            skipLangs: ['mermaid'],
+          },
+        },
+      },
+    }),
+  ],
+});
+```
+
+With `skipLangs: ['mermaid']`, Analog keeps Mermaid blocks on the existing `<pre class="mermaid">` path for `loadMermaid`, while Shiki skips loading and tokenizing the Mermaid grammar.
 
 By default, `shikiOptions` has the following options.
 
--- a/packages/platform/src/lib/content/shiki/index.spec.ts
+++ b/packages/platform/src/lib/content/shiki/index.spec.ts
@@ -1,0 +1,116 @@
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+
+const createHighlighterMock = vi.fn(async () => ({
+  codeToHtml: vi.fn((_code: string, options: { lang: string }) => {
+    return `<pre class="shiki" data-lang="${options.lang}"></pre>`;
+  }),
+}));
+
+vi.mock('shiki', () => ({
+  createHighlighter: createHighlighterMock,
+}));
+
+vi.mock('marked-shiki', () => ({
+  default: (options: unknown) => options,
+}));
+
+let getShikiHighlighter: typeof import('./index.js').getShikiHighlighter;
+
+beforeEach(async () => {
+  vi.resetModules();
+  createHighlighterMock.mockClear();
+  ({ getShikiHighlighter } = await import('./index.js'));
+});
+
+describe('getShikiHighlighter', () => {
+  it('does not load skipped languages into shiki and preserves the mermaid render path', async () => {
+    const highlighter = getShikiHighlighter({
+      highlighter: {
+        additionalLangs: ['mermaid'],
+        skipLangs: ['mermaid'],
+      },
+    });
+
+    expect(createHighlighterMock).toHaveBeenCalledWith(
+      expect.objectContaining({
+        langs: expect.not.arrayContaining(['mermaid']),
+      }),
+    );
+
+    const extension = highlighter.getHighlightExtension() as {
+      highlight: (
+        code: string,
+        lang: string,
+        props: string[],
+      ) => Promise<string>;
+    };
+
+    await expect(extension.highlight('graph TD;', 'mermaid', [])).resolves.toBe(
+      '<pre class="mermaid">graph TD;</pre>',
+    );
+  });
+
+  it('returns plain fenced code blocks for skipped non-mermaid languages', async () => {
+    const highlighter = getShikiHighlighter({
+      highlighter: {
+        additionalLangs: ['yaml'],
+        skipLangs: ['yaml'],
+      },
+    });
+
+    const extension = highlighter.getHighlightExtension() as {
+      highlight: (
+        code: string,
+        lang: string,
+        props: string[],
+      ) => Promise<string>;
+    };
+
+    await expect(extension.highlight('name: analog', 'yaml', [])).resolves.toBe(
+      '<pre class="language-yaml"><code class="language-yaml">name: analog</code></pre>',
+    );
+  });
+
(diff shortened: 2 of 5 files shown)


benpsnyder:
I checked this against [analogjs/analog#2029](https://github.com/analogjs/analog/issues/2029), the issue discussion, and the current implementation in this branch. `skipLangs` is the right direction, but a few things still look incomplete before this fully resolves the issue:

1. The current implementation still pushes `additionalLangs` into Shiki's `langs`, and the new test explicitly expects `mermaid` to remain there. If the goal is to avoid the Mermaid grammar being loaded by Shiki to prevent the build-time OOM reported in [analogjs/analog#2029](https://github.com/analogjs/analog/issues/2029), this likely still needs adjustment.

2. The skipped-language output shape does not match Analog's existing Mermaid path. Right now skipped Mermaid returns `<pre class="language-mermaid"><code class="language-mermaid">...`, while the current Mermaid handling expects `<pre class="mermaid">...`. That means `skipLangs: ['mermaid']` would bypass Shiki, but it also appears to bypass the client-side Mermaid rendering path.

3. The issue explicitly called out a documentation gap as well as the missing API. The docs still show `additionalLangs: ['mermaid']` in the Shiki example and do not mention `skipLangs` or the constrained-memory Mermaid guidance, so the documented user path is still the one that can trigger the original problem.

I did verify that the targeted tests and build pass locally:
- `pnpm nx test platform --runTestsByPath packages/platform/src/lib/content/shiki/index.spec.ts`
- `pnpm nx build platform`

So this looks close, but I don't think it fully closes [analogjs/analog#2029](https://github.com/analogjs/analog/issues/2029) yet without the behavior/docs follow-up above.

benpsnyder:
Follow-up pushed on `fix/2029-mermaid-shiki-oom`.

Addressed the skipped-language HTML injection path in `packages/platform/src/lib/content/shiki/shiki-highlighter.ts` by escaping the fallback `<pre><code>` output before interpolation, while leaving normal `codeToHtml()` handling unchanged.

Also added a regression test in `packages/platform/src/lib/content/shiki/index.spec.ts` that verifies skipped code containing HTML characters is rendered as escaped text.

Verification re-run:
- `pnpm nx test platform --runTestsByPath packages/platform/src/lib/content/shiki/index.spec.ts`

brandonroberts:
This needs to be validated against the reproduction that it actually fixes the OOM issue.

brandonroberts:
Beta branch

benpsnyder:
[On alpha] Validated against the reproduction at [dmorfav/analyzing_oom](https://github.com/dmorfav/analyzing_oom) (linked from [analogjs/analog#2029 (comment)](https://github.com/analogjs/analog/issues/2029#issuecomment-3706276547)).

## Test harness

Exercised `@analogjs/platform`'s real `getShikiHighlighter()` + `getHighlightExtension()` through a `marked` instance against two module trees, same config in both runs (`additionalLangs: ['mermaid']`, plus `skipLangs: ['mermaid']` in the fixed run):

- **Baseline:** `npm pack @analogjs/platform@3.0.0-alpha.29` — pristine published alpha, no fix.
- **Fixed:** `packages/platform/dist/` built from `fix/2029-mermaid-shiki-oom`, swapped into the reproduction repo's `node_modules/@analogjs/platform`.

## Results

| Assertion | Baseline | Fixed |
|---|---|---|
| `shiki.getLoadedLanguages().includes('mermaid')` — OOM root cause | **true** | **false** |
| Loaded-lang count | 27 (includes `mermaid` + `mmd`) | 25 (neither) |
| Mermaid fence renders as `<pre class="mermaid">…</pre>` | yes | yes — **byte-identical** (verified by `diff`) |
| `ts` fence receives full shiki highlighting | yes | yes |

Baseline loaded langs included `mermaid`, `mmd`. Fixed run stripped both — `createHighlighter(...)` never loads the mermaid TextMate grammar (or its embedded grammars) into the oniguruma registry, eliminating the OOM path described in the issue.

## What this proves

1. **Root cause confirmed** — the published alpha loads mermaid's TextMate grammar into shiki, matching the issue's diagnosis.
2. **Fix works** — `skipLangs: ['mermaid']` filters mermaid out of `langs` before `createHighlighter(...)`, so shiki never loads it.
3. **No regression** — mermaid fences still emit `<pre class="mermaid">…</pre>` (downstream mermaid rendering preserved) and non-skipped languages still get full shiki highlighting. Output diff for the mermaid block is byte-identical between baseline and fixed.

Full Vite build on the reproduction completes successfully with the fix applied.

benpsnyder:
6114608a8374e4eae804c4985f089e59c99263f3 is what makes it work converting from `alpha` to `beta` base

benpsnyder:
@brandonroberts this got very "un-fun" for `beta` branch so I am reverting to target this for `alpha`. I opened a new PR that can be worked on for targeting `beta` here https://github.com/analogjs/analog/pull/2282

brandonroberts (approved):
LGTM

Issue #2068, opened 2026-02-17
[storybook] vite builder does not build scss


purplefrizzel:
Not sure if this is a bug. We have a project which uses Nx, with quite a lot of libraries. We have already migrated the build process of the app to Vite, and we are attempting to migrate our two Storybook instances over to Vite as well because Webpack is just too slow for our use case. We tried using the Storybook vite builder, but had issues with it, which we believe is down to the use of Nx. So we are looking at using `@analogjs/storybook-angular` instead, we have managed to get a build working, but for some reason our scss is not building. We have a large shared lib of scss, and most of our components have their own scss files too.  We have looked at multiple suggestions online, but non of these worked. The Storybook config is quite simple, we are not doing anything out of the ordinary. We have one project level config which shares some properties with the two library level configs, the project.json is using the setup which is shown here https://analogjs.org/docs/integrations/storybook#setting-up-css and we have also followed this https://analogjs.org/docs/integrations/storybook#with-nx.

It would be ideal if we could get some support on this issue, as we would like to migrate over to Vite for Storybook, and the analogjs implementation is almost working we just need styles to build correctly. Any help will be greatly appreciated.

If you need some more details just ask, I can send over our main.ts files, and other configs if it will help. We do use some storybook addons, but I can't see them being an issue here, namely `@storybook/addon-docs`, `@storybook/addon-a11y`, `@storybook/addon-themes`. We don't really use a lot of third party styles, this is our config in the project.json for styles: 

```json
"styles": [
 "libs/shared/ui/styles/shared-ui.scss",
  "libs/shared/ui/.storybook/storybook.scss",
  "node_modules/@angular/material/prebuilt-themes/deeppurple-amber.css",
  "node_modules/katex/dist/katex.css",
  "node_modules/flag-icons/css/flag-icons.min.css"
],
"stylePreprocessorOptions": {
   "loadPaths": ["libs/shared/ui/styles"]
}
```

Other than this styles section, our project.json file is pretty much the same as the docs. Our top-level main.ts file is using `nxViteTsPaths` and `angular({ tsconfig: '.storybook/tsconfig.json', }),` as suggested in the docs.

brandonroberts:
@purplefrizzel glad to try to get your issue resolved. If you can reproduce it in a small GitHub repo that would help. Any additional context would help also in terms of warnings or errors.

prclm:
I had issus with importing styles from node packages as well. 
[...]

PR #2243, opened 2026-04-05
feat(storybook-angular): add scoped style diagnostics for Nx SCSS setup

benpsnyder:
## PR Checklist

Refs analogjs/analog#2068

## Affected scope

- Primary scope: storybook-angular
- Secondary scopes: docs

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

## What is the new behavior?

This PR does not claim a full runtime fix for analogjs/analog#2068 because the issue thread still does not contain a minimal reproduction.

Instead, it hardens the current Storybook preset around the exact cases discussed in the issue:
- adds regression coverage for Nx workspace-relative global SCSS files loaded through Storybook builder `styles`
- adds regression coverage for SCSS `loadPaths` resolution from the workspace root
- adds regression coverage for bare package CSS imports such as `@angular/material`, `katex`, and `flag-icons`
- adds scoped debug logging for Storybook style resolution via `DEBUG=analog:storybook:styles,analog:storybook:styles:v`
- updates the Storybook docs to clarify the difference between global `styles`, Sass `loadPaths`, and `nxViteTsPaths()`

## Test plan

- [ ] `nx format:check`
- [ ] `pnpm build`
- [ ] `pnpm test`
- [x] Manual verification

Manual verification:
- `pnpm nx test storybook-angular --runTestsByPath packages/storybook-angular/src/lib/preset.spec.ts`

- [ ] Yes
- [x] No

## Other information

Issue context reviewed before writing the tests/docs/debugging support:
- original issue body for analogjs/analog#2068
- maintainer request for a minimal reproduction
- follow-up comments about global styles vs component styles
- comment recommending bare package CSS imports instead of `node_modules/...` paths

The new scoped logging is intended to make the next issue follow-up actionable without requiring users to patch local code before sharing diagnostics.


benpsnyder:
@brandonroberts ready

The new Storybook docs still tell Nx users to add `nxViteTsPaths()`, but that no longer matches the current workspace setup after the package-native resolution change.

For current Analog Nx workspaces, normal workspace library imports already resolve without `nxViteTsPaths()`. Keeping this guidance here will push users back toward config we just removed. I think these docs should instead say:

- you usually do **not** need `nxViteTsPaths()` in Nx Storybook
- only add it if the workspace still relies on custom `compilerOptions.paths` aliases outside the package-native resolution flow

The affected sections are:

- `packages/storybook-angular/README.md`
- `apps/docs-app/docs/integrations/storybook/index.md`

packages/storybook-angular/src/lib/debug.ts
```diff
@@ -0,0 +1,4 @@
+import { createDebug } from 'obug';
+
+export const debugStyles = createDebug('analog:storybook:styles');
+export const debugStylesV = createDebug('analog:storybook:styles:v');
```
brandonroberts: Let's use `analog-storybook:styles`

`styles:v` should be more descriptive or just combined in the other debug flag

apps/docs-app/docs/integrations/storybook/index.md
```diff
@@ -319,7 +348,9 @@ export default config;
 
 ### With Nx
 
-For Nx workspaces, import and use the `nxViteTsPaths` plugin from the `@nx/vite` package. Add the plugin to the `plugins` array in the `.storybook/main.ts`.
+For current Analog Nx workspaces, package-native workspace resolution already handles normal workspace library imports in Storybook. You usually do not need to add `nxViteTsPaths()` just to resolve workspace packages.
```
brandonroberts: What's package-native workspace resolution? "Analog Nx workspaces" shouldn't be any different than Nx workspaces

brandonroberts (changes requested):

packages/storybook-angular/src/lib/debug.ts
benpsnyder: Updated this to use `analog-storybook:styles` and removed the separate `styles:v` namespace.

brandonroberts (approved):

Issue #2222, opened 2026-04-03
Vitest 4.0.5 broke Angular tests’ isolation when running with analog on github ci cd actions


thanh-nguyen-codeleap:
### Please provide the environment you discovered this bug in.

Node           : 22.22.1
OS             : darwin-arm64
Native Target  : ubuntu
pnpm           : 10.27.0
nx (global)        : 22.4.5
nx                 : 22.5.4
@nx/js             : 22.5.4
@nx/eslint         : 22.5.4
@nx/workspace      : 22.5.4
@nx/angular        : 22.5.4
@nx/eslint-plugin  : 22.5.4
@nx/storybook      : 22.5.4
@nx/vite           : 22.5.4
@nx/web            : 22.5.4
typescript         : 5.9.3
@nx/eslint/plugin
@analogjs/storybook-angular   : 2.4.0
@analogjs/vite-plugin-angular : 2.4.0
@analogjs/vitest-angular      : 2.4.0
@jsverse/transloco            : 8.2.1
@ngrx/component-store         : 21.0.1
@ngxpert/hot-toast            : 6.1.0
@storybook/angular            : 10.2.16
angular-eslint                : 21.3.1

### Which area/package is the issue in?

vitest-angular

### Description

After upgrading to vitest to versions above 4.0.4, tests started to run inconsistently on github ci cd action while still being successful on my local machine. The issue seems to be related to this issue which analog failed to isolate the test 
https://github.com/vitest-dev/vitest/issues/8944

### Please provide the exception or error you saw

```true
src/lib/ui/profile/compact-profile.component.spec.ts > CompactProfileComponent > should create
  TypeError: rootElement.setAttribute is not a function
   ❯ DOMTestComponentRenderer.insertRootElement ../node_modules/.pnpm/darwin_arm64-fastbuild-ST-fdfa778d11ba/bin/packages/platform-browser/testing/src/dom_test_component_renderer.ts:24:11
   ❯ _TestBedImpl.createComponent ../node_modules/.pnpm/darwin_arm64-fastbuild-ST-fdfa778d11ba/bin/packages/core/testing/src/test_bed.ts:688:27
   ❯ Function.createComponent ../node_modules/.pnpm/darwin_arm64-fastbuild-ST-fdfa778d11ba/bin/packages/core/testing/src/test_bed.ts:414:33
   ❯ src/lib/ui/profile/compact-profile.component.spec.ts:14:23
       12|     }).compileComponents();
       13|
[...]

PR #2244, opened 2026-04-05
fix(vitest-angular): reinitialize testbed per setup call

benpsnyder:
## PR Checklist

Closes analogjs/analog#2222

## Affected scope

- Primary scope: `vitest-angular`
- Secondary scopes: `tests`, `docs`

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

## What is the new behavior?

This PR hardens `setupTestBed()` against Vitest project-order leakage inside a shared worker/process and adds a supported debugging path for follow-up reports.

Before this change, `setupTestBed()` initialized Angular's test environment only once per process. If another setup file called `setupTestBed()` later with different options such as additional providers, those options were silently ignored because the global singleton guard skipped reinitialization.

This PR changes `setupTestBed()` to reset the current Angular test environment before reinitializing it on later calls. That keeps each setup file's configuration authoritative instead of depending on whichever project initialized first.

The PR also:
- adds a focused regression project under `tests/vitest-angular/src/reconfigure-test-bed` that verifies repeated `setupTestBed()` calls in one Vitest process actually reconfigure the environment
- documents that each setup file can call `setupTestBed()` independently without leaking providers/options across Vitest projects
- adds scoped debug tokens `analog:vitest:testbed` and `analog:vitest:testbed:v` so future reports can capture setup/reconfigure behavior without patching local code

## Test plan

- [ ] `nx format:check`
- [x] `pnpm nx build vitest-angular`
- [ ] `pnpm test`
- [x] Manual verification

Manual verification:
- `pnpm nx build vitest-angular`
- `pnpm exec vitest run --config tests/vitest-angular/vitest.config.ts --project reconfigure-test-bed`
- `pnpm exec vitest run --config tests/vitest-angular/vitest.config.ts`

- [ ] Yes
- [x] No

## Other information

The existing `reset-test-bed-between-tests` fixture still passes. The new regression covers the different failure mode here: later `setupTestBed()` calls with different configuration being ignored in the same Vitest process.

For future issue reports, users can now run Vitest with:
- `DEBUG=analog:vitest:testbed pnpm exec vitest run ...`
- `DEBUG=analog:vitest:testbed:v pnpm exec vitest run ...`


brandonroberts:
This needs to be validated against the Angular TestBed source code for compatibility and performance impact. I don't think this solution is a trivial one.

benpsnyder:
> This needs to be validated against the Angular TestBed source code for compatibility and performance impact. I don't think this solution is a trivial one.

I validated this against Angular's `TestBed` source and measured the specific reconfigure path before keeping the change.

- `resetTestEnvironment()` in Angular already calls `resetTestingModule()` internally, so the previous implementation was doing that reset work twice before re-initializing the environment.
- Angular also uses the `resetTestEnvironment()` -> `initTestEnvironment(...)` sequence in its own browser testing when it needs to switch environments, so the re-init approach itself is compatible with Angular's lifecycle.

Compatibility:

- `pnpm nx test tests-vitest-angular` passed: `6` test files, `10` tests.
- Observed end-to-end runtime for that focused validation run on my machine: `2.97s`.

Performance:

- I microbenchmarked the exact reconfigure cycle this change affects: `setupTestBed(); setupTestBed({ providers: [...] }); TestBed.inject(...)`.
- Environment: Node `v24.14.1`, Angular `21.2.8`, JSDOM-backed browser globals.
- Method: `50` warmups, `400` measured iterations per variant, plus `25` compatibility-only iterations per variant.

Results:

- Previous double-reset path:
  mean `0.2557ms`, median `0.2265ms`, p95 `0.3410ms`, stddev `0.2194ms`.
- Current single-reset path:
  mean `0.2117ms`, median `0.1921ms`, p95 `0.2610ms`, stddev `0.1298ms`.
- Delta on the measured reconfigure path:
  `0.0440ms` lower mean latency per cycle, `17.2%` faster on average.

So the change is source-backed for compatibility, and the measured impact on the code path it changes is a small but real performance improvement rather than a regression.

benpsnyder:
@coderabbitai I verified this against the current code and did not apply the suggested fix.

- The observation is real: `packages/vitest-angular/setup-testbed.ts` always registers `beforeEach(getCleanupHook(false))` and `afterEach(getCleanupHook(true))`, and `tests/vitest-angular/src/reconfigure-test-bed/test-setup.ts` calls `setupTestBed()` twice, so that setup file does register duplicate cleanup handlers.
- The proposed fix is not safe as written. Reusing `ANGULAR_TESTBED_SETUP` as a global guard for hook registration suppresses cleanup hook registration in later setup files, because that symbol also tracks successful TestBed initialization.
- When I validated that guarded variant earlier, the later `reset-test-bed-between-tests` browser project failed with `Cannot configure the test module when the test module has already been instantiated`, which shows cleanup was skipped where it was still needed.
- Because of that, I did not change the code for this finding.

Verification:

- Current code still passes `pnpm nx test tests-vitest-angular`: `6` test files, `10` tests.

Conclusion:

- This finding is partially correct about duplicate registration in a repeated `setupTestBed()` call.
- The requested fix should not be applied in its proposed form.
- Any deduping needs a separate guard that does not share state with the TestBed initialization flag.

brandonroberts (changes requested):
Having one test does not validate that the TestBed doesn't leak across tests. This needs to be validated as a solution for Angular. Reconfiguring the TestBed environment everytime is not a proposed solution in Angular

benpsnyder:
> Having one test does not validate that the TestBed doesn't leak across tests. This needs to be validated as a solution for Angular. Reconfiguring the TestBed environment everytime is not a proposed solution in Angular

I am a bit confused as to how to proceed. May I recommend ... let's take this to discord chat and maybe put this up against our internal repo and run a batter of tests.

brandonroberts:
Thanks @benpsnyder. Closing this — #2222 turned out to be a different root cause than the `setupTestBed()` reinitialization this PR addresses.

The isolation leak was the plugin forcing `test.pool` to `vmThreads` even in browser mode, which disables per-file isolation (`isolate` has no effect under VM pools, and browser isolation inherits that since Vitest 4.0.7). Fixed in #2423, now merged.

---

Issue #2227, opened 2026-04-03
[AnalogJS v3] RFC: First-Class Design Token Support


benpsnyder:
## Problem

Every Angular component library ships its own CSS variable naming convention. Teams that adopt one system get locked into its token schema, and teams that combine libraries face incompatible theming layers they must reconcile by hand.

Consider two popular choices for Analog + Tailwind projects:

| Semantic Concept | daisyUI | spartan.ng | Angular Material 3 | PrimeNG |
|---|---|---|---|---|
| Primary color | `--color-primary` | `--primary` | `--mat-sys-primary` | `--p-primary-color` |
| Primary foreground | `--color-primary-content` | `--primary-foreground` | `--mat-sys-on-primary` | `--p-primary-contrast-color` |
| Destructive / Error | `--color-error` | `--destructive` | `--mat-sys-error` | `--p-red-500` |
| Background | `--color-base-100` | `--background` | `--mat-sys-surface` | `--p-surface-0` |
| Border radius | `--radius-box` | `--radius` | `--mat-sys-corner-medium` | `--p-border-radius` |

These systems all solve the same problem — semantic, themeable design tokens consumed via Tailwind utility classes or CSS custom properties — but they are mutually incompatible. A team using daisyUI today that wants to adopt spartan.ng components for a few pages must maintain two parallel theming systems, or manually rewrite one to match the other.

This isn't a niche concern. The Angular ecosystem now has at least five actively maintained component libraries with different token conventions, and CSS-first libraries like daisyUI bring non-Angular options into play for any Tailwind-based project. The fragmentation is structural and growing.

## Proposal

Add first-class [Style Dictionary](https://styledictionary.com) integration to AnalogJS, building on the existing Tailwind infrastructure, so that any 3rd-party design system can be adopted, combined, or swapped through a single token pipeline.

[...]

Issue #2229, opened 2026-04-04
[AnalogJS v3] RFC: Refactor Stylesheet Pipeline: beta vs alpha, Tailwind v4, externalRuntimeStyles, and HMR


benpsnyder:
## Summary

This is a comparative RFC based on verified source research across:

- `analog` `beta` at e8e80802fbbca1b6cec3c243044aec7f6bceee12
- local `alpha`-based Analog work on branch `fix/tailwind`
- Angular source snapshot `angular-main` (`packages/compiler-cli/.../handler.ts`)
- Vite `main`

The conclusion is narrower and stronger now:

1. `beta` still treats component stylesheets as an HMR-adjacent implementation detail.
2. Tailwind support on `beta` is fragmented across generators and consumer config.
3. The local `alpha` refactor is already moving toward the right long-term shape: one Analog-owned stylesheet pipeline/registry, explicit Tailwind preprocessing, and HMR separated from stylesheet externalization.

I want maintainer/community feedback on whether this is the right direction for v3 / `alpha` before I keep splitting it into upstream PRs.

---

## TL;DR

`beta` behavior is still fundamentally:

- `liveReload` is the main public switch
- `externalRuntimeStyles` is coupled to HMR/watch behavior
- stylesheet state is stored in two loose Maps
- Tailwind is mostly expected to work via consumer-added Vite/PostCSS config

The refactor direction is:

- `hmr` becomes the primary concept, `liveReload` stays as compatibility alias
- stylesheet externalization is computed independently from HMR
- Tailwind-aware preprocessing is framework-owned
- external `styleUrls` are registered and served through an Analog stylesheet registry
- production still uses the PostCSS path where that is the real integration boundary

---

### Analog `beta` (`e8e8080`)

- `packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts:139,177-178,1038-1044,1102-1115`
- `packages/vite-plugin-angular/src/lib/host.ts:43-78,80-100`
- `packages/platform/src/lib/options.ts:64-68`
- `packages/platform/src/lib/platform-plugin.ts:58-68`
- `packages/create-analog/index.js:205-214,373-376`
- `packages/create-analog/template-latest/vite.config.ts:14-16`
- `packages/create-analog/files/styles.css:1`
- `packages/nx-plugin/src/generators/app/lib/add-tailwind-helpers.ts:53-63,103-117`
[...]

PR #2245, opened 2026-04-05
feat(platform): add generic style-pipeline hooks for community plugins

benpsnyder:
## Summary

Related to [First-class Style Dictionary support analogjs/analog#2227](https://github.com/analogjs/analog/issues/2227)  
Related to [Settle the stylesheet pipeline direction analogjs/analog#2229](https://github.com/analogjs/analog/issues/2229)  
Community follow-up: [Community-maintained style pipeline for AnalogJS snyder-tech/stacks#1](https://github.com/snyder-tech/stacks/issues/1)

This PR is no longer proposing a first-class token engine in Analog core.

Instead, it narrows the framework work down to the generic API surface that community plugins actually need if they want deep Angular integration for token pipelines, generated bridge CSS, and theme-aware stylesheet workflows.

## What This PR Adds

| Capability | Where it lives | Available to community plugins? | Why Analog should own it |
| --- | --- | --- | --- |
| Generic `experimental.stylePipeline` config surface | `@analogjs/platform` | Yes | Gives community plugins an Analog-first, strongly typed entry point instead of each app hand-wiring raw Vite internals |
| Workspace-root-aware Vite plugin composition | `@analogjs/platform` | Yes | Analog already owns app/workspace orchestration and can pass the resolved workspace root reliably to plugin factories |
| Angular stylesheet-resource hooks | `@analogjs/vite-plugin-angular` | Yes | A standalone Vite plugin cannot participate in Angular's compiler resource pipeline on its own |
| Component stylesheet preprocessing before Angular externalizes or inlines CSS | `@analogjs/vite-plugin-angular` | Yes | This is the seam needed for token-aware transforms, bridge injection, and component-style authoring ergonomics |
| Structured stylesheet transform results (`code`, `dependencies`, `diagnostics`, `tags`) | `@analogjs/vite-plugin-angular` | Yes | Lets community plugins describe what a stylesheet now depends on, not just rewrite text |
| Angular stylesheet registry access | `@analogjs/vite-plugin-angular` | Yes | Community plugins can observe the live externalized stylesheet map without reimplementing Angular-specific HMR bookkeeping |
| HMR-safe external stylesheet identity and refresh | `@analogjs/vite-plugin-angular` | Indirectly | Analog already tracks Angular externalized stylesheet ids and request aliases; plugins benefit from that framework-owned correctness |
| Strongly typed config helpers (`defineStylePipeline`, `defineAngularStylePipeline`, etc.) | `@analogjs/platform`, `@analogjs/vite-plugin-angular` | Yes | Keeps plugin and app authoring ergonomic without bloating the public API surface |
| Reserved debug scopes for platform and Angular style-pipeline work | `@analogjs/platform`, `@analogjs/vite-plugin-angular` | Yes | Gives plugin authors a shared place to instrument tricky watch/HMR/resource behavior |

## Why This Belongs In Analog

A community Vite plugin can already do a lot on its own:

- generate CSS
- generate manifests
- watch token/config files
- inject global styles
- expose virtual modules

The part it **cannot** own cleanly is the Angular compiler/resource boundary that Analog already controls.

That boundary includes:

| Angular-specific need | Why a plain Vite plugin is not enough |
| --- | --- |
| `CompilerHost` resource interception | Angular style resources are resolved and transformed before they become normal Vite modules |
| Component stylesheet preprocessing | Token-aware or bridge-aware transforms need to run where Angular hands styles to Vite |
| Externalized stylesheet identity | Angular component styles are hashed/externalized; Analog already maps source files to browser-visible stylesheet ids |
| Resource-aware HMR invalidation | HMR needs to understand Angular stylesheet wrapper modules and direct CSS requests, not just file paths |
| Incremental resource tracking | Angular recompilation needs modified resource file knowledge, not only generic Vite watcher events |

That is why the framework-owned part should stay small but real.

## What Stays Out Of Core

This PR intentionally does **not** put any of the following into Analog core:

| Concern | Recommended home |
| --- | --- |
| Style Dictionary dependency tree | Community packages |
| Panda CSS config/codegen semantics | Community packages |
| Tokiforge runtime theming behavior | Community packages |
| PrimeNG / MUI / Shoelace / daisyUI / Spartan target contracts | Community packages |
| Design-token schemas and bridge-generation policy | Community packages |

The implementation path for those lives in [Community-maintained style pipeline for AnalogJS snyder-tech/stacks#1](https://github.com/snyder-tech/stacks/issues/1).

## Developer Experience This Unlocks

For community plugin authors, Analog now provides a clean place to plug in when they need more than global CSS generation:

| Plugin need | Enabled by this PR |
| --- | --- |
| Style Dictionary or custom token output plugins | Yes, through `experimental.stylePipeline.plugins` |
| Angular-aware bridge injection into component CSS | Yes, through `angularPlugins.preprocessStylesheet()` |
| Recording token/bridge dependencies per stylesheet | Yes, through structured transform results |
| Emitting selector/theme diagnostics during preprocessing | Yes, through structured transform results |
| HMR-safe refresh of generated/externalized styles | Yes, through the existing stylesheet registry and refresh path |
| Shared debug output for style-pipeline issues | Yes, through reserved platform/angular debug scopes |

## Implementation Notes

This PR trims the earlier token prototype down to the minimal generic pieces:

- removes the first-class Style Dictionary prototype from `@analogjs/platform`
- adds a narrow `experimental.stylePipeline` contract in `@analogjs/platform`
- forwards Angular-specific style-pipeline plugins into `@analogjs/vite-plugin-angular`
- exposes typed authoring helpers for both platform-side and Angular-side configuration
- lets Angular-side plugins return structured transform metadata that Analog tracks in the stylesheet registry and HMR path
- documents the community-plugin usage model instead of proposing library-specific framework behavior in core

## Test Plan

| Check | Status |
| --- | --- |
| `pnpm nx test platform --runTestsByPath packages/platform/src/lib/style-pipeline.spec.ts packages/platform/src/lib/platform-plugin.spec.ts` | Passed |
| `pnpm nx test vite-plugin-angular --runTestsByPath packages/vite-plugin-angular/src/lib/style-pipeline.spec.ts packages/vite-plugin-angular/src/lib/host.spec.ts packages/vite-plugin-angular/src/lib/stylesheet-registry.spec.ts packages/vite-plugin-angular/src/lib/angular-vite-plugin.spec.ts packages/vite-plugin-angular/src/lib/angular-vite-plugin-live-reload.spec.ts` | Passed |
| `pnpm nx build vite-plugin-angular` | Passed |
| `pnpm nx build platform` | Passed |

## Merge Recommendation

This is now a much smaller and more maintainable ask than the earlier token prototype:

- Analog owns the generic framework seam
- community packages own token engines, bridge generation, and library-specific behavior

If maintainers are comfortable with the framework boundary, this is now shaped as a mergeable core hook surface rather than a first-class token feature.

### DIFF
--- a/packages/vite-plugin-angular/src/lib/style-pipeline.spec.ts
+++ b/packages/vite-plugin-angular/src/lib/style-pipeline.spec.ts
@@ -1,31 +1,24 @@
 import { describe, expect, it, vi } from 'vitest';
-import {
-  defineAngularStylePipeline,
-  defineAngularStylePipelinePlugins,
-} from '../../../platform/src/lib/style-pipeline.js';
 import { AnalogStylesheetRegistry } from './stylesheet-registry.js';
 import * as debug from './utils/debug.js';
 import {
+  type AngularStylePipelineOptions,
+  type AngularStylePipelinePlugin,
   configureStylePipelineRegistry,
   stylePipelinePreprocessorFromPlugins,
 } from './style-pipeline.js';
 
 describe('angular style-pipeline hooks', () => {
   it('keeps angular style-pipeline options strongly typed during config authoring', () => {
-    expect(
-      defineAngularStylePipeline({
-        plugins: defineAngularStylePipelinePlugins([
-          {
-            name: 'plugin-a',
-          },
-        ]),
-      }),
-    ).toEqual({
-      plugins: [
-        {
-          name: 'plugin-a',
-        },
-      ],
+    const plugins: AngularStylePipelinePlugin[] = [
+      {
+        name: 'plugin-a',
+      },
+    ];
+    const options: AngularStylePipelineOptions = { plugins };
+
+    expect(options).toEqual({
+      plugins,
     });
   });
 


brandonroberts (changes requested):
## Review — Request Changes

### Scope

1. **`@standard-schema/spec`** is added as a runtime dependency in `packages/platform/package.json` but has zero references in the platform source. Remove it.

2. **`style-dictionary: ^5.4.0`** added to `pnpm-workspace.yaml` catalog but no package references it. Remove the catalog entry.

3. **`StylePipelineManifest` and `StylePipelineOutput`** are exported from `packages/platform/src/index.ts` but have no consumer, no test, and no documentation in this PR. Remove them and introduce when a real consumer exists.

### Package placement

Non-compilation types and helpers should live in `@analogjs/platform`, not `vite-plugin-angular`. The compiler boundary should stay minimal.

Specifically, `StylesheetDependency`, `StylesheetDiagnostic`, `StylesheetTransformContext`, `StylesheetTransformResult`, `normalizeStylesheetTransformResult`, `composeStylePreprocessors`, `defineAngularStylePipeline`, and `defineAngularStylePipelinePlugins` are generic stylesheet transform primitives and config helpers — not Angular compiler integration. Move these to `@analogjs/platform` and import them in `vite-plugin-angular` where needed.

Only the actual `preprocessStylesheet` integration into the compiler host and `configureStylePipelineRegistry` need to live at the compiler boundary.

### Code findings

4. **`StylePreprocessor` return type broadened** — changed from `=> string` to `=> string | StylesheetTransformResult` in `style-preprocessor.ts`. This is a breaking change. Please document it in the PR body.

5. **Style-pipeline preprocessor silently no-ops during HMR** — `stylePipelinePreprocessorFromPlugins` in `style-pipeline.ts` returns `code` unchanged when `context` is `undefined`. The live-reload path passes `undefined` as context, so community plugin preprocessors are silently skipped during HMR. Add a debug log when bailing due to missing context.

6. **`recomputeSourceMetadata` is O(N²) at startup** — iterates all `servedById` entries on every registration in `stylesheet-registry.ts`. Consider indexing records by `sourcePath` or deferring computation to query time.

7. **Dependency normalization duplicated** — `normalizeStylesheetDependencies` in `angular-vite-plugin.ts` and the inline `.map(dep => typeof dep === 'string' ? { id: dep } : dep)` in `host.ts` do the same thing. Share the helper.

### Guidelines

- PR uses "Related to" but never `Closes #` — if this resolves #2227 or #2229, use `Closes`.
- Title scope is `platform` but ~60% of code changes are in `vite-plugin-angular`. The squash-merge commit message will only reference `platform`.

benpsnyder:
@brandonroberts I wanted to sanity check the package naming before I go any further with this extraction.

Your review comment made the ownership direction clear to me:

- the non-compilation types/helpers should not live in `@analogjs/vite-plugin-angular`
- the compiler boundary should stay minimal
- only the actual Angular compiler integration should remain there

I agree with that, and that is what pushed me to extract the shared primitives out of the compiler package.

Where I’m less sure is whether the extracted package should stay narrowly named for the seam it actually owns today, or whether you want me to generalize it now in anticipation of more Analog plugin seams later.

Right now the shared package only owns style-related contracts:

- Vite plugin collection helpers for this seam
- Angular stylesheet pipeline plugin types
- stylesheet transform result types and helper functions

That makes `@analogjs/style-pipeline` feel accurate to me for the current code.

My reasoning is:

- your comment clearly argues for extraction out of the compiler package
- it clearly argues for minimizing the compiler boundary
- it does not, by itself, seem to argue that the extracted package should already be broader than the style seam

So my default recommendation would be:

- keep the package narrow for this PR
- keep the current style-specific exports
- only broaden the package name when there is a second real plugin seam that shares the same ownership model

If we stay narrow, I think the shape should be:

- `@analogjs/style-pipeline`
- `@analogjs/style-pipeline/style-preprocessor`

That seems like the cleanest reflection of the current boundary:

- compiler package consumes the shared contracts, but does not own them
- shared package owns the generic style seam
- `@analogjs/platform` can still re-export the surface for Analog-first authoring

If you want to reserve a broader home now for multiple future extension seams, I think that should be an intentional API change, not just a rename. In that case I’d probably rename it to something like:

- `@analogjs/plugin-kit`

And I would move the current exports under style-specific subpaths, for example:

- `@analogjs/plugin-kit/style-pipeline`
- `@analogjs/plugin-kit/style-preprocessor`

I would avoid broader names like `@analogjs/plugins` or `@analogjs/platform-plugins`, since those feel either too vague or like they pull ownership back toward `platform`.

benpsnyder:
@brandonroberts

I reworked this PR to align with the review direction and to get it back to a mergeable shape.

What is addressed here:

- removed the unused `@standard-schema/spec` runtime dependency from `packages/platform/package.json`
- removed the unused `style-dictionary` catalog entry
- removed the unused `StylePipelineManifest` / `StylePipelineOutput` exports
- removed the new `packages/style-pipeline` package entirely
- moved the public style-pipeline / stylesheet transform authoring surface back under `@analogjs/platform`
- kept `@analogjs/vite-plugin-angular` focused on the Angular compiler-facing stylesheet integration and registry wiring
- added the debug log when stylesheet preprocessors are skipped because Angular did not provide a stylesheet context during HMR
- shared stylesheet dependency normalization through one helper
- tightened `recomputeSourceMetadata()` so it recomputes from the source-indexed public ids instead of rescanning all served records

What is still intentionally not solved in this PR:

- the `StylePreprocessor` return-type widening still needs to be called out in the PR body as a breaking change
- there is still some private duplication of stylesheet transform types/helpers inside `@analogjs/vite-plugin-angular`

I did not push the remaining duplication all the way through in this PR because doing that cleanly wants the broader interop refactor we discussed, rather than another round of package-boundary churn here. I’d prefer to land this PR with the public/package boundary corrected now, then follow up with the thinner Nitro-style interop cleanup in [Research Nitro-style Vite plugin interop hooks for Analog compilation analogjs/analog#2254](https://github.com/analogjs/analog/issues/2254).

That follow-up is where I think we should finish the remaining work:

- make the Analog/Vite interop seam more Nitro-like
- keep `@analogjs/vite-plugin-angular` as thin as possible
- remove the remaining compiler-local duplication once the setup/integration shape is clearer

Verification on this branch:

- `pnpm exec vitest run packages/vite-plugin-angular/src/lib/style-pipeline.spec.ts packages/vite-plugin-angular/src/lib/stylesheet-registry.spec.ts packages/platform/src/lib/style-pipeline.spec.ts packages/platform/src/style-pipeline.spec.ts packages/platform/src/style-preprocessor.spec.ts packages/platform/src/lib/platform-plugin.spec.ts`
- `pnpm nx build vite-plugin-angular`
- `pnpm nx build platform`

PR #2258, opened 2026-04-07
build: exclude workspace deps from optimizeDeps in dogfooding apps

benpsnyder:
# Keep Workspace Framework Packages on the Source Transform Path

## Summary

This PR excludes local `workspace:*` Analog packages from Vite dependency prebundling in the main Analog dogfooding apps.

That keeps first-party framework packages on the normal source-transform path during development instead of letting Vite treat them like ordinary third-party vendor deps.

The goal is not startup speed. The goal is framework-development correctness.

## Why this still matters after `analogjs/analog#2279`

After [retire root tsconfig paths and move workspace linking to package-native resolution analogjs/analog#2279](https://github.com/analogjs/analog/pull/2279), these apps rely more directly on workspace package resolution.

That is a good cleanup, but it also makes it more important to control how Vite sees those local framework packages during dev:

- if a local `@analogjs/*` package is prebundled too early, Vite can stop treating it like source that still needs framework-aware transforms
- that is exactly the wrong tradeoff for Analog dogfooding apps, where the point is to exercise local framework code as faithfully as possible
- these apps should bias toward correctness of local framework behavior, not maximum dep-optimizer aggressiveness

So this PR is still relevant after `analogjs/analog#2279`. If anything, the package-native resolution cleanup makes the rationale clearer.

## What changed

I added a small shared helper that reads an app `package.json`, finds dependencies declared with the `workspace:` protocol, and returns their package names for `optimizeDeps.exclude`.

Affected apps:

- `apps/analog-app`
- `apps/blog-app`
- `apps/opt-catchall-app`
- `apps/tailwind-debug-app`
- `apps/tanstack-query-app`

Each of those apps now excludes its local workspace deps from Vite prebundling while preserving any existing `optimizeDeps.include` entries.

## Why this is valuable

These are the apps we use to dogfood Analog itself.

For that workflow, local framework packages should stay visible to the same source-aware pipeline that handles app code, including cases such as:

- Angular external templates and styles
- Analog and Angular source transforms
- framework/plugin behavior across app and local library boundaries
- local package edits that should be reflected accurately during development

This keeps the dogfooding environment closer to the real framework-authoring use case.

## Scope and risk

This change is intentionally narrow:

- only dogfooding app Vite configs are touched
- package/library build configs are unchanged
- no public API changes
- no runtime feature changes for published users
- the behavior change is limited to dev-time dependency optimization policy in these apps

The tradeoff is also intentional:

- slightly less aggressive dep prebundling during dev
- in exchange for better correctness when working on local Analog packages

That is the right trade for framework-maintainer workflows.

## Validation

I re-checked this against current `alpha` and verified the affected apps still build:

```bash
pnpm nx run-many -t build -p analog-app,blog-app,opt-catchall-app,tailwind-debug-app,tanstack-query-app --skip-nx-cache
```

That completed successfully on the PR branch during review.

## Reviewer lens

The key question is not whether this improves startup speed.

The key question is whether Analog's dogfooding apps should allow local first-party framework packages onto Vite's prebundle path by default, or keep them on the source-transform path where Analog and Angular can continue treating them like live framework source.

This PR intentionally chooses the second behavior.


benpsnyder:
@brandonroberts FYI I need to revisit this and research after https://github.com/analogjs/analog/pull/2279 closes to see how relevant it is and post research findings

benpsnyder:
@brandonroberts this is verified and good to merge once CI clears

PR #2274, opened 2026-04-10
fix(angular-compiler): skip arrow fn types when finding assignment = …

ashley-hunter:
`findEqualsSign()` in class field lowering scanned from `key.end` to `value.start` for the assignment `=`, but for private    
  fields with arrow function type annotations (e.g. `#cb: (() => void) | null = null`) it matched the `=` inside `=>` first,    
  removing part of the type annotation and producing invalid output like `#cb: (() ;`.   

Fix: when a type annotation is present, start scanning from typeAnnotation.end instead of key.end.

## PR Checklist

Closes #

## Affected scope

- Primary scope: angular-compiler
- Secondary scopes:

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

## What is the new behavior?

When a private class field has a type annotation, `findEqualsSign()` now starts scanning from `typeAnnotation.end` instead of `key.end`, skipping past any `=` characters inside arrow function type annotations (`=>`). This ensures only the actual assignment `=` is matched.                                                                                                    
                                                                  
  ```typescript
  // Previously mangled output:
  #registeredClearUserData: (() ;
  #onChange: (value: string) ;                                                                                                  
   
  // Now correctly preserved:                                                                                                   
  #registeredClearUserData: (() => void) | null;                  
  #onChange: (value: string) => void;    
```

## Test plan

- [X] `nx format:check`
- [X] `pnpm build`
- [X] `pnpm test`
- [ ] Manual verification

- [ ] Yes
- [X] No

### DIFF
--- a/packages/angular-compiler/src/lib/class-field-lowering.spec.ts
+++ b/packages/angular-compiler/src/lib/class-field-lowering.spec.ts
@@ -285,4 +285,62 @@ export default class TestComponent {
     expect(result).toContain(`this[KEY] = 'value'`);
     expect(result).not.toContain(`this.[KEY]`);
   });
+
+  it('does not mangle private field with arrow function type and initializer', () => {
+    const result = compileWithLowering(`
+      import { Component } from '@angular/core';
+      @Component({ selector: 'app-test', template: '' })
+      export class TestComponent {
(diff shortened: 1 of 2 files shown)


brandonroberts (approved):
LGTM

PR #2276, opened 2026-04-10
fix(angular-compiler): constant pool helpers survive unused import elision

ashley-hunter:
Include pending sideEffects (containing `setClassMetadata` IIFEs) in the string passed to `detectTypeOnlyImportNames` when calculating the helper insertion position. This ensures decorator names like `Component` - which are referenced in value position inside `setClassMetadata` - are not falsely flagged as type-only, so the import is not skipped and helpers are placed at a safe position that survives the downstream import overwrite.

Example

Given this input:

```ts
  import { Component, signal, computed } from '@angular/core';

  @Component({
    selector: 'app-counter',
    template: `<ng-content></ng-content>`,
  })                                                                                      
  export class Counter {}
```
                                                                                          
Before (broken) `_c0` is referenced but never defined:                                  
  
```ts
  import * as i0 from "@angular/core";                                                    
  import { Component } from '@angular/core';                                              
   
  export class Counter {                                                                  
    static ɵfac = /*@__PURE__*/ (__ngFactoryType__) => {return new (__ngFactoryType__ || Counter)();};                                                                           
    static ɵcmp = /*@__PURE__*/ i0.ɵɵdefineComponent({type: Counter, selectors: [["app-counter"]], ngContentSelectors: _c0, /* ... */});                                
  }                                                               
```                                                                                  
  After (fixed) - `_c0` is correctly emitted before the class:                              
   
```ts
  import * as i0 from "@angular/core";                                                    
  import { Component } from '@angular/core';                      

  const _c0 = ["*"];

  export class Counter {
    static ɵfac = /*@__PURE__*/ (__ngFactoryType__) => {return new (__ngFactoryType__ || Counter)();};                                                                           
    static ɵcmp = /*@__PURE__*/ i0.ɵɵdefineComponent({type: Counter, selectors: [["app-counter"]], ngContentSelectors: _c0, /* ... */});                                
  }   
```

## PR Checklist

Closes #

## Affected scope

- Primary scope: angular-compiler
- Secondary scopes:

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

## Test plan

- [X] `nx format:check`
- [X] `pnpm build`
- [X] `pnpm test`
- [ ] Manual verification

- [ ] Yes
- [X] No

### DIFF
--- a/packages/angular-compiler/src/lib/compile.ts
+++ b/packages/angular-compiler/src/lib/compile.ts
@@ -1181,7 +1181,17 @@ export function compile(
     // be fully removed by elideTypeOnlyImportsMagicString (step 4).
     // Inserting at a position inside a soon-to-be-removed range would cause
     // MagicString to discard the helpers along with the import.
-    const willElide = detectTypeOnlyImportNames(ms.toString());
+    // Include pending sideEffects (setClassMetadata IIFEs) in the detection
+    // string — they reference decorator names like `Component` in value
+    // position but haven't been appended to `ms` yet. Without this,
+    // decorator names are falsely flagged as type-only, causing the import
+    // to be skipped, insertPos to stay at 0, and helpers to be placed at
+    // position 0 where they get wiped by the subsequent import overwrite.
+    const codeForDetection =
+      sideEffects.length > 0
+        ? ms.toString() + '\n' + sideEffects.join('\n')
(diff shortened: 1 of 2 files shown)


brandonroberts:
Fixed in 8a15184d423b0818c3c6d9f7429095554eecbd05

PR #2283, opened 2026-04-13
fix(vite-plugin-angular): use virtual modules for external JIT styles

### LINKED ISSUE #2280: Browser tests fail for external SCSS styleUrl with vite-plugin-angular 2.4.5

```text
@angular/core                          21.2.8
@angular/common                        21.2.8
@angular/compiler                      21.2.8
@angular/compiler-cli                  21.2.8
@angular/platform-browser              21.2.8
@angular/platform-browser-dynamic      21.2.8
@angular/build                         21.2.7
@analogjs/vite-plugin-angular          2.4.5
@analogjs/vitest-angular               2.4.5
vite                                   7.3.2
vitest                                 4.1.4
@vitest/browser                        4.1.4
@vitest/browser-playwright             4.1.4
node                                   22
```

Reproduced on:
- Windows 11 (host)
- Linux (Docker, mcr.microsoft.com/playwright:v1.59.1-noble)

vite-plugin-angular

### Description

After upgrading `@analogjs/vite-plugin-angular` and `@analogjs/vitest-angular` from `2.3.1` to `2.4.5`, browser tests for Angular components that declare an **external** `styleUrl` to a `.scss` file fail during module load.

Unit tests for the same components under `happy-dom` (Vitest node path) pass with the same plugin configuration. Only the Vitest browser module-runner path is affected. Also tests do not fail if the component uses inline `styles: [...]` instead of an external `styleUrl` pointing to `.scss`. So the regression appears to be specific to the external style resource path rather than browser mode in general.

### Minimal reproduction

[Github Repo](https://github.com/kSherp/analog-repro)

A standalone project outside of any monorepo reproduces the bug. Validated results:

- fails on Windows 11 with `@analogjs/vite-plugin-angular@2.4.5`
- fails on Linux (Docker, `mcr.microsoft.com/playwright:v1.59.1-noble`) with
  `@analogjs/vite-plugin-angular@2.4.5`
- passes on Windows 11 after downgrading only
  `@analogjs/vite-plugin-angular` / `@analogjs/vitest-angular` to `2.3.1`
- passes on Linux after the same downgrade

```true
21:34:25 [vite] (client) Pre-transform error: [sass] expected "{".
  ╷
1 │ export default ":host {\n  color: red;\n}"
  │                                           ^
  ╵
  src\foo.component.scss 1:43  root stylesheet
  Plugin: vite:css
  File: .../analog-repro/src/foo.component.scss?analog-inline:1:43
  [sass] expected "{".
    ╷
  1 │ export default ":host {\n  color: red;\n}"
    │                                           ^
    ╵
    src\foo.component.scss 1:43  root stylesheet
 ❯  chromium  src/foo.component.browser.spec.ts (0 test)

⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯ Failed Suites 1 ⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯

 FAIL   chromium  src/foo.component.browser.spec.ts [ src/foo.component.browser.spec.ts ]
Error: Failed to import test file .../analog-repro/src/foo.component.browser.spec.ts
Caused by: TypeError: Failed to fetch dynamically imported module: .../analog-repro/src/foo.component.browser.spec.ts?import&browserv=1775932464917
⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯⎯[1/1]⎯
```

### Other information

The `load` hook for `?analog-inline` reads the source, runs it through `preprocessCSS`, and returns a JavaScript module (`export default "<compiled css>"`). The resolved id at that point still ends in `.scss?analog-inline`, so Vite's `vite:css` matches it by extension via `cssLangRE` and runs its own `transform` on the module, handing the already-compiled JavaScript payload to Sass.

`2.3.1` rewrote external `styleUrl`s to `?inline` (not `?analog-inline`) and let Vite's native inline CSS handling own the module end-to-end, so no double processing occurred. The switch to `?analog-inline` appears to have been introduced in the 2.4.x line to work around Vite's `?raw` / `?inline` security checks. However, the rewritten id still keeps the `.scss` extension, which allows `vite:css` to match it again.

One direction that closed the symptom for us locally: route the compiled inline style through a virtual module id without a stylesheet extension. With a `\0virtual:`-style id, `cssLangRE` no longer matches and `vite:css` does not re-enter, while the `load` hook still returns the same `export default "<compiled css>"` payload the JIT runtime expects. Applying this shape to our local `node_modules/@analogjs/vite-plugin-angular` made the monorepo's browser test suite pass on Windows. There may be a simpler or more idiomatic fix on your side — sharing this only as one working approach.

- [x] Yes
- [ ] No

kSherp:
## PR Checklist

In JIT mode (used implicitly by Vitest browser module-runner), external `styleUrl` / `styleUrls` imports were rewritten to ids like `/path/to/component.scss?analog-inline`. That id still resembles a stylesheet resource to Vite, so `vite:css` re-entered after this plugin's `load()` hook had already called `preprocessCSS()` and returned a JavaScript module (`export default "..."`). For external `.scss` / `.sass` / `.less` styles, this caused the CSS pipeline to try compiling JavaScript as stylesheet source — see the error trace and minimal reproduction in the issue.

Closes #2280

## Affected scope

- Secondary scopes: none

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

## What is the new behavior?

This change routes external JIT style-file imports through virtual module ids instead: `virtual:@analogjs/vite-plugin-angular:inline-style:<base64url-encoded-absolute-path>`. That keeps the final id opaque to Vite's CSS pipeline while preserving the existing JIT contract:

- `transform` rewrites `angular:jit:style:file;...` imports to the new virtual id
- `resolveId` maps that virtual prefix to a `\0`-prefixed Rollup virtual module id
- `load` decodes the path, reads the stylesheet, runs `preprocessCSS()`, and returns `export default "..."`
- `handleHotUpdate` now manually finds and invalidates the corresponding virtual style module so HMR / live-reload propagation still works when the backing stylesheet changes

Only the external JIT style-file path is changed. The existing inline JIT style path (`angular:jit:style:inline`) is unchanged and still goes through the existing `virtual:angular:jit:style:inline;...` flow.

## Test plan

- [x] `nx format:check`
- [x] `pnpm build`
- [~] `pnpm test` — three unrelated Windows-only test failures remain due to platform-sensitive path assertions, and all are path-normalization issues rather than regressions from this change. I verified locally that these assertions pass once the expected paths are normalized for Windows as well. If maintainers want, I can add those path-normalization adjustments in a follow-up or fold them into this PR.
- [x] Manual verification — browser repro passed with the locally built `@analogjs/vite-plugin-angular`

Targeted coverage added/updated in `angular-vite-plugin.spec.ts`:

- `resolveId` asserts the new virtual-id contract for external JIT styles
- virtual ids resolve to `\0`-prefixed Rollup virtual modules
- virtual ids do not match the stylesheet-resource regex that would send them back through Vite's CSS pipeline
- `load` roundtrip verifies virtual-id resolution back to the source file, then `readFile` -> `preprocessCSS()` -> `export default`
- existing `?inline` and rewritten `?analog-inline` load-hook compatibility paths are both covered

A dedicated `transform` unit test was not added because that path is tightly coupled to Angular compilation state and would require heavier integration-style mocking than the rest of this spec file.

- [ ] Yes
- [x] No

## Other information

`experimental.useAnalogCompiler` is intentionally untouched for scope control. The default `angular-vite-plugin` path is fixed here, but there may be a similar issue in the experimental compiler path. I can investigate and follow up with a parity fix separately if wanted.

### DIFF
--- a/packages/vite-plugin-angular/src/lib/angular-vite-plugin.spec.ts
+++ b/packages/vite-plugin-angular/src/lib/angular-vite-plugin.spec.ts
@@ -1,4 +1,6 @@
+import * as path from 'node:path';
 import { describe, it, expect, vi } from 'vitest';
+import { normalizePath, preprocessCSS } from 'vite';
 
 vi.mock('vite', async () => {
   const actual = await vi.importActual<typeof import('vite')>('vite');
@@ -67,7 +69,7 @@ describe('isTestWatchMode', () => {
 });
 
 describe('JIT resolveId', () => {
-  it('should resolve style files with ?analog-inline suffix', () => {
+  it('should resolve style files to virtual style ids', () => {
     const plugins = angular({ jit: true });
     const mainPlugin = plugins.find(
       (p) => p.name === '@analogjs/vite-plugin-angular',
@@ -82,8 +84,10 @@ describe('JIT resolveId', () => {
       '/project/src/app/my-component.ts',
     );
 
-    expect(result).toBeDefined();
-    expect(result).toContain('?analog-inline');
+    expect(result).toContain(
+      'virtual:@analogjs/vite-plugin-angular:inline-style:',
+    );
+    expect(result).not.toContain('?analog-inline');
     expect(result).not.toContain('?inline');
   });
 
@@ -105,6 +109,20 @@ describe('JIT resolveId', () => {
     expect(result).toBeDefined();
     expect(result).toContain('?analog-raw');
     expect(result).not.toContain('??analog-raw');
+  });
+
+  it('should resolve bare virtual style ids to rollup virtual modules', () => {
+    const plugins = angular({ jit: true });
+    const mainPlugin = plugins.find(
+      (p) => p.name === '@analogjs/vite-plugin-angular',
+    );
+    expect(mainPlugin).toBeDefined();
+
+    const resolveId = (mainPlugin as any).resolveId;
+    const virtualId =
+      'virtual:@analogjs/vite-plugin-angular:inline-style:test-style-id';
+
+    expect(resolveId(virtualId)).toBe(`\0${virtualId}`);
   });
 
   it('should intercept .html?raw imports and remap to ?analog-raw', () => {
@@ -152,20 +170,24 @@ describe('JIT resolveId', () => {
     );
 
     const resolveId = (mainPlugin as any).resolveId;
+    const importer = '/project/src/app/my-component.ts';
+    const expectedScssPath = `${normalizePath(
+      path.resolve(path.dirname(importer), './my-component.scss'),
+    )}?analog-inline`;
+    const expectedCssPath = `${normalizePath(
+      '/project/src/app/my-component.css',
+    )}?analog-inline`;
 
     // Relative .scss?inline
-    const result = resolveId(
-      './my-component.scss?inline',
-      '/project/src/app/my-component.ts',
-    );
-    expect(result).toBe('/project/src/app/my-component.scss?analog-inline');
+    const result = resolveId('./my-component.scss?inline', importer);
+    expect(result).toBe(expectedScssPath);
 
     // Absolute .css?inline
     const result2 = resolveId(
       '/project/src/app/my-component.css?inline',
       '/project/src/app/other.ts',
     );
-    expect(result2).toBe('/project/src/app/my-component.css?analog-inline');
+    expect(result2).toBe(expectedCssPath);
   });
 
   it('should intercept style ?inline imports even without jit mode', () => {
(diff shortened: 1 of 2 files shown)


brandonroberts (approved):
LGTM

brandonroberts:
Thanks @kSherp!

PR #2311, opened 2026-04-21
fix(vite-plugin-angular): let CSS ?inline imports flow through Vite's pipeline

### LINKED ISSUE #2310: (vite-plugin-angular): test.css required in browser mode

Vitest Browser Mode

vite-plugin-angular

### Description

Inclusion of CSS is vital in vitest browser mode for proper testing behavior. Since #2306 and version 2.4.9, however, all CSS is ignored by default unless the `test.css` vite option is set to `true`. At the same time the [vitest documentation warns](https://vitest.dev/config/css) that `test.css` is ignored in browser tests. 

Should the analog plugin, therefore, ignore the `test.css` option in browser mode as well and include CSS by default?

```true

```

### Other information

_No response_

- [ ] Yes
- [x] No

brandonroberts:
## PR Checklist

Closes #2310

## Affected scope

- Secondary scopes:

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

## What is the new behavior?

Instead of routing style ?inline imports through virtual module IDs and manually calling preprocessCSS, add component style paths to Vite's safeModulePaths set so the Denied ID security check passes. This lets ?inline imports flow through Vite's native CSS pipeline, which handles preprocessing, test.css semantics, and browser-vs-node differences without the plugin reimplementing any of it.

Removes shouldPreprocessTestCss, virtual style ID system, and manual preprocessCSS calls for style loading. Template ?raw imports continue to use virtual IDs.

## Test plan

- [ ] `nx format:check`
- [ ] `pnpm build`
- [ ] `pnpm test`
- [ ] Manual verification

- [ ] Yes
- [x] No


brandonroberts:
Implemented a fix in `2.4.10` that restores Vite/Vitest built-in handling without the "Denied ID" errors

PR #2335, opened 2026-05-14
fix(vite-plugin-angular): skip source .d.ts files for composite TS projects

StNekroman:
## PR Checklist

I have composite project (so declaration: true) and previous impl took content of .d.ts files - that lead to having 0 unit tests inside spec files.
This PR extends list of ignorable files.

- [ ] Yes
- [x] No

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "analogjs-platform",
-  "version": "2.5.1",
+  "version": "2.5.2",
   "license": "MIT",
   "type": "module",
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "analogjs-platform",
-  "version": "2.5.2",
+  "version": "2.5.1",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "analogjs-platform",
-  "version": "2.5.1",
+  "version": "2.5.2",
```
brandonroberts: Revert this

brandonroberts (changes requested):

package.json
StNekroman: reverted

StNekroman: @brandonroberts Any morevents in this area?  
I need to know if I can wait for next version with fix or just apply patch fix from my side and forget.

brandonroberts (approved):

brandonroberts:
Thanks @StNekroman. The next stable release will out soon. The fix is in `2.5.2-beta.3`

PR #2373, opened 2026-06-10
fix(vite-plugin-angular): skip rolldown dep optimizer transform in test mode and close transformer

### LINKED ISSUE #2371: vite-plugin-angular: rolldown dep optimizer plugin runs in test mode and never closes its worker

```
@analogjs/vite-plugin-angular: 2.5.1 (still present on beta, 2.6.1-beta.2)
@analogjs/vitest-angular: 2.5.1
vite: 8.x (rolldown — vite.rolldownVersion is set)
vitest / @vitest/browser: 4.x
node: 24
OS: macOS (not OS specific)
```

vite-plugin-angular

### Description

On Vite 8 (rolldown), `createDepOptimizerConfig` adds a compiler plugin to the dep optimizer in two flavors:

- the esbuild one (`createCompilerPlugin`) gets an `isTest` flag and skips its `onLoad` transform in test mode. It also closes its `JavaScriptTransformer` in `onEnd`.
- the rolldown one (`createRolldownCompilerPlugin`) gets neither. It always transforms, and never closes.

So when running vitest on Vite 8, every prebundled `.js`/`.cjs`/`.mjs` dependency is piped through Angular's `JavaScriptTransformer` (the Babel linker from `@angular/build/private`), built with one worker and never shut down.

With a large package in `optimizeDeps.include`, the optimizer can stall before the vitest RUN banner even shows. The same package prebundles in seconds in a plain Vite config without the plugin. Leaving the package out of `optimizeDeps.include` is the only workaround.

Code:

- `packages/vite-plugin-angular/src/lib/compiler-plugin.ts` — the esbuild variant has `if (!isTest)` and `build.onEnd(() => javascriptTransformer.close())`; the rolldown variant has neither.
- `packages/vite-plugin-angular/src/lib/utils/plugin-config.ts` — `createRolldownCompilerPlugin` is called without `opts.isTest` / close flag, while `createCompilerPlugin` right below gets both.

```
No exception. The dep optimizer stalls (or runs slowly) before the vitest RUN banner
when a large JS-heavy package is in optimizeDeps.include on Vite 8 + rolldown.
```

### Other information

Suggested fix: mirror the esbuild variant — skip the `load` transform when `isTest`, and close the transformer in `buildEnd`. `JavaScriptTransformer.close()` resets its lazy worker pool, so a later optimizer run just creates it again. I have a fix with tests ready and will open a PR.

Yes

fyvfyv:
## PR Checklist

On Vite 8 (rolldown), the dep optimizer compiler plugin runs Angular's `JavaScriptTransformer` over every prebundled `.js`/`.cjs`/`.mjs` dependency even in test mode, and never closes the transformer's worker. The esbuild variant of the same plugin skips the transform in test mode (`if (!isTest)`) and closes the transformer in `onEnd`. Large packages in `optimizeDeps.include` can stall the optimizer before the vitest RUN banner because everything funnels through that single never-closed worker.

Closes #2371

## Affected scope

- Secondary scopes: none

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

n/a

## What is the new behavior?

`createRolldownCompilerPlugin` now takes the same `isTest` and `closeTransformer` flags as the esbuild variant:

- in test mode it does not register the `load` transform, so prebundled deps are not piped through the Babel linker — same as the esbuild path;
- when `closeTransformer` is set (everything except the Astro integration), the transformer is closed in `buildEnd`. `JavaScriptTransformer.close()` resets its lazy worker pool, so a later optimizer run re-creates it on demand.

`createDepOptimizerConfig` passes `opts.isTest` and `!opts.isAstroIntegration`, mirroring the esbuild call right below it.

## Test plan

- [x] `nx format:check`
- [x] `pnpm build` (`nx build-package vite-plugin-angular`, tsc, no errors)
- [x] `pnpm test` (full workspace green; `nx test vite-plugin-angular`: 625 passed, 3 skipped)
- [x] Manual verification: new unit tests in `compiler-plugin.spec.ts` fail on the old code (the `load` hook was registered in test mode; no `buildEnd`) and pass with the fix

- [ ] Yes
- [x] No

`createRolldownCompilerPlugin` is internal — the package entry point only exports `angular`. The only call site is `createDepOptimizerConfig`.

## Other information

The rest of the dep optimizer behavior on the rolldown path is unchanged.

### DIFF
--- a/packages/vite-plugin-angular/src/lib/compiler-plugin.spec.ts
+++ b/packages/vite-plugin-angular/src/lib/compiler-plugin.spec.ts
@@ -1,0 +1,37 @@
+import { describe, expect, it } from 'vitest';
+import { createRolldownCompilerPlugin } from './compiler-plugin.js';
+
+const pluginOptions = {
+  tsconfig: 'tsconfig.spec.json',
+  sourcemap: false,
+  incremental: false,
+};
+
+describe('createRolldownCompilerPlugin', () => {
+  it('skips the dependency transform in test mode', () => {
+    const plugin = createRolldownCompilerPlugin(pluginOptions, true, true);
+
+    expect(plugin.load).toBeUndefined();
+  });
+
+  it('keeps the dependency transform outside test mode', () => {
+    const plugin = createRolldownCompilerPlugin(pluginOptions, false, true);
+
+    expect(plugin.load).toBeDefined();
+  });
+
+  it('closes the JavaScript transformer when the build ends', async () => {
+    const plugin = createRolldownCompilerPlugin(pluginOptions, false, true);
+
+    expect(plugin.buildEnd).toBeDefined();
+    await expect(
+      (plugin.buildEnd as unknown as () => Promise<void>)(),
+    ).resolves.toBeUndefined();
+  });
+
+  it('keeps the transformer open when closeTransformer is false', () => {
+    const plugin = createRolldownCompilerPlugin(pluginOptions, false, false);
+
+    expect(plugin.buildEnd).toBeUndefined();
+  });
+});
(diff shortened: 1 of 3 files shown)


brandonroberts (approved):

brandonroberts:
Thanks @fyvfyv!

PR #2434, opened 2026-07-23
fix(vite-plugin-angular): ensure compatibility with NodeNext resolution

mrginglymus:
The @angular/compiler-cli package is not compatible with NodeNext module resolution, meaning downstream consumers with skipLibCheck: false using NodeNext module resolution will hit errors when using the angular-vite-plugin.

This is because the function signature of the exported getFileMetadata function includes a reference to an export from @angular/compiler-cli.

However, as far as I can tell getFileMetadata is an internal utility that is exported only for testing purposes.

By marking it as @internal and enabling stripInternal in the corresponding tsconfig.json, we can remove the reference to @angular/compiler-cli in the published types for this package and ensure compatibility with NodeNext resolution.

## Affected scope

- Secondary scopes:

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

## Test plan

- [ ] `nx format:check`
- [ ] `pnpm build`
- [ ] `pnpm test`
- [ ] Manual verification

- [ ] Yes
- [x] No

### DIFF
--- a/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
+++ b/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
@@ -1764,6 +1764,9 @@ const hmrMetadataCache = new WeakMap<
   }
 >();
 
+/**
+ * @internal
+ */
 export function getFileMetadata(
   program: ts.BuilderProgram,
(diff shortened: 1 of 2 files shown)


brandonroberts (approved):
LGTM

PR #2456, opened 2026-07-31
fix(vite-plugin-angular): replace ts-morph with oxc-parser in component resolvers

### LINKED ISSUE #2455: Update ts-morph dependency to v25+ to mitigate CVE-2026-14257
Part of the current dependency tree looks like this:
```
@analogjs/vite-plugin-angular@2.6.4
 └─┬ ts-morph@21.0.1
   └─┬ @ts-morph/common@0.22.0
     └─┬ minimatch@9.0.9
       └── brace-expansion@2.1.4
```

Now, brace-expansion <5.0.8 is vulnerable to CVE-2026-14257. 

- minimatch uses brace-expansion@5.0.8 as of v10
- @ts-morph/common uses minimatch@10 as of v26
- ts-morph uses @ts-morph/common@26 as of v25

So, an upgrade from current ts-morph@21 to at least v25 would fix this. Since all intermediate packages are used at a rather old version, we should probably not assume, they will fix their version constraint there since they would probably have to handle breaking changes in an old version.
That being said, I do not really know about breaking changes between ts-morph v21 and v25 which need to be handled here. As far as I understand, it is mostly about the lowest supported Typescript version.

brandonroberts:
## PR Checklist

`@analogjs/vite-plugin-angular` declares `ts-morph@^21.0.0`, which pulls in a vulnerable transitive dependency:

```
@analogjs/vite-plugin-angular
 └─┬ ts-morph@21.0.1
   └─┬ @ts-morph/common@0.22.0
     └─┬ minimatch@9.0.9
       └── brace-expansion@2.1.4   ← CVE-2026-14257
```

The issue proposes bumping `ts-morph` to v25+. This PR removes the dependency instead: `ts-morph` had exactly one remaining use in the package — `component-resolvers.ts`, which extracts `templateUrl` / `styleUrl` / `styleUrls` from component source so Vite can watch those files. `oxc-parser` is already a direct dependency of this same package (used by `compile.ts`, `resource-inliner.ts`, `style-ast.ts`, `type-elision.ts`, `jit-transform.ts`, `dts-reader.ts`, `registry.ts`), and handles this extraction directly.

So the CVE chain existed only to support one small extraction routine the existing parser already covers. No new dependency is added.

Closes #2455

## Affected scope

- Secondary scopes: none

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

## What is the new behavior?

`component-resolvers.ts` is reimplemented on `oxc-parser`, following the patterns already established in `style-ast.ts` and `resource-inliner.ts`. The per-id caches and the shared last-parse memo are preserved, so `StyleUrlsResolver` and `TemplateUrlsResolver` still share a single parse of the same code.

Two behavior notes worth reviewer attention:

1. **Non-literal values are now skipped.** The old implementation did `.getText().replace(/['"`]/g, '')` on the initializer, so `styleUrl: STYLE_URL` produced the bogus path `"STYLE_URL"`, and a template literal with interpolation produced a mangled string. The new implementation accepts only string literals and interpolation-free template literals. This is a fix, but it is a behavior change — covered by an added test.

2. **The generic AST walk is deliberate.** `style-ast.ts` scopes its search to `@Component` decorators on top-level classes; this file does not. ts-morph's `getDescendantsOfKind(PropertyAssignment)` scanned the entire file, so narrowing the scope could silently break style/template HMR for components declared inside functions or in non-top-level classes. The walk visits every `ObjectExpression`, matching the previous behavior.

Also fixes a latent crash found while validating: array holes (`styleUrls: ['./a.css', , './b.css']`) are `null` entries in the oxc AST and crashed the recursive walk. `resource-inliner.ts` guards against sparse arrays for the same reason, so this is a case that occurs in practice.

## Test plan

- [x] `nx format:check` — clean
- [x] `nx build vite-plugin-angular` — succeeds (package-scoped; full `pnpm build` not run)
- [x] `vitest run --root packages/vite-plugin-angular` — 673 passed / 6 skipped across 36 files (up from 671, with the 2 added tests)
- [x] Manual verification — validated against the real parser that the oxc AST has no cycles (so the recursive walk needs no visited-set), and that the walk costs ~5.6ms on a synthetic 80KB / 400-component file, well under the ts-morph parse it replaces

Two added tests cover the cases where AST parsing is now more correct than the old text-stripping: a `styleUrl` bound to an identifier / an interpolated template literal is ignored rather than yielding a bogus path, and a `templateUrl:` occurring inside an unrelated string literal is not picked up.

Note there is no `lint` target configured for this project, so `nx lint vite-plugin-angular` was not run.

- [ ] Yes
- [x] No

Consumers who passed a non-static value to `styleUrl` / `templateUrl` previously got a nonsensical resolved path, so nothing that worked before stops working.

## Other information

Out of scope for this CVE and intentionally untouched: `packages/content-plugin` uses `ts-morph@^27` (already past the fix) and the root `devDependency` is `^27.0.2`.

`pnpm-lock.yaml` is unchanged, and that is expected — `packages/` are not pnpm workspace members (only `apps/docs-app` is), so the lockfile never contained the vulnerable `ts-morph@21` tree. It comes solely from the published package's declared `dependencies`, which is what this PR edits.

---
_Generated by [Claude Code](https://claude.ai/code/session_01EH2A3qqHtfMuagi7kfsuh7)_

### DIFF
--- a/packages/vite-plugin-angular/package.json
+++ b/packages/vite-plugin-angular/package.json
@@ -56,8 +56,7 @@
     "magic-string": "^0.30.21",
     "obug": "^2.1.1",
     "oxc-parser": "^0.121.0",
-    "tinyglobby": "^0.2.14",
-    "ts-morph": "^21.0.0"
+    "tinyglobby": "^0.2.14"
   },
   "builders": "./src/lib/tools/builders.json",
   "ng-update": {
--- a/packages/vite-plugin-angular/src/lib/component-resolvers.spec.ts
+++ b/packages/vite-plugin-angular/src/lib/component-resolvers.spec.ts
@@ -250,6 +250,21 @@ describe('component-resolvers', () => {
 
       expect(resolvedPaths).toMatchNormalizedPaths(expectedPaths);
     });
+
+    it('should ignore style urls that are not static strings', () => {
+      const code = `
+        @Component({
+          styleUrl: STYLE_URL,
+          styleUrls: [\`./\${name}.css\`]
+        })
+        export class MyComponent {}
+      `;
+
+      const styleUrlsResolver = new StyleUrlsResolver();
+      const resolvedPaths = styleUrlsResolver.resolve(code, id);
+
+      expect(resolvedPaths).toHaveLength(0);
+    });
   });
 
   describe('caching', () => {
@@ -455,6 +470,24 @@ describe('component-resolvers', () => {
         const resolvedTemplateUrls = templateUrlsResolver.resolve(code, id);
 
         expect(resolvedTemplateUrls).toMatchNormalizedPaths([expectedUrl]);
+      });
+
+      it('should ignore templateUrls inside string literals', () => {
+        const code = `
+        const snippet = "templateUrl: './not-a-real-template.html'";
+
+        @Component({
+          templateUrl: './app.component.html'
+        })
+        export class MyComponent {}
+      `;
+
+        const templateUrlsResolver = new TemplateUrlsResolver();
+        const resolvedTemplateUrls = templateUrlsResolver.resolve(code, id);
+
+        expect(resolvedTemplateUrls).toMatchNormalizedPaths([
+          './app.component.html|/path/to/src/app.component.html',
+        ]);
       });
     });
   });
--- a/packages/vite-plugin-angular/src/lib/component-resolvers.ts
+++ b/packages/vite-plugin-angular/src/lib/component-resolvers.ts
@@ -1,10 +1,5 @@
 import { dirname, resolve } from 'node:path';
-import {
-  ArrayLiteralExpression,
-  Project,
-  PropertyAssignment,
-  SyntaxKind,
-} from 'ts-morph';
+import { parseSync } from 'oxc-parser';
 import { normalizePath } from 'vite';
 
 interface StyleUrlsCacheEntry {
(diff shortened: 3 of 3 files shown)


brandonroberts:
The three red `analog-docs` checks (`Pages changed`, `Header rules`, `Redirect rules`) are all sub-checks of a single failed Netlify deploy, and that failure does not come from this PR. I'm not pushing a change for it, so here's the evidence:

- **`apps/docs-app` has no `@analogjs` dependencies** — it's a standalone Docusaurus site, so the changed code (`packages/vite-plugin-angular/src/lib/component-resolvers.ts`) isn't in its build graph.
- **The `package.json` edit can't affect the install** — `packages/` are not pnpm workspace members (`pnpm-workspace.yaml` lists only `apps/docs-app`), which is also why `pnpm-lock.yaml` is unchanged here.
- **The docs site is byte-identical to its last green deploy.** `git diff 50c8851..24e6771 -- apps/docs-app pnpm-lock.yaml pnpm-workspace.yaml` returns 0 lines — nothing about the docs app or its dependency set has changed since #2452, whose `analog-docs` deploy preview succeeded.
- **It builds clean locally on this branch.** `nx build docs-app` exits 0 and generates static files for all locales; the only warnings are pre-existing broken anchors in the German translations (`/de/docs/integrations/ionic`, `/de/`), which are unrelated and present on `beta` too.
- **The Netlify project that actually exercises this change is green.** `analog-app` builds with `@analogjs/vite-plugin-angular` and therefore runs the rewritten resolvers — its deploy preview succeeded, as did `analog-blog`.

Every other check is green: `main`* / `build-windows`, CodeQL (both analyses), `apply-scope-metadata`, and the Nx Cloud `lint --all` and `prettier:check` runs.

A re-run of the `analog-docs` deploy should clear it. Happy to rebase onto a newer `beta` if that's easier for a maintainer to trigger.

<sub>* `main` was still in progress at the time of writing; I'll follow up if it doesn't pass.</sub>

---

PR #1957, opened 2025-11-18
feat(vite-plugin-angular): introduce support for Angular Compilation API

brandonroberts:
## PR Checklist

Closes #1955 

## What is the new behavior?

Support for using Angular's Compilation API is added under an experimental flag while the integration is stabilized. The compilation API gets feature and performance improvements, and supports parallel compilation through an environment variable.

When enabled, the Angular Compilation API is used to compile TypeScript files instead of manually constructing an instance of the Angular Compiler.

```ts
/// <reference types="vitest" />

import { defineConfig } from 'vite';
import analog from '@analogjs/platform';

// https://vitejs.dev/config/
export default defineConfig(({ mode }) => ({
  build: {
    target: ['es2020'],
  },
  resolve: {
    mainFields: ['module'],
  },
  plugins: [
    analog({
      vite: {
        experimental: {
          useAngularCompilationAPI: true
        }
      }
    })
  ],
  test: {
    globals: true,
    environment: 'jsdom',
    setupFiles: ['src/test-setup.ts'],
    include: ['**/*.spec.ts'],
    reporters: ['default'],
  },
  define: {
    'import.meta.vitest': mode !== 'production',
  },
}));
```

TODO:
- [ ] Add support for liveReload
- [ ] Add support for fileReplacements
- [ ] Add support for building libraries

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

PR #2488, opened 2026-08-13
fix(vite-plugin-angular): return worker path unchanged from processWebWorker

### LINKED ISSUE #2487: Web Worker URLs are rewritten to an empty string when useAngularCompilationAPI is enabled

```
@analogjs/vite-plugin-angular: 2.7.0
@angular/build: 22.0.0
Config: experimental.useAngularCompilationAPI: true
```

No StackBlitz link, since the app I hit this in isn't public. The steps below reproduce it on any app with the flag turned on, and the transformer output further down shows the cause directly.

vite-plugin-angular

### Description

With `experimental.useAngularCompilationAPI` on, every `new Worker(new URL('./x.worker', import.meta.url))` is rewritten at build time to `new URL('', import.meta.url)`.

An empty URL resolves to the chunk that imported it, so the worker ends up pointing at the application bundle instead of the worker file. The build itself succeeds, so the first sign of trouble is that the worker never does its job.

`SharedWorker` is rewritten the same way.

The cause is the `processWebWorker` hook in `angular-vite-plugin.ts`, which returns `''`:

```ts
processWebWorker(workerFile, containingFile) {
  return '';
},
```

Angular's worker transformer rewrites the `new URL()` argument whenever the value returned from that hook differs from the path passed in, and leaves the code alone when they match. `''` always differs, so every worker gets an empty path.

Running Angular's transformer directly against a service that creates a worker shows it:

```
returning ''         -> new SharedWorker(new URL("", import.meta.url), { type: 'module' })
returning workerFile -> new SharedWorker(new URL('./cache.worker', import.meta.url), { type: 'module' })
```

**Steps to reproduce**

1. Set `experimental.useAngularCompilationAPI: true` in `vite.config.ts`.
2. Create a worker from a service with `new Worker(new URL('./app.worker', import.meta.url), { type: 'module' })`.
3. Build for production.
4. Search the output for `new Worker(new URL(`. The first argument is `""`.

With the flag off, the same code builds fine and Vite emits the worker as its own asset.

### Other information

The stub has returned `''` since the Compilation API support landed in #1957 (v2.1.0), so this is not new in 2.7.0. Before #2452, the plugin decided whether to use Angular's output with a regex for Angular decorators, which meant a worker created inside a decorated class already hit this, while a worker in a plain `.ts` file did not. #2452 keyed that decision on emitted content instead, which is correct, and it widened the set of files this affects.

- [x] Yes

GagoM:
## PR Checklist

Closes #2487

## Affected scope

- Secondary scopes: none

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

## What is the new behavior?

`processWebWorker` returned `''` for every worker. Angular's worker transformer rewrites the `new URL()` argument whenever the returned path differs from the one passed in, so every worker URL became `new URL('', import.meta.url)`, which resolves to the importing chunk rather than the worker file.

This returns the path unchanged instead, which is how the transformer is opted out of. `@angular/build` does the same in its own esbuild plugin when a worker bundle fails ("Return the original path if the build failed"). With the path left alone, Vite handles `new Worker(new URL(..., import.meta.url))` natively.

Running Angular's transformer directly against a service that creates a worker:

```
returning ''         -> new SharedWorker(new URL("", import.meta.url), ...)
returning workerFile -> new SharedWorker(new URL('./cache.worker', import.meta.url), ...)
```

The hook can't just be dropped. `AngularHostOptions` declares it as required and the AOT and JIT compilations both call `.bind()` on it.

On blast radius: the `''` stub dates back to #1957 (v2.1.0), not 2.7.0. Before #2452 the plugin used a decorator regex to decide whether to serve Angular's output, so a worker created inside a decorated class was already affected, while one in a plain `.ts` file was not. #2452 switched that to emitted-content presence, which is correct and is not touched here, and it widened the set of affected files.

## Test plan

- [x] `nx format:check`
- [x] `nx test vite-plugin-angular` (690 passed, 6 skipped)
- [x] Manual verification

Manual check: ran Angular's worker transformer against both versions of the hook and compared the emitted code, as above. Also confirmed on a production build of an app using a SharedWorker, where the URL in the bundle now points at the worker and a cross-tab test that was failing passes on repeated runs.

No unit test added. The hook is now an identity function and the bug was in how its return value is read by Angular's transformer, which a test on a one-line function wouldn't catch.

- [ ] Yes
- [x] No

## Other information

The only people who could see a difference are those expecting Angular to bundle their workers, which never worked here since `''` produced a broken URL rather than a working bundle.

The branch is based on a slightly older `beta` commit (`adc387ce4`, 2.7.0-beta.12) rather than current `beta`. The change is a single line in a file untouched since, so it merges cleanly; happy to rebase if you'd prefer it on the tip.

### DIFF
--- a/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
+++ b/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
@@ -977,8 +977,12 @@ export function angular(options?: PluginOptions): Plugin[] {
 
           return stylesheetResult?.code || '';
         },
-        processWebWorker(workerFile, containingFile) {
-          return '';
+        // Angular's worker transformer only rewrites the `new URL()` specifier
+        // when the returned path differs from the input, so returning it
+        // unchanged opts out and lets Vite handle the worker natively.
+        // Returning '' rewrote every worker URL to `new URL('', import.meta.url)`.
+        processWebWorker(workerFile) {
+          return workerFile;
         },
       },
       (tsCompilerOptions) => {


brandonroberts (approved):
LGTM

PR #2440, opened 2026-07-24
feat(vite-plugin-nitro)!: remove deprecated useAPIMiddleware and API …

luishcastroc:
## PR Checklist

Removes the deprecated `useAPIMiddleware` option and the generated `#ANALOG_API_MIDDLEWARE`
proxy middleware from `@analogjs/vite-plugin-nitro` (and the parallel Nitro plugin in
`@analogjs/platform`). API routes are served exclusively through Nitro's native file-based
routing under `src/server/routes/api`.

Closes #2429

## Affected scope

- Primary scope: **vite-plugin-nitro**
- Secondary scopes: **platform** (its `analog-nitro-plugin` / `get-page-handlers` / `renderers` mirror the same code)

Docs: a migration note was added to `docs/guides/migrating-v2-to-v3.md`.

## Recommended merge strategy for maintainer [optional]

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

## Commit preservation note [optional]

N/A — single focused change.

## What is the new behavior?

`@analogjs/vite-plugin-nitro` previously supported three ways of handling the `/api` prefix:

1. `hasAPIDir === true` — a `src/server/routes/api` directory exists; Nitro serves those routes natively.
2. `hasAPIDir === false && useAPIMiddleware === true` (the default) — injects the generated `#ANALOG_API_MIDDLEWARE`, which strips the `/api` prefix and forwards the request via `fetchWithEvent`/`proxyRequest`.
3. `hasAPIDir === false && useAPIMiddleware === false` — falls back to a Nitro `routeRules` proxy (`{ '/api/**': { proxy: { to: '/**' } } }`).

Paths 2 and 3 are removed. The three code paths collapse into one: API routes are served solely via native Nitro file-based routing (path 1). Page-endpoint handlers (`.server.ts`) are now registered unconditionally under the `/api` prefix — matching the surviving native-routing behavior, which was previously gated on the presence of an API directory.

This removes a generated virtual module along with its dev/build/prerender wiring, and the `useAPIMiddleware` option (already marked `@deprecated`).

## Test plan

- [x] `nx format:check` — prettier ran on staged files via the pre-commit hook
- [x] `pnpm build` — `nx run-many --target=build --projects=vite-plugin-nitro,platform` (affected packages)
- [x] `pnpm test` — `nx run-many --target=test --projects=vite-plugin-nitro,platform` (72 + 361 tests pass); `nx run-many --target=lint` also green
[...]

PR #2489, opened 2026-08-14
chore: merge alpha into beta

brandonroberts:
## PR Checklist

Branch sync PR: brings the `alpha` release line into `beta`. No code is authored on this branch — it points at `alpha`'s tip (`64edad67`), and `beta` is a strict ancestor, so this merges cleanly.

Closes #

## Affected scope

Every v3 package, since this is a release-line sync rather than a focused change.

- Primary scope: `platform`
- Secondary scopes: `router`, `content`, `content-plugin`, `vite-plugin-angular`, `vite-plugin-nitro`, `vitest-angular`, `nx-plugin`, `create-analog`, `storybook-angular`, `astro-angular`, plus docs and dogfooding apps

- [ ] Squash merge
- [ ] Rebase merge
- [x] Other — plain merge commit

Squashing would collapse the alpha line's commits — including the `chore: release 3.0.0-alpha.*` commits that semantic-release depends on — into a single opaque commit, breaking version continuity on `beta` and losing the per-package boundaries alpha already maintains.

## What is the new behavior?

`beta` moves from the `main` (v2) tree to the current v3 alpha tree. Highlights:

- `feat(platform)`: migrate to `nitro/vite`; split `analog()` into `analog()` + `angular()` + `nitro()` (#2343)
- `feat`: bump to Angular v22 packages
- `feat(router)!`: remove server components
- `feat(vite-plugin-nitro)!`: remove deprecated `useAPIMiddleware` and API proxy middleware (#2440)
- `feat(router)`: `definePageLoadQueries` helper, route-data `__analogQueries` hydration, TransferState mirroring of load-prefetched state
- `refactor(vite-plugin-angular)`: split the monolithic plugin into self-contained plugins and consolidate tsconfig resolution behind `TsconfigResolver`
- `build`: `angular-compiler` integrated into the build pipeline; Nitro `3.0.260603-beta`; Nx 23.0.1
- Docs: package guides synced to alpha's current API and package set

Diff against `beta`: ~1,075 files changed.

### Pre-existing breakages this sync surfaced

The first CI run exposed several failures that already existed on `alpha` but had never been exercised together. All fixes landed **on `alpha`**, and this branch was re-pointed at alpha's tip each time, so it stays a pure sync:

| Fix | Commit | Result |
| --- | --- | --- |
| `fix(vite-plugin-nitro)`: emit the `server-fn-id` export as a build entry | `d7dc36a` | `build-windows` green |
| `ci`: replace the Angular compiler overrides instead of duplicating them | `5a3b0a5` | compiler-compat installs again |
| `test(vite-plugin-angular)`: gate the optional-chaining emitter tests on v22 | `8ce6dde` | matrix no longer fails on absent AST fields |
| `chore`: lint the Angular projects through angular-eslint flat configs | `64edad6` | `nx run-many --target lint` passes locally for all 18 projects |

Alpha auto-released `3.0.0-alpha.70` from the first two.

### Known remaining failures

- **`main` → Lint.** The distributed lint terminates with `CI pipeline execution … cancelled. Reason: One or more workflows failed`, with no task output in the coordinator log, at a consistent ~1m20–1m47 across every run. In Nx Cloud DTE that message refers to the **agent** workflows from `.nx/workflows/agents.yaml`, not to lint tasks. The same lint command passes locally for all 18 projects with `--skip-nx-cache`, and `pnpm install --frozen-lockfile --prefer-offline` succeeds locally, so the lockfile and the lint rules are not the cause. The agent logs are needed to pin down which init step dies.
- **`compiler-compat` on the Angular 17/18 slots.** Fails during `prepare` with `Cannot find module '@angular-devkit/build-angular/src/tools/esbuild/angular/compiler-plugin.js'`. The workflow pins only `@angular/compiler` and `@angular/compiler-cli` while the rest of the workspace stays on v22, so the devkit resolves to a mismatched layout. Fixing it means reworking how the matrix installs — separate from this sync.

## Test plan

Run locally against the alpha tree:

- [x] `nx build vite-plugin-nitro` — artifact validation passes; all three export targets resolve under `dist/src/`
- [x] `nx lint vite-plugin-nitro`
- [x] `nx run-many --target lint --all --exclude=content,platform,router` — 18/18 projects pass
- [x] `vitest run js-emitter.spec.ts` — 29/29 pass on Angular 22; the v22 gate verified false against `@angular/compiler@19`
- [x] `pnpm install --frozen-lockfile --prefer-offline` — succeeds, `prepare` build clean
- [x] `prettier --check` on every changed file
- [x] compiler-compat patch script exercised against the real `pnpm-workspace.yaml`, plus the beta-style block and the no-`overrides:` fallback

`build-windows` is green on CI across two runs. The rest of the tree is byte-identical to `alpha`.

- [x] Yes
- [ ] No

Relative to `beta`'s previous (v2) contents, this carries the full set of v3 breaking changes: Angular v22 is required, server components are removed from `@analogjs/router`, `useAPIMiddleware` and the API proxy middleware are removed from `@analogjs/vite-plugin-nitro`, and `analog()` is split into `analog()` + `angular()` + `nitro()` in `@analogjs/platform`. Migration paths are covered by the v3 docs carried in this merge.

## Other information

- Confirm the release configuration points `beta` at the intended prerelease channel before merging — the version on this branch is `3.0.0-alpha.70`.
- CodeRabbit skipped its review (1,041 files against a 300-file limit), and CodeQL reports size-related noise with its own caveat that alerts may surface because the diff is too large. Both are expected for a sync PR.


benpsnyder:
awwwwwwww yeeeeeeeea
image

PR #2493, opened 2026-08-22
Fix hmr plus :host styles in .scss issue

### LINKED ISSUE #2492: :host styles are not properly scoped when `liveReload: true`

angular: 21.2.21
analogjs: 2.7.0
vite 8.2.1

vite-plugin-angular

### Description

We have a large app that uses component .scss files that use `:host { ... }` styles. 

These styles dont work when we set `liveReload: true` for the compiler

To work around it we added a custom plugin after the angular on in the vite config: `encapsulateAnalogComponentStyles`. It looks like:

```typescript
/**
 * Analog liveReload serves component styles as `<link>` URLs with `ngcomp`/`e`
 * query params. Analog's own encapsulateStyle transform never runs on those
 * files (TS-only filter, and it compares encapsulation to the string
 * `'emulated'` instead of ViewEncapsulation.Emulated). Without this, `:host`
 * rules are invalid CSS and the layout collapses.
 */
function encapsulateAnalogComponentStyles(): Plugin {
  return {
    name: 'encapsulate-analog-component-styles',
    enforce: 'post',
    transform(code, id) {
      if (!id.includes('ngcomp=')) {
        return null;
      }
      const url = new URL(id, 'http://localhost');
      const encapsulation = Number.parseInt(url.searchParams.get('e') ?? '', 10);
      const componentId = url.searchParams.get('ngcomp')?.replaceAll('%5E', '^');
      // ViewEncapsulation.Emulated === 0
      if (encapsulation !== 0 || !componentId) {
        return null;
      }
      return {
        code: encapsulateViteCssOutput(code, componentId),
        map: null,
      };
    },
  };
}

function encapsulateViteCssOutput(code: string, componentId: string): string {
  const VITE_CSS_CONST_RE = /const __vite__css\s*=\s*("(?:\\.|[^"\\])*")/;
  const match = VITE_CSS_CONST_RE.exec(code);
  if (match) {
    const css = JSON.parse(match[1]) as string;
    const encapsulated = encapsulateStyle(css, componentId);
    return (
      code.slice(0, match.index) +
      `const __vite__css = ${JSON.stringify(encapsulated)}` +
      code.slice(match.index + match[0].length)
    );
  }
  return encapsulateStyle(code, componentId);
}
```

However when applied the app loads completely scrambled  (we use :host on some high level components) and then things start to snap into place. Would be great if it can be fixed in `vite-angular-plugin`

```true
App layout is incorrect, no errors occur.
```

### Other information

_No response_

- [x] Yes
- [ ] No

arnoudb:
Disclaimer. I lack technical knowledge of the analogjs repo. I just identified the problem and had AI help me to fix the issue. So i completely understand if you dont like it and want to solve it differently. In that case the branch may still be helpful to document information about the error.

## PR Checklist

Closes #2492

With `liveReload: true`, component styles served as `<link>` URLs (`ngcomp` / `e` query params) never went through Analog's encapsulate transform. `:host` / `:host-context` in external SCSS were invalid CSS, so the first paint collapsed until HMR recovered. Enabling external styles on the first rebuild after that also reused the inlined `NgtscProgram`, which then threw `Unable to locate component resource: <hash>.scss` on template edits.

## Affected scope

- Secondary scopes: none

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

## What is the new behavior?

`liveReload: true` now uses a two-phase style pipeline:

1. **First paint** inlines and encapsulates component styles the normal Angular way, so `:host` in SCSS/CSS applies immediately. No more unstyled first load.
2. **Later rebuilds** emit external `ngcomp` stylesheets so CSS can hot-update independently. A dedicated post plugin (`analogjs-encapsulate-component-styles`) runs Angular `encapsulateStyle` *after* Vite compiles SCSS/Sass/Less, including Vite's `const __vite__css = "..."` wrapper.
3. **First edit after load** discards the incremental `NgtscProgram` / host once when `externalRuntimeStyles` turns on. That avoids mixing inlined-program flags with hash-remapped style URLs (`Unable to locate component resource: <hash>.scss`). Later edits reuse the new program.

Apps that do not set `liveReload: true` are unchanged. The local `encapsulateAnalogComponentStyles()` workaround is no longer needed (see issue description for the custom encapsulateAnalogComponentStyles hack)

There are 2 commits. the first fixes the layout issue, the second fixes the jumpiness of the UI on initial load as the scoped styles are applied in an afterthought in the first commit. 

## Test plan

- [ ] `nx format:check`
- [ ] `pnpm build`
- [ ] `pnpm test`
- [x] Manual verification

Targeted verification:

- [x] `pnpm nx test vite-plugin-angular` — 716 passed (encapsulate plugin + liveReload compiler-option / program-reset tests)
- [x] `pnpm exec tsc -p packages/vite-plugin-angular/tsconfig.lib.json --noEmit`
- [x] `pnpm nx build vite-plugin-angular`
- [x] Manual: app first load is styled; template edit after first paint no longer throws the missing `<hash>.scss` resource error

- [ ] Yes
- [x] No

## Other information

Branch: `fix-hmr-plus-host-style-in-scss` (already on origin).

Commits to squash:

- `fix(vite-plugin-angular): extract encapsulated component styles logic to a dedicated plugin`
- `feat(vite-plugin-angular): add liveReload compiler options for first-pass inline styles and HMR`

Pipeline details are in `packages/vite-plugin-angular/src/lib/compiler/COMPILER.md`.


brandonroberts:
## Perf impact — PR #2493

Validated against `beta` @ `305f9f0` with instrumented compilations, timed test runs, and a live browser session.

- **Production builds:** none. The compiler flags no-op outside watch mode, and the new plugin is `apply: 'serve'` so it never runs in builds.
- **Non-liveReload / fastCompile apps:** none. Everything is gated on `liveReload: true` on the classic compile path.
- **Test environments:** negligible. `isTest` forces liveReload off, leaving only a per-module substring check. Measured (`--skip-nx-cache`): base 690 tests / 18.0s vs PR 716 tests / 17.1s — within noise.
- **liveReload dev sessions:** one cost, one gain.
  - Cost: the first edit of each session discards the incremental `NgtscProgram` and pays a full recompile (seconds, scales with app size; required to avoid `Unable to locate component resource`). Happens exactly once per session; later edits are incremental.
  - Gain: first paint inlines styles — one fewer request per component stylesheet and no unstyled flash. Startup still runs a single compilation; per-stylesheet encapsulation after the flip is microsecond-level.

---

brandonroberts (approved):
LGTM

---

brandonroberts:
Thanks @arnoudb!

PR #2495, opened 2026-08-22
fix(router): guard import.meta.env access in streaming SSR renderer

### LINKED ISSUE #2494: SSR crashes at startup with "Cannot read properties of undefined (reading 'PROD')" in @analogjs/router/server — affects any project using .server.ts load files (present in stable 2.7.0)
## Description

SSR apps crash immediately at server startup with:

```
TypeError: Cannot read properties of undefined (reading 'PROD')
    at .../@analogjs/router/fesm2022/analogjs-router-server.mjs:626:21
    at ModuleJob.run (node:internal/modules/esm/module_job:343:25)
    at process.processTicksAndRejections (node:internal/process/task_queues:103:5)
    at async onImport.tracePromise.__proto__ (node:internal/modules/esm/loader:681:26)
    at async asyncRunEntryPointWithESMLoader (node:internal/modules/run_main:117:5)
```

This reproduces with a plain `PageServerLoad` (e.g. `(home).server.ts` exporting only `load`) — no Server Functions involved.

## Root cause

- `packages/router/server/src/render-stream.ts` does an unguarded `import.meta.env.PROD` / `import.meta.env.DEV` access at module top-level. `render.ts` was already patched for this exact issue in #2433 (optional chaining, with the comment "the server-function dispatch endpoint imports this entry from a Nitro bundle, where `import.meta.env` is not defined at all") — `render-stream.ts` (added the same day in #2419, for Streaming SSR) never received the same fix.
- `packages/vite-plugin-nitro/src/lib/utils/get-server-fn-handlers.ts` globs `**/*.server.ts` and only excludes `app.config.server.ts`. It doesn't distinguish plain `PageServerLoad` files from actual server-function hosts, so `hasServerFns` becomes `true` for any project using the standard `.server.ts` load convention — wiring up the `/_analog/fn/:id` dispatch handler even when the project defines no server function at all.
- That dispatch handler imports `@analogjs/router/server`, which bundles both `render` and `renderStream` into a single FESM file. Loading it outside Vite's SSR pipeline hits the unguarded `render-stream.ts` code and crashes the whole process before it can serve any request.

Net effect: virtually any SSR project using `injectLoad`/`PageServerLoad` is affected, not just Server Functions users — this is a very common pattern, not an edge case.

## Affected versions

- `@analogjs/router@2.7.0` (npm `latest`, published 2026-08-12)
- `@analogjs/router@2.7.1-beta.1` / `2.7.1-beta.2`

## Repro

1. In the analogjs/analog monorepo: `nx serve analog-app` (has both `.server.ts` load files and Server Functions wired) — crashes at boot.
2. Or in any minimal Analog SSR app: add a `(page).server.ts` exporting `load`, run the dev server.

## Suggested fix

Apply the same optional-chaining guard already used in `render.ts` to `render-stream.ts`:

```diff
- if (import.meta.env.PROD) {
+ if (import.meta.env?.PROD) {
    enableProdMode();
  }
...
- if (warnedMissingPrimitive || !import.meta.env.DEV) return;
+ if (warnedMissingPrimitive || !import.meta.env?.DEV) return;
```

PR incoming.

oliv37:
## PR Checklist

Closes #2494

## Affected scope

- Primary scope: router
- Secondary scopes:

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

## What is the new behavior?

`render-stream.ts` accessed `import.meta.env.PROD` / `import.meta.env.DEV` unguarded at module top-level. `render.ts` (co-exported from the same `server/src/index.ts` barrel into a single FESM bundle) was already patched for this exact case in #2433, with the comment explaining that the server-function dispatch endpoint loads this entry from a Nitro bundle where `import.meta.env` is not defined at all. `render-stream.ts` was added the same day in #2419 (Streaming SSR) and never received the same fix.

Because `getServerFnHandlers` (in `vite-plugin-nitro`) globs `**/*.server.ts` and only excludes `app.config.server.ts`, any project using the standard `PageServerLoad` `.server.ts` convention (not just actual Server Functions) causes `hasServerFns` to be `true`, wiring up the `/_analog/fn/:id` dispatch endpoint. That endpoint imports `@analogjs/router/server`, which crashes the whole SSR process at boot before serving any request — see #2494 for the full repro and root-cause writeup.

This PR applies the same optional-chaining guard already established in `render.ts` to `render-stream.ts`, at both call sites.

## Test plan

- [x] `nx build router` (builds clean, verified the guard is present in the emitted FESM output)
- [x] `nx format:check` on the changed file
- [ ] `pnpm test` (not run for this narrow change; happy for CI to confirm)
- [x] Manual verification: reproduced the crash against the monorepo's own `analog-app` (which has both `.server.ts` load files and Server Functions), rebuilt `@analogjs/router`, confirmed the guarded code path in the emitted bundle

- [ ] Yes
- [x] No

## Other information

Minimal, targeted fix mirroring the existing pattern in `render.ts`. Filed #2494 with the full root-cause analysis, including why this affects any SSR project with a `.server.ts` load file, not just Server Functions users.

### DIFF
--- a/packages/router/server/src/render-stream.ts
+++ b/packages/router/server/src/render-stream.ts
@@ -52,7 +52,9 @@ import { afterBodyOpen, bodyInner, headInner } from './utils/stream-html';
 import { isLikelyBot, streamingDisabledByRoute } from './utils/stream-request';
 import { DEFER_RECONCILE_RUNTIME } from './defer-reconcile-runtime';
 
-if (import.meta.env.PROD) {
+// Optional chaining: the server-function dispatch endpoint imports this entry
+// from a Nitro bundle, where `import.meta.env` is not defined at all.
+if (import.meta.env?.PROD) {
   enableProdMode();
 }
 
@@ -116,7 +118,7 @@ function installCaptureDispatcher(): void {
 
 let warnedMissingPrimitive = false;
 function warnMissingPrimitiveOnce(): void {
-  if (warnedMissingPrimitive || !import.meta.env.DEV) return;
+  if (warnedMissingPrimitive || !import.meta.env?.DEV) return;
   warnedMissingPrimitive = true;
   console.warn(
     '[@analogjs/router] renderStream: the streaming hook was not found on ' +


brandonroberts (approved):
LGTM

Validated against a fresh `create-analog` app on `@analogjs/router@2.7.0` with a single `PageServerLoad` file and no server functions: reproduced the crash in both the Nitro dev worker (boot failure, all load endpoints 500) and the production `/_analog/fn/:id` route, then confirmed this exact two-line guard fixes both — loads return 200 and the dispatch route degrades to a proper 404. Matches the guard `render.ts` and `provide-server-context.ts` already carry from #2433.

---

brandonroberts:
Thanks @oliv37!

Issue #2293, opened 2026-04-15
fix(vite-plugin-angular): @apply inside :host fails in component CSS with tailwindCss prefix configuration


muhammaduxair:
## Summary

Component CSS that uses `@apply` with a prefixed Tailwind utility inside `:host { }` silently fails during dev serve. The same `@apply` directives work correctly when placed at the top level (outside `:host`). This affects any workspace using `tailwindCss: { prefixes: ['...'] }` with Angular's Compilation API (`useAngularCompilationAPI: true`).

Discovered during the Analog alpha.36 upgrade on a lib-heavy Angular monorepo running on Windows with Tailwind CSS v4.2 and the `prefix(sa)` configuration.

## Problem

Given a component CSS file:

```css
:host {
  @apply sa:relative sa:z-10;
  .mos-version-info {
    @apply sa:flex sa:items-center sa:gap-1 sa:text-xs;
  }
}
```

And a Vite config:

```ts
analog({
  experimental: { useAngularCompilationAPI: true },
  vite: {
    tailwindCss: {
      rootStylesheet: repoConfig.resolveProject('src/app/styles/main.css'),
      prefixes: ['sa:'],
    },
  },
});
```

Where `main.css` contains:

```css
@import 'tailwindcss' prefix(sa);
```

**Observed behavior:**

| CSS structure | Clean server restart | Result |
|---|---|---|
| `:host { @apply sa:flex; }` | Yes | Styles not applied |
| `.selector { @apply sa:flex; }` (no `:host`) | Yes | Works correctly |
| Raw CSS inside `:host` (no `@apply`) | Yes | Works correctly |
| `sa:flex` in HTML template class attribute | N/A | Works correctly |

All tests performed with full cache clears (`node_modules/.vite`, `.analog`, Nx cache) and clean server restarts between each test.

### What works

- The `buildStylePreprocessor` correctly detects `sa:` usage and injects `@reference "main.css"` into component CSS
- The stylesheet registry stores the preprocessed CSS with `@reference` present
- The `load` hook serves CSS that includes the `@reference` directive
- Global `main.css` compiles correctly — template-level Tailwind classes (`class="sa:flex"`) work fine
- `@apply` without `:host` wrapper compiles correctly through the full pipeline

### What fails

When `:host` wraps the `@apply` directives, the styles are not applied in the browser despite the server-side pipeline appearing to process them. The exact failure point is unclear but is somewhere between Analog's stylesheet registry serving the CSS and the browser applying it.

[...]

PR #2504, opened 2026-09-02
fix(vite-plugin-angular): fix the Angular Compilation API dev path

brandonroberts:
## PR Checklist

`useAngularCompilationAPI: true` was unusable with `analog()` in dev on `alpha`. Found while validating the `analog.setup()` interop model from [RFC: Nitro-style Vite plugin interop hooks analogjs/analog#2254](https://github.com/analogjs/analog/issues/2254) against `apps/tailwind-debug-app`, whose committed config uses the Compilation API path. Related: [@apply inside :host fails with tailwindCss prefix analogjs/analog#2293](https://github.com/analogjs/analog/issues/2293), which the debug app exists to reproduce.

Closes #

## Affected scope

- Secondary scopes: none

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

Not needed; single commit.

## What is the new behavior?

Two defects, the second only visible once the first was fixed:

1. **Raw TypeScript reached the browser.** `@analogjs/platform`'s deps plugin excludes `**/*.ts` from Vite's own transform so the Angular plugin owns compilation, but the Compilation API transform returned nothing for files without an Angular decorator. Files such as `app.config.ts` and the generated route tree were served untranspiled and the page died on `import type`. Angular already emits every file in the program, so the transform now returns that output for any emitted file. The "not in the TypeScript program" warning stays decorator-specific.

2. **`:host` styles vanished.** `encapsulationPlugin` is registered once by the `angular()` factory and was gated on the ngtsc plugin's `shouldExternalizeStyles()`. That closure reads a `watchMode` variable assigned only in the ngtsc `config` hook, which never runs in Compilation API mode, so the predicate was always false and `:host` was never rewritten to `[_nghost-...]`. Fetching the component stylesheet with `Accept: text/css` on that path returned raw `:host { ... }`. The `ngcomp=` query only exists on requests Angular's runtime makes for externalized component styles, so it is now the only gate and the plugin no longer depends on either compilation path's private state.

## Test plan

- [x] `nx format:check` on changed files
- [x] `vitest run` for `vite-plugin-angular`: 865 passed, 6 skipped, 3 new tests (`compilation-api-plugin.spec.ts`, new `encapsulation-plugin.spec.ts`)
- [x] Manual verification: `apps/tailwind-debug-app-e2e` `host-apply.spec.ts` against the dev server on the app's committed config (Compilation API on, live reload on, PostCSS and route-tree import intact). Before: parse error overlay, then after fix 1 the `:host` background assertion failed. After: 2 passed. Also 2 passed with live reload off and Tailwind provided through an `analog.setup()` plugin instead of the `tailwindCss` option.

`tsc --noEmit` for the package reports the same 23 pre-existing errors as `alpha`.

- [ ] Yes
- [x] No

`encapsulationPlugin()` drops its predicate parameter; it is internal to the package.

## Other information

The factory-level `liveReloadPlugin` in `angular()` is registered with the ngtsc plugin's `classNames` and `fileEmitter` maps, which stay empty in Compilation API mode. It is inert there rather than broken, so this PR leaves it alone.

https://claude.ai/code/session_01KAcJKwrVLs2yyyy4Scnmv9

---
_Generated by [Claude Code](https://claude.ai/code/session_01KAcJKwrVLs2yyyy4Scnmv9)_

### DIFF
--- a/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
+++ b/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
@@ -1544,7 +1544,7 @@ export function angular(options?: PluginOptions): Plugin[] {
     routerPlugin(),
     angularFullVersion < 190004 && pendingTasksPlugin(),
     nxFolderPlugin(),
-    encapsulationPlugin(shouldExternalizeStyles),
+    encapsulationPlugin(),
   ].filter(Boolean) as Plugin[];
 
   function resolveTsConfigPath() {
--- a/packages/vite-plugin-angular/src/lib/compilation-api/compilation-api-plugin.spec.ts
+++ b/packages/vite-plugin-angular/src/lib/compilation-api/compilation-api-plugin.spec.ts
@@ -435,4 +435,67 @@ describe('compilationAPIPlugin', () => {
     expect(result).toBeDefined();
     expect(result.code).toBe('compiled output');
   });
+
+  it('serves emitted output for TypeScript files without Angular decorators', async () => {
+    const configFile = join(tempRoot, 'src/app.config.ts');
+    createAngularCompilationMock.mockResolvedValue({
+      initialize: vi.fn().mockResolvedValue({
+        externalStylesheets: new Map(),
+        templateUpdates: new Map(),
+      }),
+      update: vi.fn(),
+      diagnoseFiles: vi.fn().mockResolvedValue({ errors: [], warnings: [] }),
+      emitAffectedFiles: vi
+        .fn()
+        .mockResolvedValue([
+          { filename: configFile, contents: 'export const appConfig = {};' },
+        ]),
+    });
+
+    const { compilationAPIPlugin } =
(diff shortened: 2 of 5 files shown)


brandonroberts:
CI status on `c4b4773`: the `ci / main` job is red, but no task in this PR failed.

The Lint step's distributed execution was terminated by Nx Cloud with "One or more workflows failed" before any lint task ran (the orchestrator log shows only `Distributed Execution Started` → `Terminated`). That means one of the `analog-linux` agents died in its init steps. `build-windows`, bundle size, Prettier, and both Netlify previews passed on the same commit, and the previous alpha-based PR run passed Lint on this workflow. The diff only touches `packages/vite-plugin-angular` source and specs.

I don't have permission to re-run the failed jobs from this session (403 on the re-run endpoint). A re-run of the failed jobs on [run 33694642433](https://github.com/analogjs/analog/actions/runs/33694642433) should clear it; if it fails again on a real task I'll fix it.

---

PR #2506, opened 2026-09-03
fix(vite-plugin-angular): preserve source maps across TypeScript transforms

### LINKED ISSUE #2505: Source maps are lost or corrupted by repeated TypeScript transforms

analogjs 2.7.1
vite 8.2.2
angular 21.2.20

vite-plugin-angular

### Description

We upload the sourcemaps to sentry but noticed that the errors pointed to the wrong line numbers in the source.

Production source maps generated by Analog can become incorrectly composed as Angular
applications pass through compilation, optimization, and Vite’s TypeScript processing.

The Angular build optimizer plugin explicitly configures Vite’s oxc or esbuild transform. This can cause TypeScript processing to run again after Angular has already generated JavaScript and a corresponding source map. The additional transform changes the generated output without reliably preserving the existing mappings.

Other parts of the compilation pipeline also contribute to the problem:
- Production Angular compilation does not consistently emit source maps when Vite’s build.sourcemap option is enabled.
- The Angular optimizer may generate or update an inline source map, but that map is not always returned to Vite.
- In `fastCompile` mode, the map returned by Angular represents the code before TypeScript syntax is stripped. Removing type annotations and modifiers then shifts generated positions without composing those changes into the map.
- Relative source paths and sourceRoot values may remain unresolved, making original TypeScript and template sources difficult for downstream tooling to locate.

As a result, browser developer tools, stack traces, breakpoints, and error-reporting services can point to the wrong line, or column—or fail to resolve the original source entirely.

```true
[plugin @analogjs/vite-plugin-angular-optimizer] [SOURCEMAP_BROKEN] Sourcemap is likely to be incorrect: a plugin (@analogjs/vite-plugin-angular-optimizer) was used to transform files, but didn't generate a sourcemap for the transformation. Consult the plugin documentation for help: https://rolldown.rs/guide/troubleshooting#warning-sourcemap-is-likely-to-be-incorrect

was the smoking gun
```

### Other information

_No response_

- [x] Yes
- [ ] No

arnoudb:
### PR Checklist

Closes #2505

Production source maps could become missing, stale, or incorrectly composed as modules passed through Angular compilation, the Angular build optimizer, and Vite’s TypeScript processing.

The most important issue was that the optimizer plugin explicitly configured Vite’s `oxc`/`esbuild` transform. This could cause TypeScript processing to run again over code and source maps already generated by Angular. Other stages either disabled production maps, discarded optimizer maps, or returned a compiler map that did not account for the final TypeScript syntax-stripping transform. Consequently, stack traces, breakpoints, and error-reporting tools could resolve to incorrect source positions.

### Affected scope

- Secondary scopes: none

### Recommended merge strategy for maintainer

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

### What is the new behavior?

This PR preserves a coherent source-map chain throughout the Angular build pipeline:

1. **Avoid duplicate TypeScript transforms**
   - The optimizer plugin no longer configures `oxc` or `esbuild` itself.
   - Angular production constants such as `ngDevMode` and `ngJitMode` remain configured through Vite’s top-level `define` option.
   - Vite remains responsible for its built-in TypeScript processing, avoiding another Oxc/esbuild pass over Angular-generated output and its source map.

2. **Retain Angular optimizer maps**
   - When `build.sourcemap` is enabled, the Angular `JavaScriptTransformer` now processes both generated and third-party source maps.
   - Inline maps produced while linking or optimizing Angular packages are extracted and returned directly to Vite rather than being left embedded in the transformed code.
   - Existing maps for modules that do not require Angular optimization remain in the transform chain. Source-map comments are only removed when maps are intentionally disabled.

3. **Emit maps from production Angular compilation**
   - Production compilation now honors Vite’s `build.sourcemap` setting instead of always disabling TypeScript source-map generation.
   - Source content is included so downstream tools can resolve original TypeScript without depending on separately deployed source files.
   - Relative source entries and `sourceRoot` are normalized into resolvable paths before maps are passed to Vite. This also keeps mappings for component code and external templates usable by later transforms.

4. **Compose maps in the fast-compile pipeline**
   - Fast compilation still performs its required Oxc/esbuild pass to remove TypeScript-only syntax; this is not an additional transform introduced by the PR.
   - The Angular compiler or JIT map is now supplied as the input map to that syntax-stripping pass.
   - The resulting composed map is returned instead of the stale pre-strip compiler map, so removed annotations and modifiers no longer shift reported source positions.
   - Both the Oxc path and the esbuild fallback are covered.

Together, these changes ensure each necessary transform runs once and contributes to a single composed map from final JavaScript back to the original Angular/TypeScript source.

Source-map generation remains opt-in for production through `build.sourcemap`. Builds that do not request production maps retain their existing behavior, and no public API changes are introduced.

### Test plan

- [ ] `nx format:check`
- [ ] `pnpm build`
- [ ] `pnpm test`
- [ ] Manual verification

Targeted verification:

- [x] `pnpm nx test vite-plugin-angular` — 724 passed, 6 skipped
- [x] Focused source-map tests covering optimizer map extraction and retention, production compiler maps, normalized source paths, Oxc composition, and the esbuild fallback
- [x] End-to-end fast-compile assertion confirming generated positions map back to the correct original TypeScript columns after type annotations are removed
- [x] `pnpm nx build vite-plugin-angular`

### Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

### DIFF
--- a/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
+++ b/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
@@ -254,10 +254,9 @@ export function angular(options?: PluginOptions): Plugin[] {
   let liveReloadProgramHasExternalStyles = false;
   const declarationFiles: DeclarationFile[] = [];
   const fileTransformMap = new Map<string, string>();
-  let styleTransform: (
-    code: string,
-    filename: string,
-  ) => Promise<vite.PreprocessCSSResult>;
+  let styleTransform:
+    | ((code: string, filename: string) => Promise<vite.PreprocessCSSResult>)
+    | undefined;
   let pendingCompilation: Promise<void> | null;
   let compilationLock = Promise.resolve();
   // Persistent Angular Compilation API instance. Kept alive across rebuilds so
@@ -267,6 +266,16 @@ export function angular(options?: PluginOptions): Plugin[] {
   let angularCompilation:
     | Awaited<ReturnType<typeof createAngularCompilationType>>
     | undefined;
+
+  function getStyleTransform() {
+    return (
+      styleTransform ??
+      (async (code: string) => ({
+        code,
+        deps: new Set<string>(),
+      }))
+    );
+  }
 
   function angularPlugin(): Plugin {
     let isProd = false;
@@ -381,7 +390,8 @@ export function angular(options?: PluginOptions): Plugin[] {
       async buildStart() {
         if (!jit) {
           const pluginContext = this;
-          const cssTransformHook = this.environment.config.plugins.find(
+          const environmentConfig = this.environment?.config ?? resolvedConfig;
+          const cssTransformHook = environmentConfig.plugins?.find(
             (plugin) => plugin.name === 'vite:css',
           )?.transform;
           const cssTransform = (
@@ -406,24 +416,22 @@ export function angular(options?: PluginOptions): Plugin[] {
                   >)
             | undefined;
 
-          if (!cssTransform) {
-            throw new Error('Unable to locate the Vite CSS transform hook');
-          }
+          if (cssTransform) {
+            styleTransform = async (code: string, filename: string) => {
+              const result = await cssTransform.call(
+                pluginContext,
+                code,
+                filename,
+              );
 
-          styleTransform = async (code: string, filename: string) => {
-            const result = await cssTransform.call(
-              pluginContext,
-              code,
-              filename,
-            );
-
-            return {
-              code:
-                typeof result === 'string' ? result : (result?.code ?? code),
-              map: typeof result === 'object' ? result?.map : undefined,
-              deps: new Set<string>(),
+              return {
+                code:
+                  typeof result === 'string' ? result : (result?.code ?? code),
+                map: typeof result === 'object' ? result?.map : undefined,
+                deps: new Set<string>(),
+              };
             };
-          };
+          }
         }
 
         // Defer the first compilation in test mode
@@ -1055,7 +1063,10 @@ export function angular(options?: PluginOptions): Plugin[] {
           let stylesheetResult;
 
           try {
-            stylesheetResult = await styleTransform(data, `${filename}?direct`);
+            stylesheetResult = await getStyleTransform()(
+              data,
+              `${filename}?direct`,
+            );
           } catch (e) {
             console.error(`${e}`);
           }
@@ -1390,7 +1401,7 @@ export function angular(options?: PluginOptions): Plugin[] {
       externalComponentStyles = tsCompilerOptions['externalRuntimeStyles']
         ? new Map()
         : undefined;
-      augmentHostWithResources(host, styleTransform, {
+      augmentHostWithResources(host, getStyleTransform(), {
         inlineStylesExtension: pluginOptions.inlineStylesExtension,
         isProd,
         inlineComponentStyles,
--- a/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
+++ b/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
@@ -370,7 +370,7 @@ export function angular(options?: PluginOptions): Plugin[] {
         const invalidateCompilationOnFsChange = createFsWatcherCacheInvalidator(
           invalidateFsCaches,
           invalidateTsconfigCaches,
-          () => performCompilation(resolvedConfig),
+          () => performCompilation(server.environments.client.config),
           pluginOptions.include.map(
             (glob) =>
               `${normalizePath(resolve(pluginOptions.workspaceRoot))}${glob}`,
@@ -387,7 +387,9 @@ export function angular(options?: PluginOptions): Plugin[] {
       async buildStart() {
         // Defer the first compilation in test mode
         if (!isVitestVscode) {
-          pendingCompilation = performCompilation(resolvedConfig);
+          pendingCompilation = performCompilation(
+            this.environment?.config ?? resolvedConfig,
+          );
           await pendingCompilation;
           pendingCompilation = null;
 
@@ -421,7 +423,10 @@ export function angular(options?: PluginOptions): Plugin[] {
         if (TS_EXT_REGEX.test(ctx.file)) {
           let [fileId] = ctx.file.split('?');
 
-          pendingCompilation = performCompilation(resolvedConfig, [fileId]);
+          pendingCompilation = performCompilation(
+            ctx.server.environments.client.config,
+            [fileId],
+          );
 
           let result;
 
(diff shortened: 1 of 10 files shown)


arnoudb:
Build process hangs and doesn't return to the shell. looking into it

arnoudb:
Tried a few things, but the windows build fails due to memory allocation and i honestly have no idea how to solve it

brandonroberts:
@arnoudb you can ignore the windows CI issue

arnoudb:
> @arnoudb you can ignore the windows CI issue

Ah thnx for stepping in! Was about to ask for help on this issue. Tried a few things but can't get it to work on the windows box. On my mac it builds fine. Hard for me to debug...

I'll go trough the PR once more to see if the rest is working as expected.

brandonroberts:
Thanks @arnoudb. Could you double-check the development build path? With `vite build --mode development --sourcemap` (e.g. in `apps/analog-app`) I'm seeing the optimizer throw `Angular optimizer did not generate a source map for ...` on fesm files that don't need linking.

arnoudb:
> Angular optimizer did not generate a source map for

@brandonroberts I fixed the windows build (my pr broke it 😅), however i'm having a hard time reproducing the error you mentioned. 

I run from the project root: 
`pnpm exec nx build analog-app -c development --sourcemap --skip-nx-cache`

the only error i see is:
`ERROR TypeError: Cannot read properties of undefined (reading 'email')`

but I see that error also on the beta branch so i dont think it's related

packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
```diff
@@ -385,16 +411,64 @@ export function angular(options?: PluginOptions): Plugin[] {
         });
       },
       async buildStart() {
+        if (!jit) {
```
brandonroberts, line 421: I don't want to rely on extracting internal Vite plugins for css when preprocessCSS is the equivalent public API

brandonroberts (changes requested):

benpsnyder:
I tested analogjs/analog#2506 at `0e84b4abc` in an isolated worktree and found a reproducible build-watch regression in `releaseCompilation()`.

`styleTransform` is initialized in `configResolved()`, but the new `buildEnd()` cleanup clears it. For `vite build --watch`, `watchMode` is false because it only checks `command === 'serve'`. The next compilation therefore uses the identity fallback in `getStyleTransform()` and skips CSS preprocessing.

With an external SCSS file containing `$color: red; h1 { color: $color; }`, two successive `buildStart → transform → buildEnd` cycles produce compiled `color: red` on the first build and raw `$color` syntax in the component styles on the second. This probe uses the real Angular compiler and Vite CSS preprocessing, with `build.watch: {}`.

Could we retain the configured preprocessor when releasing compilation state?

```diff
     tsconfigOptionsCache.clear();
     includeCache = [];
-    styleTransform = undefined;
```

I added a regression test asserting that both builds contain `color: red` and neither contains `$color`. It fails on build 2 at the current head and passes with this one-line change. The full package test run passes with the fix: **733 passed, 6 skipped** (`pnpm nx test vite-plugin-angular --skip-nx-cache`, Node 24.15.0, locked Vite 8.0.16).

`pnpm nx build vite-plugin-angular --skip-nx-cache` also passes with the fix. This validation covers the default compiler's hook lifecycle; it is not a browser or filesystem-watcher end-to-end test. An experimental Angular Compilation API probe hit Zone.js worker errors in the test harness, so I am not claiming that path as validated.

<details>
<summary>Sample AI prompt to reproduce and implement the fix</summary>

```text
Investigate and fix the SCSS build-watch regression in analogjs/analog#2506
(reviewed head: 0e84b4abcc52b8ae2571c03e270cdee5378a423a).

Read the repository's AGENTS.md and contribution guidance. Work in an isolated
worktree of the PR's current head, preserving other work. Check whether the
finding still applies if the PR has advanced.

In packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts, trace
configResolved -> buildStart -> transform -> buildEnd -> releaseCompilation.
styleTransform is configured once, then cleared at buildEnd. With command
'build' and build.watch: {}, watchMode remains false, and the second compilation
uses getStyleTransform's identity fallback instead of preprocessing SCSS.

First add a failing regression test to angular-vite-plugin.spec.ts, reusing the
existing buildStart fixture and createAppBuildPlugin helper. Use Vite's real
resolveConfig with configFile: false, the fixture root, mode: 'production',
build.watch: {}, and css.preprocessorMaxWorkers: 0. Configure the Angular plugin
once, then run two successive buildStart -> transform -> buildEnd cycles.
The external SCSS fixture is '$color: red; h1 { color: $color; }'. Assert that
both emitted components contain 'color: red' and neither contains '$color'.
Use the real CSS preprocessor and make unexpected diagnostics fail the test.
Confirm that the current implementation fails on the second compilation.

Apply the smallest fix: retain styleTransform when releasing compilation state.
Keep the existing compiler/cache cleanup. Avoid changing watchMode globally,
because it also controls development/HMR behavior. Inspect the experimental
Angular Compilation API and JIT/disabled branches for effects of the change,
and distinguish verified behavior from test-harness limitations.

Run the focused regression, then the package tests and build through Nx:
pnpm nx test vite-plugin-angular --skip-nx-cache
pnpm nx build vite-plugin-angular --skip-nx-cache
Use a supported Node version and the PR's locked dependencies. Check formatting
of touched files. Report the diff, before/after test results, and any limits.
Do not post, commit, or push without the maintainer's instruction.
```

</details>

packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
arnoudb: Using `preprocessCSS` instead of `vite:css`gave me some problems. Maybe i'm testing this wrong but `preprocessCSS` seems to hang and the process needs explicit cleanup. Without it the build hangs and doesn't give back the prompt after a `vite build`. This leads to more code. Not really happy about the result, but don't know how to fix in another way.


BTW, the way i test it and notice the hanging build.

in my own project (big angular 21 project) i link to the checked out branch we discuss here like:

```json
 "devDependencies": {
    "@analogjs/vite-plugin-angular": "portal:../../analogjs/node_modules/@analogjs/vite-plugin-angular",
    "@analogjs/vitest-angular": "portal:../../analogjs/node_modules/@analogjs/vitest-angular",
   ...
```

without the last commit 'yarn build` runs but doesn give back the prompt

arnoudb:
NB when i rebased my branch on the latest beta branch, i noticed a regression. Commit c5a35924 breaks my global stylesheet.  (angular 21.2.20) project. That commit is the one that updates nx and angular. But i assume this still should work with an angular 21 project?

arnoudb:
@benpsnyder i removed the line you suggested to retain the configured preprocessor when releasing compilation state

arnoudb:
Unfortunately i can't see why the check fails on Vercel. I get a 404 when following the link.

brandonroberts (approved):
LGTM

PR #2544, opened 2026-09-07
fix(vite-plugin-angular): emit source-linked workspace packages

### LINKED ISSUE #2507: vite-plugin-angular: project dependencies are emitted as empty files in a pnpm workspace when linking to source code

https://github.com/eblocha/analogjs-vite-plugin-angular-repro

vite-plugin-angular

### Description

Versions

- @analogjs/vite-plugin-angular: 2.7.1
- @analogjs/astro-angular: 2.7.1
- vite: 8.2.2
- @angular/core: 21.2.22
- Package manager: pnpm 12 workspaces (linkWorkspacePackages: true)

Summary

This happens _without_ `fastCompile: true`.

With a pnpm workspace layout, where an application depends on a workspace library, the library's source code is emitted as an empty string.

Root cause

In angular-vite-plugin.js, the transform handler (in writeOutputFile):

```ts
let content = '';
let map: string | undefined;
let mapFilename: string | undefined;
emitBuilder.emit(
  sourceFile,
  (filename, data) => {
```
(around line 1483)

The `emitBuilder.emit` method skips files which are under a  `node_modules` path. This is an internal check inside of Typescript itself. Since the library being imported is resolved to a path with node modules (under `app/node_modules/library`), `emit` never calls the writer callback.

Workarounds tried

- fastCompile: true avoids this, because it transforms source code itself instead of passing it to typescript
- Passing include (the plugin's own option, not fastCompile) covering the library's source globs, e.g.:

```ts
angular({
  include: ['/path/to/lib/**/*.ts'],
})
```

- causes the files to get pre-analyzed by typescript, and marked internally as "non-node-modules" paths. When the emit builder reaches them, it continues with compilation as a result.

For me, `fastCompile: true` had some other issues related to the imperative component API (ViewContainerRef.createComponent), but I am still debugging those problems.

```true
[MISSING_EXPORT] "DemoDirective" is not exported by "../lib/src/index.ts".
   ╭─[ src/main.ts:5:10 ]
   │
 5 │ import { DemoDirective } from 'repro-lib';
   │          ──────┬──────  
   │                ╰──────── Missing export
───╯

    at aggregateBindingErrorsIntoJsError (file:///home/eb/projects/analogjs-vite-plugin-angular-repro/node_modules/.pnpm/rolldown@1.2.7/node_modules/rolldown/dist/shared/error-BP8lJHle.mjs:48:18)
    at unwrapBindingResult (file:///home/eb/projects/analogjs-vite-plugin-angular-repro/node_modules/.pnpm/rolldown@1.2.7/node_modules/rolldown/dist/shared/error-BP8lJHle.mjs:18:128)
    at #build (file:///home/eb/projects/analogjs-vite-plugin-angular-repro/node_modules/.pnpm/rolldown@1.2.7/node_modules/rolldown/dist/shared/rolldown-DfXO5K1m.mjs:133:34)
    at async buildEnvironment (file:///home/eb/projects/analogjs-vite-plugin-angular-repro/node_modules/.pnpm/vite@8.2.2_esbuild@0.28.1_sass@1.97.3/node_modules/vite/dist/node/chunks/node.js:33821:66)
    at async Object.build (file:///home/eb/projects/analogjs-vite-plugin-angular-repro/node_modules/.pnpm/vite@8.2.2_esbuild@0.28.1_sass@1.97.3/node_modules/vite/dist/node/chunks/node.js:34242:19)
    at async Object.buildApp (file:///home/eb/projects/analogjs-vite-plugin-angular-repro/node_modules/.pnpm/vite@8.2.2_esbuild@0.28.1_sass@1.97.3/node_modules/vite/dist/node/chunks/node.js:34239:153)
    at async CAC.<anonymous> (file:///home/eb/projects/analogjs-vite-plugin-angular-repro/node_modules/.pnpm/vite@8.2.2_esbuild@0.28.1_sass@1.97.3/node_modules/vite/dist/node/cli.js:776:3) {
  errors: [Getter/Setter]
}
```

### Other information

_No response_

- [ ] Yes
- [ ] No

benpsnyder:
Fix empty output for source-linked pnpm workspace packages in the normal Angular compiler. When TypeScript skips an external-classified source requested by Vite, promote that source to a compilation root and retry the normal Angular emit.

## PR Checklist

Closes analogjs/analog#2507.

## Affected scope

- Secondary scopes: none

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

No commit boundaries need preservation.

## What is the new behavior?

A pnpm workspace library exporting a TypeScript barrel now builds with `angular()` without requiring a manual `include` glob. TypeScript previously classified the real linked source as external and skipped its emit, so Vite received an empty module and reported missing exports.

The retry only applies to an empty result whose source TypeScript identifies as external. Roots are deduplicated and retained for incremental recompilation; tsconfig change/add/unlink events clear the discovered roots, and unlinking a promoted source removes its root and cached output. Emission continues through Angular's existing serialized compilation path and its transformers. Existing node_modules filters, fastCompile, explicit includes, and missing-file fallbacks are unchanged.

## Test plan

- [x] Frozen dependency installation, pnpm 11.6.0 on Node 24.15.0.
- [x] New real-compiler regression fails before the fix with an empty linked barrel.
- [x] `pnpm nx test vite-plugin-angular` — 721 passed, 6 existing skips.
- [x] `pnpm nx build vite-plugin-angular` — passed.
- [x] Packed the built plugin and installed it in the [reporter's reproduction](https://github.com/eblocha/analogjs-vite-plugin-angular-repro). `pnpm run build` now passes: 245 modules on Angular 21.2.22, TypeScript 5.9.3, and Vite 8.2.2, with no `include` workaround. The original dependency reproduced the missing `DemoDirective` export.
- [x] Prettier on changed files and `git diff --check`.

The regression uses an actual symlinked package with a TypeScript export, checks both its barrel and compiled Angular directive, repeats the barrel transform, verifies an edit to the directive survives incremental recompilation, then unlinks the barrel and checks that the next compilation cannot return stale output. The unlink regression also fails against the implementation before the cleanup fix.

- [ ] Yes
- [x] No

## Other information

Targets beta. [analogjs/analog#2520](https://github.com/analogjs/analog/pull/2520) refactors alpha's compiler roots but does not address TypeScript's external-source emit classification. This fix should be adapted to that refactor when ported to alpha; no changes are made to its branch here.

## [optional] What gif best describes this PR or how it makes you feel?

Not applicable.

### DIFF
--- a/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
+++ b/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
@@ -33,6 +33,7 @@ import {
 } from './component-resolvers.js';
 import {
   augmentHostWithCaching,
+  augmentHostWithModuleResolution,
   augmentHostWithResources,
   augmentProgramWithVersioning,
   mergeTransformers,
@@ -77,6 +78,7 @@ import { fastCompilePlugin } from './fast-compile-plugin.js';
 import { ANGULAR_DECORATOR_CALL_RE } from './compiler/index.js';
 import {
   TS_EXT_REGEX,
+  EXCLUDED_TS_EXT_REGEX,
   createTsConfigGetter,
   getTsConfigPath,
   createDepOptimizerConfig,
@@ -1387,55 +1389,7 @@ export function angular(options?: PluginOptions): Plugin[] {
           return file;
         },
       });
-      const resolutionCache = ts.createModuleResolutionCache(
-        host.getCurrentDirectory(),
-        host.getCanonicalFileName.bind(host),
-        tsCompilerOptions,
-      );
-      host.getModuleResolutionCache = () => resolutionCache;
-      host.resolveModuleNameLiterals = (
-        literals,
-        containingFile,
-        redirectedReference,
-        compilerOptions,
-        containingSourceFile,
-      ) =>
-        literals.map((literal) => {
-          const resolution: ts.ResolvedModuleWithFailedLookupLocations =
-            ts.resolveModuleName(
-              literal.text,
-              containingFile,
-              compilerOptions,
-              host,
-              resolutionCache,
-              redirectedReference,
-              ts.getModeForUsageLocation(
-                containingSourceFile,
-                literal,
-                compilerOptions,
-              ),
-            );
-          const resolvedModule = resolution.resolvedModule;
-          if (
-            resolvedModule?.isExternalLibraryImport &&
-            TS_EXT_REGEX.test(resolvedModule.resolvedFileName) &&
-            !EXCLUDED_TS_EXT_REGEX.test(resolvedModule.resolvedFileName) &&
-            !normalizePath(resolvedModule.resolvedFileName).includes(
-              '/node_modules/',
-            )
-          ) {
-            // A workspace symlink resolves to source outside node_modules.
-            // Classify it before program creation so TypeScript emits it.
-            return {
-              ...resolution,
-              resolvedModule: {
-                ...resolvedModule,
-                isExternalLibraryImport: false,
-              },
-            };
-          }
-          return resolution;
-        });
+      augmentHostWithModuleResolution(host, tsCompilerOptions);
       cachedHost = host;
       cachedHostKey = hostKey;
 
@@ -1718,9 +1672,6 @@ export function angular(options?: PluginOptions): Plugin[] {
 }
 
 const COMPONENT_RESOURCE_EXT_REGEX = /\.(html|htm|css|scss|sass|less)$/;
-// Spec files stay included — a newly added spec must join the program's
-// root names in Vitest watch mode.
-const EXCLUDED_TS_EXT_REGEX = /\.d\.[cm]?ts$/;
 
 export function createFsWatcherCacheInvalidator(
   invalidateFsCaches: () => void,
--- a/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
+++ b/packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
@@ -211,6 +211,7 @@ export function angular(options?: PluginOptions): Plugin[] {
   let cachedHost: ts.CompilerHost | undefined;
   let cachedHostKey: string | undefined;
   let includeCache: string[] = [];
+  const linkedSourceRoots = new Set<string>();
   function invalidateFsCaches() {
     includeCache = [];
   }
@@ -380,6 +381,7 @@ export function angular(options?: PluginOptions): Plugin[] {
         server.watcher.on('unlink', invalidateCompilationOnFsChange);
         server.watcher.on('change', (file) => {
           if (file.includes('tsconfig')) {
+            linkedSourceRoots.clear();
             invalidateTsconfigCaches();
           }
         });
@@ -733,7 +735,22 @@ export function angular(options?: PluginOptions): Plugin[] {
             pendingCompilation = null;
           }
 
-          const typescriptResult = fileEmitter(id);
+          let typescriptResult = fileEmitter(id);
+          const sourceFile = builder?.getSourceFile(id);
+          if (
+            typescriptResult?.content === '' &&
+            sourceFile &&
+            builder!.getProgram().isSourceFileFromExternalLibrary(sourceFile) &&
+            !linkedSourceRoots.has(normalizePath(sourceFile.fileName))
+          ) {
+            // TypeScript skips emit for source-linked package entries until
+            // they are roots. Only promote files requested by Vite.
+            linkedSourceRoots.add(normalizePath(sourceFile.fileName));
+            pendingCompilation = performCompilation(resolvedConfig, [id]);
+            await pendingCompilation;
+            pendingCompilation = null;
+            typescriptResult = fileEmitter(id);
+          }
 
           // File not in the Angular program — skip and let other plugins
           // or Vite's built-in transform handle it. Warn if it looks like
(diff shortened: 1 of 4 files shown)


packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
```diff
@@ -733,7 +747,22 @@ export function angular(options?: PluginOptions): Plugin[] {
...
+          if (
+            typescriptResult?.content === '' &&
+            sourceFile &&
+            builder!.getProgram().isSourceFileFromExternalLibrary(sourceFile) &&
+            !linkedSourceRoots.has(normalizePath(sourceFile.fileName))
+          ) {
+            // TypeScript skips emit for source-linked package entries until
+            // they are roots. Only promote files requested by Vite.
+            linkedSourceRoots.add(normalizePath(sourceFile.fileName));
+            pendingCompilation = performCompilation(resolvedConfig, [id]);
```
brandonroberts: This is going to be very inefficient in performing compilation for each file

brandonroberts (changes requested):

packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
benpsnyder: Addressed in [b24382814](https://github.com/benpsnyder/analog/commit/b243828147b6a3e996c5a074ee0f05d77ff524a5).

Integrated the current `beta` compiler/source-map changes and resolved the merge conflict in [b1a208335](https://github.com/benpsnyder/analog/commit/b1a208335cad2b90cae1225a55e3f68f99040b04). The validation below covers this final head.

Linked workspace sources are now classified during TypeScript's cached module resolution, before Angular constructs its program. The resolver preserves TypeScript's per-import resolution mode and only changes the external-library flag for non-declaration TypeScript files resolving outside `node_modules`. This removes the per-file compilation retry and the promoted-root set while retaining normal Angular emit and deleted-file cache cleanup.

The real-compiler regression imports a barrel, an Angular directive, and eight additional package entry points, including concurrent transforms. The previous implementation constructs **10 Angular programs** for this fixture; the new implementation constructs **1**, with exactly one additional program after an edit. The tests cover AOT and JIT in build/serve modes, type checking, external-package/declaration boundaries, and unlink cleanup.

Validation:

- `pnpm nx test vite-plugin-angular`: **738 passed, 6 existing skips**.
- `pnpm nx build vite-plugin-angular`, changed-file formatting, and `git diff --check`: passed.
- Packed plugin installed in a fresh checkout of the reporter's reproduction: builds pass without an `include` workaround on **Vite 6.4.3, 7.3.6, and 8.2.2**, using Angular 21.2.22 / TypeScript 5.9.3. The original dependency reproduces the missing export.
- The actual resolution callback also passes **34 conditional-export and linked-source emit checks** against TypeScript **5.2.2, 5.4.5, 5.9.3, and 6.0.3**, matching each version's native resolution.

Additional checks found inherited limitations: the package-wide lint target reports 19 existing errors, with no new diagnostics in either changed file. An Angular 17.3.12 / TypeScript 5.4.5 smoke test fails at the unchanged `js-emitter.js` import (`ts.createPrinter is not a function`), also reproduced with the previous artifact before the new resolver executes.

packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
```diff
@@ -1301,6 +1307,55 @@ export function angular(options?: PluginOptions): Plugin[] {
...
+              containingFile,
+              compilerOptions,
+              host,
+              resolutionCache,
+              redirectedReference,
+              ts.getModeForUsageLocation(
+                containingSourceFile,
+                literal,
+                compilerOptions,
+              ),
```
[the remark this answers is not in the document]
benpsnyder: I checked this against the released compiler APIs and could not reproduce a compatibility problem. The three-argument signature is already present in [TypeScript 5.4.2](https://github.com/microsoft/TypeScript/blob/v5.4.2/src/compiler/program.ts) and [5.4.5](https://github.com/microsoft/TypeScript/blob/v5.4.5/src/compiler/program.ts); both pass `compilerOptions` through to the resolution-mode worker. TypeScript 5.2.2 uses its two-argument implementation, which also supplies the mode used by its native resolver.

I ran the actual callback extracted from this PR through unmodified TypeScript **5.2.2, 5.4.5, 5.9.3, and 6.0.3**. All **34 conditional-export cases** matched each version's native program resolution and emitted the linked source. These cover NodeNext ESM/CJS, dynamic imports, import-equals, explicit type-import resolution modes, bundler imports, and `module: preserve` import/require where supported.

No version branch or minimum-version increase is needed for this call. The separate Angular 17 smoke-build limitation [documented in the main review reply](https://github.com/analogjs/analog/pull/2544#discussion_r3953001880) occurs in the pre-existing emitter import, before this callback runs.

benpsnyder:
@brandonroberts should be good to go

packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
```diff
@@ -1381,6 +1387,55 @@ export function angular(options?: PluginOptions): Plugin[] {
           return file;
         },
       });
+      const resolutionCache = ts.createModuleResolutionCache(
```
brandonroberts: This code needs to live in the `host.ts` file with the other TS host augmentations

brandonroberts (changes requested):

packages/vite-plugin-angular/src/lib/angular-vite-plugin.ts
benpsnyder: done

brandonroberts (approved):

PR #2548, opened 2026-09-07
fix(vite-plugin-nitro): trace TypeScript server imports without bundling

### LINKED ISSUE #2457: Importing `typescript` in a server route fails (`Cannot access 'typescript' before initialization` in dev, `__filename is not defined in ES module scope` in production)
## Summary

Importing the `typescript` package inside an Analog server route fails at runtime. The same code works correctly in a plain Node/Fastify application and when loading TypeScript via `createRequire()`, but fails when using `import()`.

## Background

I was building a server-side TypeScript analysis endpoint and needed to load the TypeScript compiler.

The following works correctly:

```ts
import { createRequire } from "node:module";

const require = createRequire(import.meta.url);

const ts = require("typescript");
```

However, replacing it with:

```ts
const ts = await import("typescript");
```

causes runtime failures in both development and production.

The issue occurs in both development and production builds.

## Environment

- @analogjs/platform: 2.6.4
- @analogjs/vite-plugin-angular: 2.6.4
- Angular: 22.1.0
- Vite: 8.1.5
- Nitro: 2.13.4
- TypeScript: 6.0.3
- Node.js: 22.22.3
- OS: Windows 11

## Reproduction

Create a new Analog project:

```bash
npm create analog@latest
npm install
npm install typescript
```

Create a server route:

```ts
import { defineEventHandler } from "h3";

export default defineEventHandler(async () => {
  const ts = await import("typescript");

  return {
    version: ts.version,
  };
});
```

Call the endpoint.

For example:

POST /api/v1/test

## Expected

The endpoint should return:

```json
{
  "version": "6.0.3"
}
```

### Development (`npm run dev`)

HTTP 500

```
Cannot access 'typescript' before initialization
```

### Production (`npm run build` + `npm run preview`)

HTTP 500

Server console:

```
ReferenceError: __filename is not defined in ES module scope

at getNodeSystem(...)
dist/analog/server/chunks/_/typescript.mjs
```

The generated bundle contains:

```js
const executingFilePath =
  __filename.endsWith("sys.js")
    ? _path.join(_path.dirname(__dirname), "__fake__.js")
    : __filename;
```

which fails because `__filename` is not available in an ES module.

## Additional Observations

The following imports work correctly:

```ts
await import("node:fs");
await import("marked");
```

Loading TypeScript through Node's CommonJS loader also works:

```ts
import { createRequire } from "node:module";

const require = createRequire(import.meta.url);

const ts = require("typescript");
```

The same code works correctly in a standalone Fastify application.

I also tested:

- `nitro.externals.external`
- `ssr.external`
- `ssr.noExternal`

but the behavior remained unchanged.

The generated production output contains:

dist/analog/server/chunks/_/typescript.mjs

indicating that the TypeScript package is emitted as part of the server bundle.

## Investigation

During debugging I verified the following:

- The issue is isolated to importing the `typescript` package.
- `node:fs` and `marked` import successfully.
- The issue occurs in both development and production.
- Loading TypeScript through `createRequire()` avoids the problem.
- The same code works correctly in a Fastify application.
- The production stack trace points to `dist/analog/server/chunks/_/typescript.mjs`, where the generated code accesses `__filename`, which is unavailable in an ES module.

## Possible Cause

From the generated production output, it appears that the `typescript` package is being bundled into:

```
dist/analog/server/chunks/_/typescript.mjs
```

The generated file still references CommonJS globals such as `__filename`, which results in:

```
ReferenceError: __filename is not defined in ES module scope
```

I'm not sure whether this originates in Analog, Nitro, or the underlying bundling pipeline, but I wanted to report it here since it is reproducible in an Analog application.

## Additional Information

If needed, I can provide a minimal reproduction repository.

### Workaround

Loading TypeScript through `createRequire()` works correctly:

```ts
import { createRequire } from "node:module";

const require = createRequire(import.meta.url);

const ts = require("typescript");
```

This avoids the issue, but using `await import("typescript")` continues to fail.

## Question

Is importing the `typescript` package inside Analog server routes expected to be supported?

If so, I'd be happy to provide a minimal reproduction repository to help investigate.

benpsnyder:
Keep Node server routes importing `typescript` executable in development and deployed output. Externalize the compiler only when it is imported, and let Nitro trace it into the deployment's dependencies.

## PR Checklist

Closes analogjs/analog#2457.

## Affected scope

- Primary scope: vite-plugin-nitro
- Secondary scopes: none

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

No commit boundaries need preservation.

## What is the new behavior?

A Node API route using `await import('typescript')` retains that bare import. Previously Nitro bundled TypeScript's CommonJS runtime into an ESM chunk that failed on `__filename`; the reported dev initialization failure also reproduces in Nitro 2.

A small Nitro module registers a resolver before Nitro's default resolver and adds the application's resolved compiler entry to Nitro's existing trace list only when imported. This preserves deployment packaging, unlike a top-level Rollup external rule alone. Applications that do not import TypeScript receive no compiler payload. Non-Node and `noExternals` configurations retain their existing behavior.

User Rollup external rules and hooks are preserved. Trace initialization occurs after the final config merge, including the Windows SSR externals branch.

## Test plan

- [x] Frozen install and `pnpm nx build vite-plugin-nitro`.
- [x] `pnpm nx test vite-plugin-nitro` — 70 passed, 11 existing skips.
- [x] Reproduced the original runtime error with plain Nitro 2.13.4 / TypeScript 6.0.3.
- [x] Qualified the rebuilt plugin artifact through a real Vite 8 / Nitro 2 fixture: a dev GET and production GET both return TypeScript 6.0.3.
- [x] Relocated the built server outside the app's dependency tree; GET still succeeds. Output includes the traced TypeScript dependency and no bundled `typescript.mjs` chunk.
- [x] Removed the TypeScript import and rebuilt: no copied compiler directory, chunk, or output dependency.
- [x] Unit regressions check lazy tracing, deduplication, preserved user hooks/external rules, Windows SSR configuration, and non-Node/noExternals behavior.
- [x] Prettier and `git diff --check`.

Runtime qualification was on Linux; the Windows-specific configuration branch is covered by a mocked-platform regression. This does not claim a fresh Windows runtime execution of the API fixture.

- [ ] Yes
- [x] No

## Other information

Targets beta's Nitro 2 adapter. The open Nitro 3 migration remains separate. Other CommonJS packages and TypeScript deep imports are outside this narrow fix.

## [optional] What gif best describes this PR or how it makes you feel?

Not applicable.

### DIFF
--- a/packages/vite-plugin-nitro/src/lib/vite-plugin-nitro.ts
+++ b/packages/vite-plugin-nitro/src/lib/vite-plugin-nitro.ts
@@ -6,6 +6,7 @@ import { dirname, join, relative, resolve } from 'node:path';
 import { platform } from 'node:os';
 import { fileURLToPath, pathToFileURL } from 'node:url';
 import { existsSync, readFileSync } from 'node:fs';
+import { createRequire } from 'node:module';
 
 import { buildServer } from './build-server.js';
 import { buildSSRApp } from './build-ssr.js';
@@ -189,6 +190,32 @@ export function nitro(options?: Options, nitroOptions?: NitroConfig): Plugin[] {
                 ],
               }
             : {}),
+          modules: [
+            (nitro) => {
+              nitro.hooks.hook('rollup:before', (_nitro, rollupConfig) => {
+                if (!nitro.options.node || nitro.options.noExternals) return;
+                rollupConfig.plugins = [
+                  {
+                    name: 'analog-typescript-external',
+                    resolveId(source) {
+                      if (source !== 'typescript') return null;
+                      const entry = createRequire(
+                        resolve(workspaceRoot, rootDir, 'package.json'),
+                      ).resolve(source);
+                      // Nitro retains this array for tracing. Add the compiler
+                      // only when imported, including in relocated deployments.
+                      const traceInclude =
+                        nitro.options.externals.traceInclude!;
+                      if (!traceInclude.includes(entry))
+                        traceInclude.push(entry);
+                      return { id: source, external: true };
+                    },
+                  },
+                  rollupConfig.plugins,
+                ];
+              });
+            },
+          ],
           rollupConfig: {
             onwarn(warning) {
               if (
@@ -439,6 +466,8 @@ export function nitro(options?: Options, nitroOptions?: NitroConfig): Plugin[] {
           nitroConfig,
           nitroOptions as Record<string, any>,
         );
+        nitroConfig.externals ??= {};
+        nitroConfig.externals.traceInclude ??= [];
 
         return {
           environments: {
--- a/packages/vite-plugin-nitro/src/lib/typescript-externals.spec.ts
+++ b/packages/vite-plugin-nitro/src/lib/typescript-externals.spec.ts
@@ -45,7 +45,7 @@ describe('TypeScript server externalization', () => {
       options: { node: true, externals: config.externals },
       hooks: { hook },
     };
-    await (config.modules![0] as Function)(instance);
+    await (config.modules![0] as (nitro: typeof instance) => void)(instance);
     const existing = { name: 'consumer-plugin' };
     const rollup = { plugins: [existing] };
     await hook.mock.calls[0][1](instance, rollup);
@@ -78,7 +78,7 @@ describe('TypeScript server externalization', () => {
         options: { ...options, externals: config.externals },
         hooks: { hook },
       };
-      await (config.modules![0] as Function)(instance);
+      await (config.modules![0] as (nitro: typeof instance) => void)(instance);
       const rollup = { plugins: [] };
       await hook.mock.calls[0][1](instance, rollup);
       expect(rollup.plugins).toEqual([]);
--- a/packages/vite-plugin-nitro/src/lib/typescript-externals.spec.ts
+++ b/packages/vite-plugin-nitro/src/lib/typescript-externals.spec.ts
@@ -1,0 +1,88 @@
+import { createRequire } from 'node:module';
+import { resolve } from 'node:path';
+import { describe, expect, it, vi } from 'vitest';
+import { nitro } from './vite-plugin-nitro';
+import { buildServer } from './build-server';
+
+vi.mock('./build-server');
+vi.mock('node:os', async (importOriginal) => ({
+  ...(await importOriginal<typeof import('node:os')>()),
+  platform: () => 'win32',
+}));
+
+async function configure(ssr = false) {
+  const userHook = vi.fn();
+  const external = vi.fn(() => false);
+  const plugin = nitro(
+    {
+      ssr: false,
+    },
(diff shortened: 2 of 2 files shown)


packages/vite-plugin-nitro/src/lib/vite-plugin-nitro.ts
```diff
@@ -189,6 +190,32 @@ export function nitro(options?: Options, nitroOptions?: NitroConfig): Plugin[] {
                 ],
               }
             : {}),
+          modules: [
+            (nitro) => {
+              nitro.hooks.hook('rollup:before', (_nitro, rollupConfig) => {
```
brandonroberts, line 195: I'd rather not depend on rollup plugins in preparation for nitro v3

brandonroberts (changes requested):

packages/vite-plugin-nitro/src/lib/vite-plugin-nitro.ts
benpsnyder: Agreed. I traced this to Nitro 2’s mlly validation falsely detecting ESM syntax inside TypeScript’s diagnostic strings. Adding externals.external still hits that rejection. I’d like to pursue the underlying fix upstream and retain the portable-output regression coverage, while checking Nitro v3 separately.

benpsnyder: The upstream fixes are now drafted: [unjs/mlly#370](https://github.com/unjs/mlly/pull/370) makes ESM detection token-aware, and [nitrojs/nitro#4614](https://github.com/nitrojs/nitro/pull/4614) opts both Nitro 2 externals resolvers into comment stripping.

Both are needed: baseline and either change alone reproduce HTTP 500; paired packed builds return TypeScript 6.0.3 from relocated output with a traced compiler dependency and no bundled compiler chunk. Development, legacy externals, unused-compiler output, and Analog API-route integration with Vite 6–8 are covered.

The [16-case packed runtime matrix passes on both Linux and Windows](https://github.com/benpsnyder/nitro/actions/runs/34663991217).

Nitro v3 remains separate: `nitro@3.0.260903-beta` passes the direct development request but fails relocated production with `__filename is not defined` in the bundled compiler. The published Analog alpha also has an earlier undeclared `magic-string` import.

I'm retaining this PR and its portable-output regression coverage and leaving this thread open until the released upstream dependency path is qualified. @pi0 kindly assist - thx friend

pi0: thanks for issues/PRs. nitro v2 maintenance might take time. please consider dropping it in your roadmap.

anything can help to resolve issues on mainstream v3 i can gladly assist :)

brandonroberts: I'm also fine with not fixing this. There's a known workaround

PR #2375, opened 2026-06-11
feat(create-analog): Node.js compat with Angular v22 and @angular/animations deprecation

alejandrocuba:
## PR Checklist

- The monorepo and starter templates have older Node versions (Node v20) that do not align with Angular v22
- The deprecated `@angular/animations` package is also specified as a dependency in the repo and starter templates.
- pnpm updated to v11

## Affected scope

- Primary scope: create-analog

## Recommended merge strategy for maintainer [optional]

- [x] Squash merge
- [ ] Rebase merge
- [ ] Other

## What is the new behavior?

- Updated Node engine requirements to v22, v24, v26
- Updated pnpm engine requirements to v11, and migrated legacy configurations from `package.json` to the `pnpm-workspace.yaml` file under `allowBuilds` and `overrides`.
- Removed deprecated `@angular/animations`

## Test plan

- [x] `nx format:check`
> All matched files use Prettier code style
- [x] `pnpm build`:
> Completed 20 build tasks
- [x] `pnpm test`
> Successfully ran target test for 18 projects and 9 tasks they depend on
- [x] Manual verification
> Verified clean workspace installation with `pnpm i`

## Does this PR introduce a breaking change?

- [ ] Yes
- [x] No

This PR drops support for Node v20. Anyone working in the workspace or running scaffolding templates must upgrade their Node.js runtime environment to at least v22.22.3

PR #2560, opened 2026-09-10
feat(astro-angular): support content projection for island components

### LINKED ISSUE #917: Support ng-content inside the astro-angular package

astro-angular

### Information

Content projection is done easily inside Angular itself using `<ng-content />`.
Anything added inside the opening and closing components tags would be rendered here. 

For example:

```
<my-card variant="lg">
Hello World
</my-card>
```

This would render the my-card with any attributes given to the component. And additionally if the component has a ng-content slot defined it would render my Hello world text inside of here.

I would expect that in Astro the same behaviour would happen but i understand it does not.
If we would support this we would get even more Angular goodies inside Astro.

Besides having normal slots, you can also have named slots.
For example this template:

```
    <h2>Multi-slot content projection</h2>

    Default:
    <ng-content></ng-content>

    Question:
    <ng-content select="[question]"></ng-content>
```

You would render the corresponding content of it using an attribute like this:

```
<app-zippy-multislot>
  <p question>
    Is content projection cool?
  </p>
  <p>Let's learn about content projection!</p>
</app-zippy-multislot>
```

This would probably be more tricky to support, but maybe this could also be tackled?

Instead of ng-content you could make inputs for everything, but this will not give the same freedom and therefor will only work for simple components.

- [X] Yes
- [ ] No

(I have no idea where to start tho and it might be to hard for me)

thebspin:
## PR Checklist

  Angular components rendered as Astro islands discarded their children. Both server renderers received the slot content as
  the third argument and never used it, and the client hydrators ignored it too. The docs listed "Content projection to island
  components is not supported" as a known limitation.

  Closes #917

  ## Affected scope

  - Secondary scopes: none

  - [x] Squash merge
  - [ ] Rebase merge
  - [ ] Other

  ## What is the new behavior?

  Children of an Angular island are projected into its `ng-content` slots using the component's own `select` semantics, so
  existing Angular components work unchanged and no Astro `slot` attributes are needed:

  ```astro
  <CardComponent client:visible>
    <p question>Is content projection cool?</p>
    <p>Let's learn about content projection!</p>
  </CardComponent>

  - Astro's slot HTML is parsed into the render document and distributed over reflectComponentType(...).ngContentSelectors
    (Angular does not apply select matching to projectableNodes, so the integration does it).
  - Text-only children go to the default slot. Unmatched nodes are dropped when there is no default slot, as in Angular.
    ng-content fallback content still renders when nothing is projected.
  - Works for static, hydrated (client:*) and client:only islands, on both the default client path and useAngularHydration.
  - Both server renderers now use createApplication + createComponent with bindings (first commit), since bootstrapApplication
    has no way to pass projectableNodes. This replaces the BEFORE_APP_SERIALIZED / APP_BOOTSTRAP_LISTENER input hooks.
  - Docs: the limitation is removed and a Content Projection section is added (EN, zh-hans, README).

  One trade-off worth knowing: the renderer cannot emit Astro's <astro-slot> markers, because wrapping the nodes would make
  Angular's selectors match the wrapper instead of the real element. Astro therefore falls back to <template
  data-astro-template> for hydrated islands, which the client island already reads as a slot source. The slot markup appears
  twice in the HTML response of hydrated islands (once rendered, once in the inert template, which is removed on hydration).
  Static islands are unaffected.

  Test plan

  - [x] nx format:check
  - [x] nx build astro-angular
  - [x] nx test astro-angular (new projection.spec.ts and server.spec.ts, the latter renders a real component through both
    server renderers)
  - [x] Manual verification in apps/astro-app with Playwright on both useAngularHydration: true and false:
    - static, client:load and text-only islands project correctly in SSR HTML and after hydration, no NG05xx hydration errors,
      fallback templates removed
    - a nested client:load island inside projected content hydrates and stays interactive
    - strictStylePlacement middleware leaves projected content intact

  Note: running astro-app locally currently requires removing provideAnimations from the card demo, since @angular/animations
  was dropped from the workspace in #2375. That is pre-existing and not touched here.

  Does this PR introduce a breaking change?

  - [ ] Yes
  - [x] No

  Other information

  Verified against Angular 22.1.5 and Astro 6.0.4. The bindings and BootstrapContext APIs used here exist since Angular 20,
  which matches the package's peer dependency range


brandonroberts (changes requested):
The approach looks reasonable, but I’d request these changes before merging:

- **Preserve projected DOM during hydration.** Freshly parsed nodes diverge from the SSR nodes. Toggling projected content with `@if` leaves stale content and creates duplicates. Add a client hydration regression test covering hide/show behavior.
- **Restore bootstrap lifecycle ordering.** Run initial change detection before invoking `APP_BOOTSTRAP_LISTENER`. Both server renderers currently notify listeners before inputs, `ngOnInit`, and the rendered view are initialized.
- **Match Angular’s `ngProjectAs` semantics.** Replace synthetic-element matching with selector comparison that handles aliases correctly, including `:not(...)` and compound selectors.
- **Use the last wildcard slot.** Exclude every wildcard from selective matching and route unmatched content to the last default slot, consistent with Angular.

The existing 29 tests pass, but focused reproductions expose these gaps.

thebspin:
> The approach looks reasonable, but I’d request these changes before merging:
> 
> * **Preserve projected DOM during hydration.** Freshly parsed nodes diverge from the SSR nodes. Toggling projected content with `@if` leaves stale content and creates duplicates. Add a client hydration regression test covering hide/show behavior.
> * **Restore bootstrap lifecycle ordering.** Run initial change detection before invoking `APP_BOOTSTRAP_LISTENER`. Both server renderers currently notify listeners before inputs, `ngOnInit`, and the rendered view are initialized.
> * **Match Angular’s `ngProjectAs` semantics.** Replace synthetic-element matching with selector comparison that handles aliases correctly, including `:not(...)` and compound selectors.
> * **Use the last wildcard slot.** Exclude every wildcard from selective matching and route unmatched content to the last default slot, consistent with Angular.
> 
> The existing 29 tests pass, but focused reproductions expose these gaps.

I tried to implement the feedback as good as possible.
What do you think of the current code?

brandonroberts:
Looks better. This still doesn't support components in the projected content, correct?

thebspin:
Correct, but with a nuance i believe... Angular components inside the slot do work, they just become their own islands (static or hydrated). What doesn't work is the Angular-level connection between the card and a component in its slot:
@ContentChild(ren), inputs/outputs, DI.

As far as I can tell that's exactly how the other integrations behave. React gets its children as a StaticHtml blob via
dangerouslySetInnerHTML, Vue and Svelte do the same with innerHTML / {@html}. A Svelte component nested in another Svelte component's slot in an .astro file is also two separate islands with no shared context. Real parent/child composition happens inside the framework file, and that's the same for us: nest the components in an Angular template and use that as the island.

Making .astro children real Angular children would mean compiling the slot content as an Angular template at runtime,
which I don't think belongs in this PR. 

If you want i can add this to Current Limitations with the "compose inside an Angular template" workaround.
Unless i'm missing out on something?

brandonroberts:
> Correct, but with a nuance i believe... Angular components inside the slot do work, they just become their own islands (static or hydrated). What doesn't work is the Angular-level connection between the card and a component in its slot: @ContentChild(ren), inputs/outputs, DI.
> 
> As far as I can tell that's exactly how the other integrations behave. React gets its children as a StaticHtml blob via dangerouslySetInnerHTML, Vue and Svelte do the same with innerHTML / {@html}. A Svelte component nested in another Svelte component's slot in an .astro file is also two separate islands with no shared context. Real parent/child composition happens inside the framework file, and that's the same for us: nest the components in an Angular template and use that as the island.
> 
> Making .astro children real Angular children would mean compiling the slot content as an Angular template at runtime, which I don't think belongs in this PR.
> 
> If you want i can add this to Current Limitations with the "compose inside an Angular template" workaround. Unless i'm missing out on something?

Thanks for the explanation. I think updating the current limitations with that is fine

thebspin:
> Thanks for the explanation. I think updating the current limitations with that is fine

I did a push. (I did us AI to translate the docs to Chinese because, well i don't speak Chinese.. If you prefer not updating the Chinese docs i can revert those ofcourse

brandonroberts (approved):

